PR #13, opened 2019-07-25
Adding CGO_ENABLED option to fix failed docker build

dcaceres88:
Hi, 

This pr is for solving the problem addressed in [https://github.com/cloudera/yunikorn-core/issues/18](url).

Thanks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ build: init
 .PHONY: build_image
 build_image: init
 	@echo "building binary for scheduler docker image"
-	GOOS=linux GOARCH=amd64 \
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 \
 	go build -a -o=${RELEASE_BIN_DIR}/${BINARY} -ldflags \
 	'-extldflags "-static" -X main.version=${VERSION} -X main.date=${DATE}' \
 	-tags netgo -installsuffix netgo \
--- a/deployments/scheduler/scheduler-load.yaml
+++ b/deployments/scheduler/scheduler-load.yaml
@@ -49,4 +49,3 @@ spec:
   selector:
     app: yunikorn
   type: LoadBalancer
-
--- a/deployments/scheduler/scheduler-load.yaml
+++ b/deployments/scheduler/scheduler-load.yaml
@@ -23,7 +23,6 @@ spec:
             - name: config-volume
               mountPath: /etc/yunikorn/
           ports:
-          ports:
             - containerPort: 9080
         - name: yunikorn-scheduler-web
           image: yunikorn/yunikorn-web:latest
--- a/deployments/scheduler/scheduler-load.yaml
+++ b/deployments/scheduler/scheduler-load.yaml
@@ -19,6 +19,10 @@ spec:
         - name: yunikorn-scheduler-k8s
           image: yunikorn/yunikorn-scheduler-k8s:latest
           imagePullPolicy: IfNotPresent
+          volumeMounts:
+            - name: config-volume
+              mountPath: /etc/yunikorn/
+          ports:
           ports:
             - containerPort: 9080
         - name: yunikorn-scheduler-web
--- a/deployments/scheduler/scheduler-load.yaml
+++ b/deployments/scheduler/scheduler-load.yaml
@@ -49,3 +49,4 @@ spec:
   selector:
     app: yunikorn
   type: LoadBalancer
+


sunilgovind:
I think this approach is good.
For config-volume change, i think this is fine and wont cause any issues in local non-gcp deployment. @yangwwei  @wilfred-s  any thoughts ?

wilfred-s:
Adding CGO_ENABLED=0 means we are compiling libc into the binary. Due to the GPL license of libc I am not sure if we really want to do that when we push public images.
Second point to make is the fact that it all works with the flags shows that there is a compatibility problem between the binary and libraries. Most likely libc version differences, net libraries, missing or something like that.
CGO_ENABLED=0 will also prevent us from using plugins in the future. All those things combined I am not for it at all.

For the volume mount: what is the reason for using the directory?
I agree it won't harm us at all but it might be worth considering to change the ConfigMap we document and use now to the directory _/etc/yunikorn_ which means that we can use the whole directory to provide multiple configs into the image.

yangwwei:
Hi @wilfred-s 

It probably should be fine, I checked some other projects under Apache 2.0 license like us, they use this option too. If we are concerning about the license, we can run some legal scans when we create the image with ` CGO_ENABLED=0`. BTW, where do you see the info about by enabling `cgo` means to get `libc` included in binary? 

@dcaceres88 could you please follow https://github.com/cloudera/yunikorn-core/blob/master/docs/how-to-contribute.md to sign and returns us an ICLA? We cannot merge the PR before that.

Thanks

wilfred-s:
After further internal discussion we have come to the conclusion that the CGO option is OK and it should not affect the images and licensing when we build. Approving that part of the change: we should set it as a default.

For the volume change that part is not needed as it is already fixed, can you please update the pull request accordingly?

dcaceres88:
Hi guys, 

@yangwwei , Today I sent the agreement. 

About the CGO_ENABLED, I agree with all of you about the solution. 
On the other hand, about the volumeMounts,  I have one question.  (https://github.com/cloudera/yunikorn-k8shim/commit/5a799d31d08e628e6e1a253700fd854a8e23eb2d) was PR that @yangwwei done to fix scheduler-load.yaml inside /deployments/scheduler directory. But, we use today this file to make a fresh deployment and failed: 
`FATAL    shim/shimscheduler.go:94    failed to register to yunikorn-core    {"error": "open queues.yaml: no such file or directory"}`. 
Instead, if we add volumeMounts parameter ( like I explained in previous comments) the deploy is working. Can you review this point?, because I think that it's necessary add volumeMounts parameter to scheduler-load.yaml file (https://github.com/cloudera/yunikorn-k8shim/blob/master/deployments/scheduler/scheduler-load.yaml).
Sorry for my insistence on this parameter, but our deployments didn't work without this change. Maybe is not the best approach to resolve the problem, but at the moment I don´t know any other way to do that. 

Thanks.

wilfred-s:
Can you please show the output of the command: `kubectl describe configmaps yunikorn-configs`

The config map should be created for this to work. I have deployed multiple cluster over the last 2 days and it has worked in all deployments (docker desktop, minikube and AWS EKS)

dcaceres88:
The output of "`kubectl describe configmaps yunikorn-configs`" command is:
```
Data
====
queues.yaml:
----
partitions:
  -
    name: default
    queues:
      -
        name: root
        queues:
          -
            name: search
            resources:
              guaranteed:
                memory: 400000
                vcore: 40000
              max:
                memory: 600000
                vcore: 60000
          -
            name: sandbox
            resources:
              guaranteed:
                memory: 100000
                vcore: 10000
              max:
                memory: 100000
                vcore: 10000
Events:  <none>
```

We followed all the steps in the documentation, first create ConfigMap, then apply the scheduler-load.yaml and RBAC.  

My colleague @maprietom created fresh deployment over minikube and had the same problem as me. 

Are you using scheduler.yaml
(https://github.com/cloudera/yunikorn-k8shim/blob/master/deployments/scheduler/scheduler.yaml)
 or scheduler-load.yaml
(https://github.com/cloudera/yunikorn-k8shim/blob/master/deployments/scheduler/scheduler-load.yaml) for your deployments?. Because scheduler.yaml have volumeMounts parameter, but scheduler-load.yaml no. If we use  scheduler.yaml, we don't have load balancer but the deployment works ok. 


Thanks

yangwwei:
Thanks @dcaceres88 , @wilfred-s . You are correct, the changes for volume mounts are necessary, I don't know why I missed that in my last commit. I should have tried both yaml files, ah, can't recall what happened. Anyway, thanks for providing the fix @dcaceres88 , and thanks for the review @wilfred-s. Let me merge this.

PR #16, opened 2019-07-26
Updated go.sum with new checksum for gotest.tools v2.2.0+

TarunParimi:
Trivial change of updating the checksum to make the build work on go v1.11.4+ https://github.com/golang/go/issues/29278


yangwwei:
We don't need this anymore, @TarunParimi helped to provide a workaround in #18. Closing now.

PR #18, opened 2019-07-29
Added a note in README to handle checksum mismatch

TarunParimi:


README.md
```diff
@@ -19,6 +19,7 @@ This command will build a binary `k8s_yunikorn_scheduler` under `_output/bin` di
 Run `./k8s_yunikorn_scheduler -help` to see all options.
 
 **Note**: it may take few minutes to run this command for the first time, because it needs to download all dependencies.
+**Note**: In case you get an error relating to `checksum mismatch`. Run `go clean -modcache` and then rerun `make build`.
```
yangwwei: this is a single line with 2 **Note**
instead, can we just remove the second note, I think this is fine with a single Node, two sentences.

Issue #19, opened 2019-08-05
scheduler state recovery - shim side changes


yangwwei:
this issue tracks the efforts of making scheduler be able to recover its state upon restart

yangwwei:
this is fixed in #20

PR #20, opened 2019-08-05
scheduler state recovery - shim side changes

yangwwei:
* rename node-controller to schedulerNodes, which manages nodes and their status from shim side
 * add state machine to node, adding recovering state
 * add ability to let shim scheduler wait for nodes recovering
 * some improvements to test classes, such as mock for kubeClient, nodeLister and schedulerApi etc


yangwwei:
@wilfred-s pls help to review this patch when you get time. Aplogoligies it contains quite a lot of line changes, some of them are caused by class-refactoring and small improvements. I really should have a separate patch for that. I'll pay attention next time. Thx

pkg/cache/node_events.go
```diff
@@ -14,27 +14,24 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package client
+package cache
```
wilfred-s: This change looks weird: I think it is github folding a delete and add into one change.

pkg/cache/nodes.go
```diff
@@ -51,8 +70,35 @@ func equals(n1 *v1.Node, n2 *v1.Node) bool {
...
+					NodeId:           pod.Spec.NodeName,
+					ApplicationId:    appId,
+					Partition:        common.DefaultPartition,
+				})
+			}
+		} else {
+			return err
+		}
+	}
+	return fmt.Errorf("cannot existing allocation because pod %s is not assigned to a node", pod.Name)
```
wilfred-s: The text looks like it is missing the word "add"

pkg/cache/nodes.go
```diff
@@ -63,15 +109,21 @@ func (nc *NodeController) addNode(obj interface{}) {
...
+	// once node is added to scheduler, first thing is to recover its state
+	dispatcher.Dispatch(CachedSchedulerNodeEvent{
+		NodeId:    newNode.name,
+		Event:     events.RecoverNode,
+	})
 }
 
-func (nc *NodeController) updateNode(oldObj, newObj interface{}) {
+func (nc *schedulerNodes) updateNode(oldObj, newObj interface{}) {
+	nc.lock.Lock()
```
wilfred-s: Can we move the locking down to the point were we need it to make the lock cycle as short as possible?
Just before we call UpdateNode()

pkg/cache/nodes.go
```diff
@@ -51,8 +70,35 @@ func equals(n1 *v1.Node, n2 *v1.Node) bool {
...
+			}
+		} else {
+			return err
+		}
+	}
+	return fmt.Errorf("cannot existing allocation because pod %s is not assigned to a node", pod.Name)
+}
+
+func (nc *schedulerNodes) addNode(obj interface{}) {
+	nc.lock.Lock()
```
wilfred-s: Can we move the locking down to the point were we need it to make the lock cycle as short as possible?
Just before we call AddNode()

pkg/cache/nodes.go
```diff
@@ -111,8 +162,10 @@ func (nc *NodeController) updateNode(oldObj, newObj interface{}) {
 	}
 }
 
-func (nc *NodeController) deleteNode(obj interface{}) {
-	log.Logger.Debug("node-controller: DeleteNode")
+func (nc *schedulerNodes) deleteNode(obj interface{}) {
+	nc.lock.Lock()
```
wilfred-s: Can we move the locking down to the point were we need it to make the lock cycle as short as possible?
Just before we call RemoveNode()

pkg/cache/node.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (n *SchedulerNode) addExistingAllocation(allocation *si.Allocation) {
+	n.lock.Lock()
+	defer n.lock.Unlock()
+	n.existingAllocations = append(n.existingAllocations, allocation)
+}
+
+func (n *SchedulerNode) getNodeState() string {
+	n.lock.RLock()
```
wilfred-s: According to the doc fsm has its own internal locking and we do not need to guard here.

pkg/cache/node.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+}
+
+func (n *SchedulerNode) handle(ev events.SchedulerNodeEvent) error {
+	log.Logger.Debug("scheduler node state transition",
+		zap.String("nodeId", ev.GetNodeId()),
+		zap.String("preState", n.fsm.Current()),
+		zap.String("pendingEvent", string(ev.GetEvent())))
+	if err := n.fsm.Event(string(ev.GetEvent()), ev.GetArgs()...); err != nil {
+		return err
```
wilfred-s: fsm can return an error if there is a transition that does not change the state. Healthy to Healthy i.e. err != nil does not mean the transition really failed:
```
	// handle the same state transition not nil error (limit of fsm).
	if err.Error() == "no transition" {
		return nil
	}
```

pkg/cache/context_recovery.go
```diff
@@ -0,0 +1,69 @@
...
+			zap.Int("numOfRecoveredNodes", numOfHealthyNodes))
+
+		if time.Now().After(deadline) {
+			log.Logger.Warn("timeout waiting for exiting recovery state",
+				zap.String("timeout", maxTimeout.String()))
+			return fmt.Errorf(fmt.Sprintf("timeout waiting for exiting recovery state, timeout=%s",
+				maxTimeout.String()))
+		}
+
+		time.Sleep(time.Duration(1) * time.Second)
```
wilfred-s: This can be simplified to just time.Second

pkg/cache/context_recovery_test.go
```diff
@@ -0,0 +1,188 @@
...
+	sn1 := context.nodes.getNode("host0001")
+	sn2 := context.nodes.getNode("host0002")
+
+	assert.Assert(t, sn1 != nil)
+	assert.Assert(t, sn2 != nil)
+
+	if err := utils.WaitForCondition(func() bool {
+		return sn1.getNodeState() == string(events.States().Node.Recovering) &&
+			sn2.getNodeState() == string(events.States().Node.Recovering)
+	}, time.Duration(1*time.Second), time.Duration(5*time.Second)); err != nil {
```
wilfred-s: This can be simplified to:
```
}, time.Second, 5*time.Second); err != nil {
```

pkg/cache/context_recovery_test.go
```diff
@@ -0,0 +1,188 @@
...
```
wilfred-s: duration simplification here also

pkg/cache/context_recovery_test.go
```diff
@@ -0,0 +1,188 @@
...
+		t.Logf("context stays waiting for recovery, error: %v", err)
+	}
+
+	// node2 recovery is done
+	dispatcher.Dispatch(CachedSchedulerNodeEvent{
+		NodeId:    "host0002",
+		Event:     events.NodeAccepted,
+	})
+
+	if err := context.waitForRecovery(nodeLister, time.Duration(3 * time.Second)); err != nil {
```
wilfred-s: duration simplification here also

pkg/cache/context_recovery_test.go
```diff
@@ -0,0 +1,188 @@
...
+	// node1 recovery is done
+	dispatcher.Dispatch(CachedSchedulerNodeEvent{
+		NodeId:    "host0001",
+		Event:     events.NodeAccepted,
+	})
+
+	nodeLister := test.NewNodeListerMock()
+	nodeLister.AddNode(&node1)
+	nodeLister.AddNode(&node2)
+	if err := context.waitForRecovery(nodeLister, time.Duration(3 * time.Second)); err == nil {
```
wilfred-s: duration simplification here also

pkg/cache/context_recovery_test.go
```diff
@@ -0,0 +1,188 @@
...
+		return sn1.getNodeState() == string(events.States().Node.Recovering) &&
+			sn2.getNodeState() == string(events.States().Node.Recovering)
+	}, time.Duration(1*time.Second), time.Duration(5*time.Second)); err != nil {
+		t.Fatalf("all nodes should under recoverying state")
+	}
+
+	nodeLister := test.NewNodeListerMock()
+	nodeLister.AddNode(&node1)
+	nodeLister.AddNode(&node2)
+	if err := context.waitForRecovery(nodeLister, time.Duration(3 * time.Second)); err == nil {
```
wilfred-s: duration simplification here also

pkg/cache/nodes_test.go
```diff
@@ -71,42 +77,55 @@ func TestAddNode(t *testing.T) {
...
-	nc.addNode(&newNode)
+	nodes.addNode(&newNode)
+
 	// values are verified in injected fn
 	// verify register is not called, update is called and just called once
-	assert.Equal(t, api.RegisterCount, 0)
-	assert.Equal(t, api.UpdateCount, 1)
+	if err := utils.WaitForCondition(func() bool {
+		return api.GetRegisterCount() == 0
+	}, time.Duration(1*time.Second), time.Duration(5*time.Second)); err != nil {
```
wilfred-s: duration simplification here also

pkg/cache/nodes_test.go
```diff
@@ -71,42 +77,55 @@ func TestAddNode(t *testing.T) {
...
+		t.Fatalf("%v", err)
+	}
+
+	if err := utils.WaitForCondition(func() bool {
+		return api.GetUpdateCount() == 1
+	}, time.Duration(1*time.Second), time.Duration(5*time.Second)); err != nil {
```
wilfred-s: duration simplification here also

pkg/cache/nodes_test.go
```diff
@@ -235,13 +259,24 @@ func TestDeleteNode(t *testing.T) {
...
-	nc.deleteNode(&node)
-	assert.Equal(t, api.RegisterCount, 0)
+	api.UpdateFunction(ignoreNodeUpdateFn)
+	nodes.addNode(&node)
+	nodes.deleteNode(&node)
+
+
+	if err := utils.WaitForCondition(func() bool {
+		return api.GetRegisterCount() == 0
+	}, time.Duration(1*time.Second), time.Duration(5*time.Second)); err != nil {
```
wilfred-s: duration simplification here also

pkg/cache/nodes_test.go
```diff
@@ -235,13 +259,24 @@ func TestDeleteNode(t *testing.T) {
...
+		t.Fatalf("%v", err)
+	}
+
 	// update should be called twice
 	// one for add, the other one for delete
-	assert.Equal(t, api.UpdateCount, 2)
+	if err := utils.WaitForCondition(func() bool {
+		return api.GetUpdateCount() == 2
+	}, time.Duration(1*time.Second), time.Duration(5*time.Second)); err != nil {
```
wilfred-s: duration simplification here also

pkg/shim/scheduler.go
```diff
@@ -141,19 +143,25 @@ func (ss *KubernetesShim) triggerSchedulerStateRecovery() func(e *fsm.Event) {
...
 	return func(e *fsm.Event) {
 		log.Logger.Info("recovering scheduler states...please wait...")
-		// TODO add recovery process
-
-		// success
-		dispatcher.Dispatch(ShimSchedulerEvent{
-			event: events.RecoverSchedulerSucceed,
-		})
+		// wait for recovery, max timeout 3 minutes
+		if err := ss.context.WaitForRecovery(time.Duration(3) * time.Minute); err != nil {
```
wilfred-s: duration simplification here also

pkg/cache/nodes.go
yangwwei: Yep, in next patch, this no longer returns an error and simply skips the pods that not assigned to a node.

pkg/cache/node.go
yangwwei: Yes. Make sense. I just removed the lock for node, and task. app and scheduler are fine.

pkg/cache/nodes.go
yangwwei: Yes, we can, but I don't think it's necessary. These few lines are not time-consuming operations, it can hardly impact the performance. But having them at the beginning, IMO it is more readable. When we do performance evaluations, I expect there will be a ton of places that can be optimized and such places can be ignored. Would that be OK?

yangwwei:
hi @wilfred-s 
follow-up commit 2 and 3 just uploaded, could you please help to review? Thx

pkg/cache/nodes.go
```diff
@@ -92,62 +91,45 @@ func (nc *schedulerNodes) addExistingAllocation(pod *v1.Pod) error {
...
+				})
+			}
+		}
+	}
 }
 
-func (nc *schedulerNodes) updateNode(oldObj, newObj interface{}) {
+func (nc *schedulerNodes) updateNode(oldNode, newNode *v1.Node) {
 	nc.lock.Lock()
 	defer nc.lock.Unlock()
```
wilfred-s: Why are we locking the schedulerNodes here?
The only thing we do is generate an event to pass on to the proxy, we're not updating the cache any more which I think removes the need for locking completely

oops read over the update call line, cancel this looks good now

pkg/cache/context_recovery.go
```diff
@@ -20,50 +20,194 @@ import (
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/listers/core/v1"
 	"time"
 )
 
 func (ctx *Context) WaitForRecovery(maxTimeout time.Duration) error {
-	return ctx.waitForRecovery(ctx.nodeInformer.Lister(), maxTimeout)
+	// currently, disable recovery when testing in mocked cluster,
+	// mock pod/node lister is not easy
```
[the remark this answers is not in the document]
yangwwei: Yes, just updated the comment in commit-4. Actually, we only disable recovery when testing with a mocked cluster for now. But we have separate UTs for waitForAppRecovery/waitForNodeRecovery.

yangwwei:
Hi @wilfred-s 

Sorry it took me 2 commits to fix the form for all time.Durations. Now they are all using simplest form, e.g `5*time.Second`. Pls take a look, thx!

pkg/cache/node_events.go
wilfred-s: the diff shows this correctly

wilfred-s:
this solves #19

PR #21, opened 2019-08-12
Interface update and cleanup (SI pull request #2)

wilfred-s:
Corresponding changes for the SI pull request #2 in which we update the interface spec


wilfred-s:
The use of new information will be in the placement rules changes. This is the plumbing that needed to happen to make sure we can leverage the new interface.

I left any non crucial changes out explicitly to keep this update as simple as possible and just make sure that the changed field names are used in the existing code and all tests pass.

wilfred-s:
I did not make this clear in the previous update: the AddApplicationRequest has had a queue from the initial commit of the code. The allocation also had a queue which has been removed. So there is no need to add the queue anywhere as the queue reference that is removed was never used.

yangwwei:
Thanks for the clarification, merging this now.

PR #22, opened 2019-08-13
Placement rules: add user and tag to shim application request

wilfred-s:
Adding the user and tag information to the shim
Currently the tags are limited to the namespace only.
The user information is the ServiceAccountName from the pod, no other user information is currently availble from the K8s side.


pkg/cache/context.go
```diff
@@ -555,11 +555,17 @@ func (ctx *Context) getOrCreateApplication(pod *v1.Pod) *Application {
...
 		return application
-	} else {
-		newApp := NewApplication(appId, utils.GetQueueNameFromPod(pod), ctx.schedulerApi)
-		ctx.applications[appId] = newApp
-		return ctx.applications[appId]
 	}
+	// create the tags for the application
+	// labels or annotations from the pod can be added when needed
+    tags := map[string]string{}
+    tags["namespace"] = pod.Namespace
```
yangwwei: Do we need some check for the value of `pod.Namespace` here?
According to the doc
```
Namespace defines the space within each name must be unique. An empty namespace is
equivalent to the "default" namespace, but "default" is the canonical representation.
Not all objects are required to be scoped to a namespace - the value of this field for
those objects will be empty.
```
this might be empty.

Issue #23, opened 2019-08-15
Recovering application event is missing user info


wilfred-s:
The event that is generated for a recovering application does not include the user information that is lifted out of a pod. On a new submit we pull the user info, we should do the same for consistency when we recover the pod.

PR #24, opened 2019-08-15
Add user and tag to shim application request for recovery

wilfred-s:
Add the missing user and tags infor to the event to add a recovered application to the scheduler core.


wilfred-s:
This fixes issue #23

PR #30, opened 2019-08-22
Placement rules for applications 

wilfred-s:
Add example for a namespace placement rule.
Fix for tests to work with PR35 from the core.

NOTE: yunikorn-core mod ref needs updating after PR35 is committed


wilfred-s:
force push fixed up the change, had a dangling local change that has been merged already via #24

deployments/examples/README.md
```diff
@@ -57,3 +57,10 @@ Use [minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/) as a wo
     <br>See [Dynamic Volume Provisioning](https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/) in the kubernetes docs.
   
 _NOTE_: dynamic provisioning can interfere with the existing volume examples given.
+
+## namespace
+The namespace example uses a placement rule and special queue configuration. The pod is a simple sleep pod which will be scheduled based on the namespace it runs in. The pod does not have a queue set.
+
+* create the config volume using the local [queues.yaml](./namespace/queues.yaml) file not the standard one.
+* add the development namespace  using [development](./namespace/development.yaml) file: `kubectl create -f development.yaml`
```
yangwwei: there is an extra space between "namespace" and "using"

deployments/examples/README.md
```diff
@@ -57,3 +57,10 @@ Use [minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/) as a wo
     <br>See [Dynamic Volume Provisioning](https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/) in the kubernetes docs.
   
 _NOTE_: dynamic provisioning can interfere with the existing volume examples given.
+
+## namespace
+The namespace example uses a placement rule and special queue configuration. The pod is a simple sleep pod which will be scheduled based on the namespace it runs in. The pod does not have a queue set.
+
+* create the config volume using the local [queues.yaml](./namespace/queues.yaml) file not the standard one.
```
yangwwei: suggest renaming "config volume" to "configmap", and add a link to https://github.com/cloudera/yunikorn-core/blob/master/docs/user-guide.md#create-the-configmap.

wilfred-s:
The EBS volume is the only one that is capitalised. All the list in other docs are lower case so that one should be changed (will change that independent of this PR)

The example that is added is an auto create example. That was part of the last checkin. There are no changes needed to any of the files to support the auto create of any namespace queue. It shows one pod (development) in an existing queue and one in a created queue (production).

If needed/wanted users could remove the fixed queues default & development as a cleanup. For the example I would like to keep it as is to show the difference and  the fact that the rule works in both cases.

yangwwei:
Sounds good, approving this. Thanks @wilfred-s !

Issue #31, opened 2019-08-23
State transitions triggered by events are incorrectly logged as error


wilfred-s:
The state machine we use reports back an error with the text _no transition_ if the event caused the state machine to stay in the same state as it was before the event was processed.
We currently log the error and process it as an error but the result really should be handled as a success.

yangwwei:
Hi @wilfred-s 
This doesn't sound like a bug in the state machine, this is probably indicating our state transition design has some issue. 

For example, when an app is already `submitted`, but we try to `submit`  it again, we may get this error. To fix it, we can simply add another transition

```
event: SubmitApplication
src: Submitted
dst: Submitted
```
do you have details when you saw such error?

wilfred-s:
It is not a bug in the state machine. It is documented that the state machine responds this way if the event does not trigger a state change **and** it did not fail to process.

Logging an error which is not an error is confusing, pollutes the logs and we should prevent that from happening.

This is from a run I did yesterday:
```
2019-08-22T07:27:20.071Z	DEBUG	cache/context.go:243	add pod	{"namespace": "development", "podName": "sleep1", "podUID": "46ae0480-c4ae-11e9-b999-9e0ec2ca0b8b", "state": "Pending"}
2019-08-22T07:27:20.205Z	INFO	shim/scheduler.go:219	scheduling	{"app": "application-namespace-1", "pendingTask": "sleep1"}
2019-08-22T07:27:20.205Z	DEBUG	cache/application.go:117	application state transition	{"appId": "application-namespace-1", "preState": "Running", "pendingEvent": "RunApplication"}
2019-08-22T07:27:20.205Z	ERROR	cache/context.go:627	failed to handle application event	{"event": "RunApplication", "error": "no transition"}
github.com/cloudera/yunikorn-k8shim/pkg/cache.(*Context).ApplicationEventHandler.func1
	/Users/wilfred/repos/yunikorn/src/github.com/cloudera/yunikorn-k8shim/pkg/cache/context.go:627
github.com/cloudera/yunikorn-k8shim/pkg/dispatcher.Start.func1
	/Users/wilfred/repos/yunikorn/src/github.com/cloudera/yunikorn-k8shim/pkg/dispatcher/dispatcher.go:130
```

yangwwei:
This is a `RunApplication` when app is already at "Running" state, interesting our code is already like

```
{Name: string(events.RunApplication),
  Src: []string{states.Accepted, states.Running},
[...]

PR #32, opened 2019-08-23
Handle "no transistion" error from state machine as a success and noop

wilfred-s:
This fixes issue #31 

Transparently handle a no transition event as a success


wilfred-s:
this closes #31

PR #34, opened 2019-08-28
make sure containers are released when on task failures

### LINKED ISSUE #29: Resource Usage not dropping after scaling down replicas in a deployment
This issue occurred while following https://github.com/cloudera/yunikorn-core/blob/master/docs/user-guide.md
I am using docker-for-desktop Kubernetes on my Mac 
I am scheduling an nginx service via yunikorn as mentioned in the guide. I am editing the nginx deployment to scale up the number of replicas to a high number which cannot be spawned due to unavailability of resources (some pods are pending). Now, when I edit the deployment again to scale down the number of replicas, the resource usage and number of running applications do not update on the yunikorn UI though k8s dashboard shows that the pods have been deleted.  Also on the UI, under the cluster info tab, the application and container counts do not seem to be getting updated in this scenario.

yangwwei:
this fixes #29 

Here is what happens, when the shim failed to bind a pod to the specified node, we marked the task as failed. However, we are not asking scheduler-core to release these allocations, this is causing the issue seeing in #29. So this PR contains the following changes
- make sure containers are released on failures
- added test case to cover this scenario
- minor improvements to the logs

### DIFF
--- a/pkg/shim/scheduler_mock.go
+++ b/pkg/shim/scheduler_mock.go
@@ -19,7 +19,7 @@ package main
 import (
 	"fmt"
 	"github.com/cloudera/yunikorn-core/pkg/api"
-	utils "github.com/cloudera/yunikorn-core/pkg/common/configs"
+	coreconfigs "github.com/cloudera/yunikorn-core/pkg/common/configs"
 	"github.com/cloudera/yunikorn-core/pkg/entrypoint"
 	"github.com/cloudera/yunikorn-core/pkg/log"
 	"github.com/cloudera/yunikorn-k8shim/pkg/cache"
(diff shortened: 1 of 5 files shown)


yangwwei:
@wilfred-s can you pls help to review this PR? thx

pkg/shim/scheduler_mock.go
```diff
@@ -21,13 +21,16 @@ import (
 	"github.com/cloudera/yunikorn-core/pkg/api"
 	utils "github.com/cloudera/yunikorn-core/pkg/common/configs"
 	"github.com/cloudera/yunikorn-core/pkg/entrypoint"
+	"github.com/cloudera/yunikorn-core/pkg/log"
 	"github.com/cloudera/yunikorn-k8shim/pkg/cache"
 	"github.com/cloudera/yunikorn-k8shim/pkg/callback"
 	"github.com/cloudera/yunikorn-k8shim/pkg/client"
 	"github.com/cloudera/yunikorn-k8shim/pkg/common"
 	"github.com/cloudera/yunikorn-k8shim/pkg/common/test"
+	utils2 "github.com/cloudera/yunikorn-k8shim/pkg/common/utils"
```
wilfred-s: Can we rename the utils import (line 22) and change then two references to that and keep this as utils?

yangwwei: Done

PR #38, opened 2019-08-29
helm-chart for yunikorn

sunilgovind:
Initial version of helm-chart


helm-charts/yunikorn-all/charts/yunikorn-k8s-shim/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
```
yangwwei: for now, let's use `0.1.0` for versions

helm-charts/yunikorn-all/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for YuniKorn
+name: yunikorn-all
```
yangwwei: can we rename `yunikorn-all` to `yunikorn`

helm-charts/yunikorn-all/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for YuniKorn
```
yangwwei: the description should just explain what this is. So suggest rephrasing it to something like "YuniKorn scheduler for Kubernetes"

helm-charts/yunikorn-all/charts/yunikorn-k8s-shim/templates/ingress.yaml
```diff
@@ -0,0 +1,38 @@
+{{- if .Values.ingress.enabled -}}
```
yangwwei: have you tested if this is working?
If not, I suggest to remove it? I don't think we need this for now since the default flag is also false.

helm-charts/yunikorn-all/charts/yunikorn-k8s-shim/values.yaml
```diff
@@ -0,0 +1,36 @@
...
+  annotations: {}
+    # kubernetes.io/ingress.class: nginx
+    # kubernetes.io/tls-acme: "true"
+  path: /
+  hosts:
+    - chart-example.local
+  tls: []
+  #  - secretName: chart-example-tls
+  #    hosts:
+  #      - chart-example.local
```
yangwwei: this can also be removed, otherwise, people might get confused

helm-charts/yunikorn-all/values.yaml
```diff
@@ -0,0 +1,48 @@
...
+  pullPolicy: IfNotPresent
+
+nameOverride: ""
+fullnameOverride: ""
+
+service:
+  type: ClusterIP
+  port: 80
+
+ingress:
```
yangwwei: can we remove this too

sunilgovind:
@yangwwei update latest one with comments fix. pls check once in ur local cluster as well.
 I tested and i can see all services and pods are up.

sunilgovind:
we wanted to helm-charts at top level as RE will be looking at his path. I ll check and see whether we can set custom path with RE

yangwwei:
> we wanted to helm-charts at top level as RE will be looking at his path. I'll check and see whether we can set custom path with RE

I am preferring to place this to a sub-folder, will leave it like this for now, and revisit once you get back from RE team. Thanks.

PR #39, opened 2019-08-29
Add configmap in helm chart to further simplify the deployment

yangwwei:
this PR does following things
- add configmap to the helm chart, so we don't need to create it manually
- make sure all label app value is "yunikorn"

### DIFF
--- a/helm-charts/yunikorn/templates/configmap.yaml
+++ b/helm-charts/yunikorn/templates/configmap.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: yunikorn-configs
+  labels:
+    app: yunikorn
+  annotations:
+    "helm.sh/hook": "pre-install"
+data:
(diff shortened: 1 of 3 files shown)


helm-charts/yunikorn/templates/configmap.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: yunikorn-configs
+  labels:
```
sunilgovind: Hi @yangwwei  
Under label, cud we add below items. So we will have same nomenclature across all yamls.
`    chart: {{ include "yunikorn.chart" . }}
    release: {{ .Release.Name }}
    heritage: {{ .Release.Service }}`

yangwwei: Hi @sunilgovind , it makes sense, I just added that in the second commit, please take a look.

sunilgovind:
Thanks @yangwwei  
Looks good to me. Lets get this in

Issue #46, opened 2019-09-25
Length of dispatcher's event channel should be configurable


yangwwei:
This should be exposed as a CLI option which can be configured during startup.

TaoYang526:
Hi, @yangwwei FYI, I think it's better to change default value of the event channel from 1024 to 1024*1024.

yangwwei:
Agreed. We had some discussion today and we agree to increase the default to a bigger one. 1024*1024 looks good. Maybe we can just increase it to this number now, leave the configuration change later. Would you like to create a PR for this?

TaoYang526:
Thanks, If you don't mind, I would like to create a PR to handle all requirements in this issue. I plan to add a CLI option named "eventChannelCapacity", then initialize the event channel of dispatcher when calling dispatcher.Start(configs), thoughts?

yangwwei:
Sure, that will be even better!

TaoYang526:
Thanks @yangwwei. I have attached PR #50 for this.

yangwwei:
this is resolved by #50

PR #48, opened 2019-10-10
Dispatcher events could be lost when event channel is full

TaoYang526:
Currently when event channel in dispatcher is full, events could be lost and an error like "failed to dispatch event, event channel is full or closed" can be found in the scheduler log.
To solve this problem, I suppose that new event should wait in Dispatcher#dispatch until it's enqueued into the channel, add a new test case for this scenario as well.

### DIFF
--- a/pkg/dispatcher/dispatcher.go
+++ b/pkg/dispatcher/dispatcher.go
@@ -17,7 +17,6 @@ limitations under the License.
 package dispatcher
 
 import (
-	"errors"
 	"fmt"
 	"github.com/cloudera/yunikorn-k8shim/pkg/common/events"
 	"github.com/cloudera/yunikorn-k8shim/pkg/conf"
@@ -41,8 +40,7 @@ const (
 )
 
 var (
-	ErrExcessAsyncDispatchLimit = errors.New("dispatcher exceeds async-dispatch limit")
-	AsyncDispatchLimit int32 = 10000
+	AsyncDispatchLimit int32
 	AsyncDispatchCheckInterval = 3 * time.Second
 	DispatchTimeout time.Duration
 	asyncDispatchCount int32 = 0
@@ -58,10 +56,11 @@ type Dispatcher struct {
 }
 
 func init() {
+	eventChannelCapacity := conf.GetSchedulerConf().EventChannelCapacity
 	once.Do(func() {
 		if dispatcher == nil {
 			dispatcher = &Dispatcher{
-				eventChan: make(chan events.SchedulingEvent, conf.GetSchedulerConf().EventChannelCapacity),
+				eventChan: make(chan events.SchedulingEvent, eventChannelCapacity),
 				handlers:  make(map[EventType]func(interface{})),
 				stopChan:  make(chan struct{}),
 				running:   atomic.Value{},
@@ -71,8 +70,13 @@ func init() {
 		}
 	})
 	DispatchTimeout = conf.GetSchedulerConf().DispatchTimeout
+	AsyncDispatchLimit = int32(eventChannelCapacity / 10)
+	if AsyncDispatchLimit < 10000 {
+		AsyncDispatchLimit = 10000
+	}
 	log.Logger.Info("Init dispatcher",
-		zap.Int("EventChannelCapacity", cap(dispatcher.eventChan)),
+		zap.Int("EventChannelCapacity", eventChannelCapacity),
+		zap.Int32("AsyncDispatchLimit", AsyncDispatchLimit),
 		zap.Float64("DispatchTimeoutInSeconds", DispatchTimeout.Seconds()))
 }
 
@@ -130,7 +134,7 @@ func (p *Dispatcher) asyncDispatch(event events.SchedulingEvent) {
 	log.Logger.Warn("event channel is full, transition to async-dispatch mode",
 		zap.Int32("asyncDispatchCount", count))
 	if count > AsyncDispatchLimit {
-		panic(ErrExcessAsyncDispatchLimit)
+		panic(fmt.Errorf("dispatcher exceeds async-dispatch limit"))
 	}
 	go func(beginTime time.Time) {
 		defer atomic.AddInt32(&asyncDispatchCount, -1)
--- a/pkg/dispatcher/dispatcher.go
+++ b/pkg/dispatcher/dispatcher.go
@@ -101,14 +101,16 @@ func (p *Dispatcher) dispatch(event events.SchedulingEvent) error {
 	if !p.isRunning() {
 		return fmt.Errorf("dispatcher is not running")
 	}
-
-	select {
-	case p.eventChan <- event:
-		return nil
-	default:
-		return fmt.Errorf("failed to dispatch event," +
-			" event channel is full or closed")
+	for p.isRunning() {
+		select {
+		case p.eventChan <- event:
+			return nil
+		case <-time.After(time.Second * 3):
+			log.Logger.Error("event channel is full, keep waiting",
+				zap.Int("channelLength", len(p.eventChan)))
+		}
 	}
+	return fmt.Errorf("dispatcher is stopped")
 }
 
 func (p *Dispatcher) drain() {
--- a/pkg/dispatcher/dispatcher.go
+++ b/pkg/dispatcher/dispatcher.go
@@ -17,6 +17,7 @@ limitations under the License.
 package dispatcher
 
 import (
+	"errors"
 	"fmt"
 	"github.com/cloudera/yunikorn-k8shim/pkg/common/events"
 	"github.com/cloudera/yunikorn-k8shim/pkg/conf"
@@ -39,7 +40,13 @@ const (
 	EventTypeScheduler
 )
 
-var asyncDispatchCheckInterval = 3 * time.Second
+var (
+	ErrExcessAsyncDispatchLimit = errors.New("dispatcher exceeds async-dispatch limit")
+	AsyncDispatchLimit int32 = 10000
+	AsyncDispatchCheckInterval = 3 * time.Second
+	DispatchTimeout time.Duration
+	asyncDispatchCount int32 = 0
+)
 
 // central dispatcher that dispatches scheduling events.
 type Dispatcher struct {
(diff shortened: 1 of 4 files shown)


pkg/dispatcher/dispatcher.go
```diff
@@ -100,14 +100,16 @@ func (p *Dispatcher) dispatch(event events.SchedulingEvent) error {
...
-	case p.eventChan <- event:
-		return nil
-	default:
-		return fmt.Errorf("failed to dispatch event," +
-			" event channel is full or closed")
+	for p.isRunning() {
+		select {
+		case p.eventChan <- event:
+			return nil
+		case <-time.After(time.Second * 3):
```
yangwwei: Should we run this in a goroutine?
if we can enqueue the event, then go for it,
otherwise, we launch a goroutine that tries to enqueue the event in every 3s, and try at most 5 minutes to die.

wilfred-s: We should not let it sit there and wait. If `dispatch` becomes blocking `Dispatch` will also become a blocking call. That will have a flow on effect and could block the whole system.

We need to be very careful about this.

TaoYang526:
@yangwwei  @wilfred-s. Thanks for the review. 
I had considered about the non-blocking solution, but afraid that it may break the handling sequence of events, which maybe harmful in some scenarios especially for the transform of state machines. I think scheduler is not in a normal state for this scenario when the event channel is full, it maybe hit a bottleneck which can be raised by large-scale cluster or potential bugs, in the blocking way, we can detect it from the error log then find out the cause. As for the length of event channel in dispatcher, @yangwwei has created #46 to make it configurable so that we can increase it for a large-scale cluster. Thoughts?

yangwwei:
Hi @TaoYang526 

The main concern about the blocking manner is:
If shim gets blocked on `Dispatch()` call for few seconds, it will probably hold the scheduler/node/app/task's lock, causing the rest of the opts also be hanging. However, if we change this to non-block with multiple retries, we give the chance for shim to catch up handing cached events. Is this a valid concern?

TaoYang526:
@yangwwei Yes, this is indeed a concern, thanks for the clarification.
As for non-blocking solution, I'm not sure if the sequence of events may need to be guaranteed in some scenarios, how about adding "nonBlocking" input parameter(default as false) or adding Dispatcher#asyncDispatch method for the scenarios which hold a lock and have no requirement about the sequence?

yangwwei:
Hi @TaoYang526 
As we are building the system with respect to the states and transitions, we do not have strict requirements on the order of events. If event A is not dispatched probably, event B that depends on the state transited after event A will not be triggered. So I think it should be fine with the non-blocking manner.

TaoYang526:
Thanks @yangwwei for the clarification! I will update this PR with non-blocking manner.

TaoYang526:
@yangwwei @wilfred-s  This PR has been updated, please help to review. Thanks.
Updates:
(1) Using non-blocking dispatch manner.
(2) Make dispatch timeout configurable.
(3) Add 2 test cases to test timeout and non-timeout scenarios when event channel is full.

pkg/dispatcher/dispatcher.go
```diff
@@ -100,16 +103,45 @@ func (p *Dispatcher) dispatch(event events.SchedulingEvent) error {
...
-		select {
-		case p.eventChan <- event:
-			return nil
-		case <-time.After(time.Second * 3):
-			log.Logger.Error("event channel is full, keep waiting",
-				zap.Int("channelLength", len(p.eventChan)))
-		}
+	select {
+	case p.eventChan <- event:
+		log.Logger.Info("dispatched", zap.Int("capacity", cap(p.eventChan)),
```
yangwwei: I think we can remove this logging, otherwise even debug it would be overwhelming

pkg/dispatcher/dispatcher.go
```diff
@@ -100,16 +103,45 @@ func (p *Dispatcher) dispatch(event events.SchedulingEvent) error {
...
+			zap.Int("length", len(p.eventChan)))
+		return nil
+	default:
+		log.Logger.Warn("event channel is full, transition to async-dispatch mode",
+			zap.Int("channelLength", len(p.eventChan)),
+			zap.String("event", fmt.Sprintf("%+v", event)))
+		p.asyncDispatch(event)
```
yangwwei: I think we need to set a limit here, to avoid this creating too many goroutines. Can we create a limit counter here, lets set a default value of 10,000. If num of goroutines goes beyond this value, then panic. This is just to protect some worst-case scenarios.

pkg/dispatcher/dispatcher.go
```diff
@@ -100,16 +103,45 @@ func (p *Dispatcher) dispatch(event events.SchedulingEvent) error {
...
+			case <-time.After(asyncDispatchCheckInterval):
+				elapseTime := time.Now().Sub(beginTime)
+				if elapseTime >= conf.GetSchedulerConf().DispatchTimeout {
+					log.Logger.Error("dispatch timeout",
+						zap.Float64("elapseSeconds", elapseTime.Seconds()),
+						zap.Int("channelLength", len(p.eventChan)),
+						zap.String("event", fmt.Sprintf("%+v", event)))
+					return
+				}
+				log.Logger.Warn("event channel is full, keep waiting...",
```
yangwwei: the logging info should contain the time to wait here, something like:
`zap.String("retryingInSeconds", asyncDispatchCheckInterval)`

pkg/dispatcher/dispatcher.go
TaoYang526: Sure, this is added for local testing, forgot to remove.

pkg/dispatcher/dispatcher.go
TaoYang526: Already done like this: `zap.Float64("elapseSeconds", elapseTime.Seconds())`.
Is it better to rename "elapseSeconds" as "retryingInSeconds" ?

wilfred-s: The elapse seconds that is logged shows the time since we have created the go routine, that is the correct thing to log. I do think that we need to log the `DispatchTimeout` but we should only log that on the start of the dispatcher as it cannot change while running (see previous remark on getting the config).

Also in this case: I don't think we need to know the channel length as it is still full otherwise we would not be here. Same for the event, type might be nice but we do not need the full event.

TaoYang526:
Hi, @yangwwei @wilfred-s 
Sorry for replying so late, all suggestions above make sense to me and will update this later, thanks for the review.

pkg/dispatcher/dispatcher.go
TaoYang526: Updates: 
(1) Set a limit for async-dispatch goroutines in Dispatcher#asyncDispatch.
(2) Remove channelLength and event fields from this log, add asyncDispatchCount field and move this log into Dispatcher#asyncDispatch.
(3) Add log in Dispatcher#init to print EventChannelCapacity and DispatchTimeoutInSeconds.

pkg/dispatcher/dispatcher.go
TaoYang526: Updates:
(1) Set beginTime before creating async-dispatch goroutines.
(2) Also remove 'event' and 'channelLength' fields from the log.

pkg/dispatcher/dispatcher.go
```diff
@@ -100,16 +103,47 @@ func (p *Dispatcher) dispatch(event events.SchedulingEvent) error {
...
+// it's only called when event channel is full.
+func (p *Dispatcher) asyncDispatch(event events.SchedulingEvent) {
+	go func() {
+		beginTime := time.Now()
+		for p.isRunning() {
+			select {
+			case p.eventChan <- event:
+				return
+			case <-time.After(asyncDispatchCheckInterval):
+				elapseTime := time.Now().Sub(beginTime)
```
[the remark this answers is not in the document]
TaoYang526: Updated.

pkg/dispatcher/dispatcher.go
```diff
@@ -100,16 +103,47 @@ func (p *Dispatcher) dispatch(event events.SchedulingEvent) error {
...
+				if elapseTime >= conf.GetSchedulerConf().DispatchTimeout {
```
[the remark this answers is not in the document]
TaoYang526: Updated.

TaoYang526:
@yangwwei @wilfred-s 
I have updated this PR, please help to review in
General updates(details in comments above):
(1) Limit async-dispatch goroutines.
(2) Update logs and time calculation.
(3) Improve test cases.

> Looking at the testing: how long does it run for?
> The default check interval is set to 3 seconds. Only the seconds test sets that lower. Does that mean the first test runs for a long time because the retry is set to 3 seconds? For the test it would also be good to see how many events get pushed into go routines.
Three test cases in this PR run for about 2.7 seconds in total(2.0s+0.6s+0.0s). Some variables like AsyncDispatchLimit,AsyncDispatchCheckInterval and DispatchTimeout in dispatcher.go can be updated in test cases and recovered in defer functions. 
The "asyncDispatchCount" variable has been created for the limit and can help to know how many events get pushed into go routines.

yangwwei:
Thanks, it looks all good to me know. @wilfred-s do you have any more comments with the latest change?

pkg/dispatcher/dispatcher.go
```diff
@@ -39,6 +40,14 @@ const (
 	EventTypeScheduler
 )
 
+var (
+	ErrExcessAsyncDispatchLimit = errors.New("dispatcher exceeds async-dispatch limit")
+	AsyncDispatchLimit int32 = 10000
```
wilfred-s: Can we use the simple `fmt.Errorf()` call in the panic directly. This does not make it any clearer.

I also think that a hard coded limit of 10,000 could be a bit low compared to the queue length of 1,000,000. Go routines are light weight and do not add a lot of overhead. We can change the channel size already. If the channel has been increased to 100,000 via the config does it make sense to have an async buffer of 10,000 beside it? I am not sure that it would really help, maybe we need to set it based on 10% of the queue length.

wilfred-s:
@TaoYang526 can you resolve the conflicts please? They are really minor and should be fixed by a simple rebase of your change.

pkg/dispatcher/dispatcher.go
TaoYang526: Thanks for these suggestions, make sense to me.

wilfred-s:
Thank you @TaoYang526 for the fix and @yangwwei for the review

TaoYang526:
Thanks @wilfred-s, @yangwwei for the review and commit.

PR #54, opened 2019-10-29
Deduplication of RunApplication events

TaoYang526:
Currently RunApplication events can be generated repeatedly in KubernetesShim#schedule, especially when scheduling interval is very short or there are a huge number of pending tasks. Since tasks should wait at Pending state until RunApplication events are handled by its application, there's a gap in which KubernetesShim#schedule should be called for several times.
Updates in this PR:
(1) Add New state as the initial state for tasks, tasks will be transitioned to Pending state after generating RunApplication events.
(2) Refactor KubernetesShim#schedule to avoid iterating tasks and printing scheduling-task log at irrelevant app states like New etc.
(3) Add UT to test deduplication of RunApplication events

Similarly, SubmitApplication events may be generated repeatedly but not handled by this PR, it does less harm than RunApplication events since events of applications are much less than events of tasks.  We can handle it later if necessary.

### DIFF
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -180,6 +180,10 @@ func (app *Application) GetApplicationState() string {
 	return app.sm.Current()
 }
 
+func (app *Application) GetNewTasks() []*Task {
+	return app.getTasks(events.States().Task.New)
+}
+
 func (app *Application) GetPendingTasks() []*Task {
 	return app.getTasks(events.States().Task.Pending)
 }
--- a/pkg/cache/context_test.go
+++ b/pkg/cache/context_test.go
@@ -56,13 +56,13 @@ func TestAddApplications(t *testing.T) {
 	assert.Equal(t, len(context.applications), 1)
 	assert.Assert(t, context.applications["app00001"] != nil)
 	assert.Equal(t, context.applications["app00001"].GetApplicationState(), events.States().Application.New)
-	assert.Equal(t, len(context.applications["app00001"].GetPendingTasks()), 0)
+	assert.Equal(t, len(context.applications["app00001"].GetNewTasks()), 0)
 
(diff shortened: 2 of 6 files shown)


TaoYang526:
Thanks @wilfred-s for the review.

> So instead of setting the state for a task and generating a NewRunApplication event we should generate an event to change the task to PENDING and when we do that generate the NewRunApplication event from the callback as we only need the ID for the application and we have the whole task.

I think the root cause of duplication is the event-driven, we can't guarantee events can be handled in time, this way may transfer from duplication of NewRunApplication events to duplication of new events, thoughts?

> The current solution also could leave tasks in the incorrect state and still generate the _NewRunApplication_ event as there is no error handling around setting the state and we always generate the _NewRunApplication_ event even if the task did not transition.

In current solution, _NewRunApplication_ events are only generated for tasks in _New_ state, and before that tasks already transition to _Pending_ state, in which scenario tasks maybe leaved in incorrect state, can you please elaborate?

yangwwei:
@wilfred-s was in-theory correct; @TaoYang526 was in-practice correct. 

Our design is to let all actions triggered by events, with proper state transitions. So explicitly set state is not a fix good enough.

The problem here is in each scheduling interval (could be 0s), shim scans all pending tasks from all apps and trigger a `RunApplicationEven` for each of the pending task. When there are lots of tasks, we got too many dup events (because a task transit from pending to submitted/running might be slow).

To fix this, I think we should not scan all tasks in `scheduler#schedule()`, just apps. And move task iteration to `app#handleRunApplicationEvent`. This will help to reduce the event num signigicantly.

Let me upload a patch for this.

yangwwei:
@wilfred-s, @TaoYang526 please take a look if https://github.com/cloudera/yunikorn-k8shim/pull/56 can resolve this issue, thanks!

yangwwei:
This is resolved now by #57

PR #56, opened 2019-10-30
perf improvement: reduce num of RunApplicationEvent

yangwwei:
For #54


pkg/cache/application.go
```diff
@@ -258,17 +259,16 @@ func (app *Application) handleRecoverApplicationEvent(event *fsm.Event) {
...
-		return
-	}
-
-	switch t := event.Args[0].(type) {
-	case *Task:
-		dispatcher.Dispatch(NewSubmitTaskEvent(app.applicationId, t.taskId))
+	pendingTasks := app.getTasks(events.States().Task.Pending)
+	log.Logger.Debug("scheduling pending tasks",
+		zap.String("appID", app.applicationId),
+		zap.Int("numOfPendingTasks", len(pendingTasks)))
```
wilfred-s: minor: should we add a check for pendingTasks before this?
```
if len(pendingTasks == 0) {
    return
}
log.logger.Debug("scheduling pending tasks")
```
removes the 0 tasks scheduled message from the logs

TaoYang526:
Thanks @yangwwei for this new solution.
I agree that it's much better to move task iteration into app#handleRunApplicationEvent, but there is still a chance to generate too many dup NewSubmitTask events, I have met too many dup events in dispatcher which slow down the scheduler when submitting an application with thousands tasks, current solution may relieve the stress but not cut off this problem.
How about adding a field like 'newPending' for task which is true at first and set to false after sending NewSubmitTask event, or could you please share some better ideas?

yangwwei:
Sorry guys, I am distracted by some urgent issue today, will get back to this after that.

wilfred-s:
Thank you for the feedback @TaoYang526.

Based on the comment I think we need a fix that takes the new state you introduced in #54 and the way we handle moving the state and iterating from this solution. You and Weiwei have looked at it in more detail than I have. Because of that I am not going to write a new solution. Can you two work together and come up with a solution that takes both into account and solves the issue.

I am happy to review and help out in whatever way needed.

yangwwei:
Please ignore commit 2, I think I just have a better idea : )

yangwwei:
Closing this one, moving squashed changes to #57 .

PR #57, opened 2019-11-07
Perf improvement: refine scheduling logic in shim to reduce num of events

yangwwei:
This is a replacement for https://github.com/cloudera/yunikorn-k8shim/pull/56. Refined the scheduling logic to reduce the num of events generated in each scheduler interval.

### DIFF
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -233,7 +233,7 @@ func (app *Application) Schedule() {
 		}
 	case states.Running:
 		if len(app.GetNewTasks()) > 0 {
-			for _, task := range app.getTasks(events.States().Task.New) {
+			for _, task := range app.GetNewTasks() {
 				// note, if we directly trigger submit task event, it may spawn too many duplicate
 				// events, because a task might be submitted multiple times before its state transits to PENDING.
 				if err := task.handle(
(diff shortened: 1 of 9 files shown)


yangwwei:
@TaoYang526 pls help to review this patch. See if this approach resolves the issue. Thanks!

TaoYang526:
Thanks @yangwwei for updating this PR.
It's an awesome idea to let task handle state transition by itself, and it has clearly solved the problem. Good job!

pkg/cache/application.go
```diff
@@ -203,6 +210,48 @@ func (app *Application) getTasks(state string) []*Task {
...
+		}
+	case states.Accepted:
+		ev := NewRunApplicationEvent(app.GetApplicationId())
+		if err := app.handle(ev); err != nil {
+			log.Logger.Warn("failed to handle RUN app event",
+				zap.Error(err))
+		}
+	case states.Running:
+		if len(app.GetNewTasks()) > 0 {
+			for _, task := range app.getTasks(events.States().Task.New) {
```
TaoYang526: ApplicationInfo#getTasks will go through taskMap without readLock, may conflict when another thread is updating taskMap. I think we can query new tasks once like this:
```
newTasks := app.GetNewTasks()
if len(newTasks) > 0 {
   for _, task := range newTasks {
   ...
   }
}
```

yangwwei: yeah, that's correct. let me fix that. thx

yangwwei:
Thanks for the review @TaoYang526 , I have just merged this.

PR #58, opened 2019-11-09
embedded admission controller into helm chart deployment

yangwwei:
This patch is to simplify the deployment of the admission controller. Before this patch, we need to manually deploy the admission controller after scheduler is installed. After patch, all can be done by helm chart. And this patch does following:
- Pack the init scripts (generating certs, secrets etc) into the scheduler docker image
- In helm chart, added a property: `embedAdmissionController`, if true, the admission controller init hook will be triggered during the deployment of the scheduler, as a `postStart` hook.
- When we delete scheduler resources, it automatically deletes resources of the admission controller, including web-server, secrets, deployment, and service. This is done via `preStop` hook.
- If `embedAdmissionController` is false, no admission controller will be installed (the old behavior)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,6 +98,8 @@ scheduler: init
 sched_image: scheduler
 	@echo "building scheduler docker image"
 	@cp ${RELEASE_BIN_DIR}/${BINARY} ./deployments/image/configmap
+	@mkdir -p deployments/image/configmap/admission-controller-init-scripts
+	@cp ./deployments/admission-controllers/schedulername-mutation/*  deployments/image/configmap/admission-controller-init-scripts/
 	@sed -i'.bkp' 's/clusterVersion=.*"/clusterVersion=${VERSION}"/' deployments/image/configmap/Dockerfile
 	@coreSHA=$$(go list -m "github.com/cloudera/yunikorn-core" | cut -d "-" -f4) ; \
 	siSHA=$$(go list -m "github.com/cloudera/yunikorn-scheduler-interface" | cut -d "-" -f5) ; \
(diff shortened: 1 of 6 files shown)


yangwwei:
hi @wilfred-s  could you help to review the changes when you get a chance? Thanks!

Makefile
```diff
@@ -98,6 +98,8 @@ scheduler: init
 sched_image: scheduler
 	@echo "building scheduler docker image"
 	@cp ${RELEASE_BIN_DIR}/${BINARY} ./deployments/image/configmap
+	@mkdir -p deployments/image/configmap/admission-controller-init-scripts
+	@cp ./deployments/admission-controllers/schedulername-mutation/*  deployments/image/configmap/admission-controller-init-scripts/
```
wilfred-s: We need to be consistent with the paths being prefaced with the `.` even when relative.
I think we get into problems when we don't do that. This is not just from this change but it has slipped in

Makefile
```diff
@@ -110,6 +112,7 @@ sched_image: scheduler
 	--label "Version=${VERSION}"
 	@mv -f deployments/image/configmap/Dockerfile.bkp deployments/image/configmap/Dockerfile
 	@rm -f ./deployments/image/configmap/${BINARY}
+	@rm -rf ./deployments/image/configmap/admission-controller-init-scripts/
```
wilfred-s: same for this point: consistency with the `.`

deployments/admission-controllers/schedulername-mutation/admission_util.sh
```diff
@@ -34,6 +34,27 @@ delete_resources() {
...
+  # depedency check
+  command -v kubectl &> /dev/null
+  if [ $? -ne 0 ]; then
+    echo "dependency check failed: kubectl is not installed"
+    exit 1
+  fi
+
+  command -v openssl &> /dev/null
+  if [ $? -ne 0 ]; then
+    echo "dependency check failed: jq is not installed"
```
wilfred-s: the echo says `jq` should be `openssl`

helm-charts/yunikorn/values.yaml
```diff
@@ -20,6 +20,8 @@ service:
   port: 9080
   port_web: 9889
 
+embedAdmissionController: true
```
wilfred-s: we need to add a comment to the effect that it has

yangwwei:
Hi @wilfred-s , thanks, I've done the fixes in 2nd commit.

PR #59, opened 2019-11-16
Fix the caBundle issue in the admission controller in helm chart

yangwwei:
During the test, I found when deploying the scheduler + admission controller using the helm chart. The webhook server log prints some "TLS handshake error". This was because of the `caBundle` value set for the webserver was missing. Kubectl is not able to retrieve cluster context using in cluster configuration (running inside of the pod). The fix is to change a way of retrieving the ca.

The fix works for both helm chart and local deployment.


yangwwei:
hi @wilfred-s  pls help to review this, thx!

PR #60, opened 2019-11-16
simplify the deployment and add some sample yaml

yangwwei:
This PR contains 2 changes

1. use a simpler config for default deployment, that has placement rule enabled, no need to set queues explicitly.
2. add some sample yamls for admission-controller tests


yangwwei:
Thanks @wilfred-s for helping to fix the comments! Merging this now.

Issue #62, opened 2019-12-10
yk-predicates should be configurable


yangwwei:
Right now we only provide a pre-defined list of predicates
we should support configurable list of predicates

yangwwei:
this is resolved by #64

PR #64, opened 2019-12-12
Support configuring predicates via command-line flag

TaoYang526:
This is initial PR for #62 .
@yangwwei @wilfred-s , please help to review. Thanks.

### DIFF
--- a/pkg/conf/schedulerconf.go
+++ b/pkg/conf/schedulerconf.go
@@ -54,7 +54,8 @@ type SchedulerConf struct {
 	EventChannelCapacity int        `json:"eventChannelCapacity"`
 	DispatchTimeout   time.Duration `json:"dispatchTimeout"`
 	KubeQPS           int           `json:"kubeQPS"`
-	KubeBurst         int           `json:"KubeBurst"`
+	KubeBurst         int           `json:"kubeBurst"`
+	Predicates        string        `json:"predicates"`
 }
 
 func GetSchedulerConf() *SchedulerConf {
@@ -98,6 +99,8 @@ func init() {
 		"the maximum QPS to kubernetes master from this client")
 	kubeBurst := flag.Int("kubeBurst", DefaultKubeBurst,
 		"the maximum burst for throttle to kubernetes master from this client")
+	predicates := flag.String("predicates", "",
+		"comma-separated list of predicates")
 
 	// logging options
 	logLevel := flag.Int("logLevel", DefaultLoggingLevel,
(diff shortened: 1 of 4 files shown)


yangwwei:
Hi @TaoYang526 

This is what we wanted. Just one thing we need to make predicates options available from CLI. Usually, we can do

`make build`

this builds the binary under `./_output/bin/k8s_yunikorn_scheduler`, then we can run `./_output/bin/k8s_yunikorn_scheduler -h`, this returns us the available options:

```
Usage of ./_output/bin/k8s_yunikorn_scheduler:
  -clusterId string
    	cluster id (default "my-kube-cluster")
  -clusterVersion string
    	cluster version (default "0.1")
  -dispatchTimeout duration
    	timeout in seconds when dispatching an event (default 5m0s)
  -eventChannelCapacity int
    	event channel capacity of dispatcher (default 1048576)
  -interval duration
    	scheduling interval in seconds (default 1s)
  -kubeBurst int
    	the maximum burst for throttle to kubernetes master from this client (default 1000)
  -kubeConfig string
    	absolute path to the kubeconfig file
  -kubeQPS int
    	the maximum QPS to kubernetes master from this client (default 1000)
  -logEncoding string
    	log encoding, json or console. (default "console")
  -logFile string
    	absolute log file path
  -logLevel int
    	logging level, available range [-1, 5], from DEBUG to FATAL.
  -name string
    	name of the scheduler (default "yunikorn")
  -policyGroup string
    	policy group (default "queues")
  -volumeBindTimeout duration
    	timeout in seconds when binding a volume (default 10s)
```

We need to add predicates options in the help content. So people know what to add if they want.

yangwwei:
And looks like we will need to put all values from https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/algorithm/predicates/predicates.go here.

TaoYang526:
How about to update the predicates flag and fatal log as follows?
```
predicates := flag.String("predicates", "", "comma-separated list of predicates, "+
		"which will be validated as per the definition of predicatesOrdering in "+
		"https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/algorithm/predicates/predicates.go, "+
		"the scheduler will exit if any invalid predicates exist.")

msg := "please refer to the definition of predicatesOrdering in https://" +
					"github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/algorithm/predicates/predicates.go " +
					"for valid predicates."
return nil, fmt.Errorf("configured predicate is invalid: %s, message: %s", parsedPredicate, msg)

```
If the value of predicates option contains invalid predicate, the pod of yunikorn-scheduler will be in the Error state and users can see the fatal logs via executing 'kubectl logs <YK-POD-ID> yunikorn-scheduler-k8s'.

TaoYang526:
Thanks @yangwwei for the suggestion about directly showing all valid predicates, updated now.

yangwwei:
LGTM, thanks @TaoYang526

PR #67, opened 2019-12-16
Update docke image file for admission control

sunilgovind:
Update docke image file for admission control changes


yangwwei:
Thank you @sunilgovind for working on this.
Like @wilfred-s mentioned in the comment, we need to update all places. otherwise, our deployment yamls, helm charts will not work.

sunilgovind:
@wilfred-s  @yangwwei 
POD_ADMISSION_CONTROLLER_BINARY name is still same.
I only changed the image name for easier identification. I think may be its fine for now to keep it this way.
Pls check latest change set and let me know if any issues.

Issue #66, opened 2019-12-16
Failed to use persistent volume


TaoYang526:
I have tested launching pod with PVC based on YK and found that although the pod can be bound successfully but unable to mount volumes.
Error message:
```
  Warning     FailedMount     1s    kubelet, node-xxx  Unable to mount volumes for pod "pod-xxx": timeout expired waiting for volumes to attach or mount for pod...
```
I can see that CheckVolumeBinding predicate checks volumes via calling volumeBinder#FindPodVolumes, but can't find where to bind pod volumes as default scheduler of K8s did in Scheduler#bindVolumes before binding pod with a node. 
I'm not sure if this is in the schedule or just something missed, @wilfred-s  @yangwwei could you please take a look at this issue? Thanks.

wilfred-s:
How did you create the PVC what kind of volume are you referencing and what are the setting for the PVC and volume?
I have tested PVC and volume mounting and it worked for me but I have seen failures when the volume is not in the zone or when auto creating a volume took a long time.

yangwwei:
I also have some doubts about missing calling bind volumes, this is actually one problem I've been looking at when we do the integration with our internal products. I have a patch for this, @TaoYang526  could you please help to check if the changes in PR #68 works for you?

TaoYang526:
Thanks @wilfred-s  @yangwwei for your attentions.
> How did you create the PVC what kind of volume are you referencing and what are the setting for the PVC and volume?

Some details of my test are as follows:
(1) Before starting this test, we have serveral local PVs configured like this:
```
NAME                CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS      CLAIM                         STORAGECLASS   REASON   AGE
local-pv-1506073c   3519Gi     RWO            Delete           Available                                 fast-disks              92d
local-pv-3f58e927   3519Gi     RWO            Delete           Available                                 fast-disks              52d
local-pv-53c7f822   3519Gi     RWO            Delete           Available                                 fast-disks              19h
...
```
(2) Launch a test job which will request a pod with one volume configured with `persistentVolumeClaim.claimName=pv-job-14c2ce8f-bd86-4a6f-a34a-jobmanager-jobmanager` and the PVC description like this:
```
Name:          pv-job-14c2ce8f-bd86-4a6f-a34a-jobmanager-jobmanager
Namespace:     blink
StorageClass:  fast-disks
Status:        Pending
Volume:
Labels:        <none>
Annotations:   <none>
[...]

PR #68, opened 2019-12-16
Fix PV/PVC issue

yangwwei:
this PR explicitly calls assumeVolume/bindVolume/forgetVolume in the shim.

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -539,16 +539,16 @@ func (ctx *Context) AssumePod(name string, node string) error {
 			if _, err := ctx.volumeBinder.Binder.AssumePodVolumes(pod, node); err != nil {
 				return err
 			}
+		}
 
-			// when add assumed pod, we make a copy of the pod to avoid
-			// modifying its original reference. otherwise, it may have
-			// race when some other go-routines accessing it in parallel.
-			assumedPod := pod.DeepCopy()
-			// assign the node name for pod
-			assumedPod.Spec.NodeName = node
-			if targetNode := ctx.schedulerCache.GetNode(node); targetNode != nil {
-				return ctx.schedulerCache.AssumePod(assumedPod)
-			}
+		// when add assumed pod, we make a copy of the pod to avoid
+		// modifying its original reference. otherwise, it may have
+		// race when some other go-routines accessing it in parallel.
+		assumedPod := pod.DeepCopy()
+		// assign the node name for pod
+		assumedPod.Spec.NodeName = node
+		if targetNode := ctx.schedulerCache.GetNode(node); targetNode != nil {
+			return ctx.schedulerCache.AssumePod(assumedPod)
 		}
 	}
 	return nil
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -251,7 +251,6 @@ func (ctx *Context) addPod(obj interface{}) {
 		return
 	}
 
-
 	if pod.Status.Phase == v1.PodPending {
 		log.Logger.Debug("add pod",
 			zap.String("namespace", pod.Namespace),
@@ -568,20 +567,28 @@ func (ctx *Context) AssumePod(name string, node string) error {
 		// when add assumed pod, we make a copy of the pod to avoid
 		// modifying its original reference. otherwise, it may have
 		// race when some other go-routines accessing it in parallel.
-		assumedPod := pod.DeepCopy()
-		// assume pod volumes, this will update bindings info in cache
-		if ctx.volumeBinder != nil {
+		if targetNode := ctx.schedulerCache.GetNode(node); targetNode != nil {
+			assumedPod := pod.DeepCopy()
+			// assume pod volumes, this will update bindings info in cache
 			// assume pod volumes before assuming the pod
 			// this will update scheduler cache with essential PV/PVC binding info
-			if _, err := ctx.volumeBinder.Binder.AssumePodVolumes(assumedPod, node); err != nil {
+			assumePodVolumesIfNecessary := func(pod *v1.Pod, node string) (bool, error) {
+				// volume builder might be null in UTs
+				if ctx.volumeBinder != nil {
+					return ctx.volumeBinder.Binder.AssumePodVolumes(pod, node)
+				}
+				// if volume binder not exist (in UTs),
+				// we just assume pods volumes are all bound,
+				// so we simply ignore binding operations
+				return true, nil
+			}
+			allBound, err := assumePodVolumesIfNecessary(assumedPod, node)
+			if err != nil {
 				return err
 			}
-		}
-
-		// assign the node name for pod
-		assumedPod.Spec.NodeName = node
-		if targetNode := ctx.schedulerCache.GetNode(node); targetNode != nil {
-			return ctx.schedulerCache.AssumePod(assumedPod)
+			// assign the node name for pod
+			assumedPod.Spec.NodeName = node
+			return ctx.schedulerCache.AssumePod(assumedPod, allBound)
 		}
 	}
 	return nil
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -535,7 +535,7 @@ func (ctx *Context) BindPodVolumes(pod *v1.Pod) error {
 		// However, the assumePodVolumes was done in scheduler-core, because these assumed pods are cached
 		// during scheduling process as they have directly impact to other scheduling processes.
 		// The actual binding happens in the shim right before binding to a node.
-		allBound, err := ctx.volumeBinder.Binder.AssumePodVolumes(assumedPod, pod.Spec.NodeName)
+		allBound, err := ctx.volumeBinder.Binder.AssumePodVolumes(assumedPod, assumedPod.Spec.NodeName)
 		if err != nil {
 			log.Logger.Debug("AssumePodVolumes failed",
 				zap.String("podName", pod.Name),
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -526,7 +526,7 @@ func (ctx *Context) IsPodFitNode(name string, node string) error {
 // call volume binder to bind pod volumes if necessary,
 // internally, volume binder maintains a cache (podBindingCache) for pod volumes,
 // and before calling this, they should have been updated by FindPodVolumes and AssumePodVolumes.
-func (ctx *Context) BindPodVolumes(pod *v1.Pod) error {
+func (ctx *Context) bindPodVolumes(pod *v1.Pod) error {
 	podKey := string(pod.UID)
 	// the assumePodVolumes was done in scheduler-core, because these assumed pods are cached
 	// during scheduling process as they have directly impact to other scheduling processes.
@@ -534,11 +534,10 @@ func (ctx *Context) BindPodVolumes(pod *v1.Pod) error {
 	// then here we just need to retrieve that value from cache, to skip bindings if volumes are already bound.
 	if assumedPod, exist := ctx.schedulerCache.GetPod(podKey); exist {
 		if ctx.schedulerCache.ArePodVolumesAllBound(podKey) {
-			log.Logger.Info("BindPodVolumes Skipped",
-				zap.String("podName", pod.Name),
-				zap.String("reason", "Volumes are bound already"))
+			log.Logger.Info("Binding Pod Volumes skipped: all volumes already bound",
+				zap.String("podName", pod.Name))
 		} else {
-			log.Logger.Info("BindPodVolumes", zap.String("podName", pod.Name))
+			log.Logger.Info("Binding Pod Volumes", zap.String("podName", pod.Name))
 			return ctx.volumeBinder.Binder.BindPodVolumes(assumedPod)
 		}
 	}
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -524,6 +524,38 @@ func (ctx *Context) IsPodFitNode(name string, node string) error {
 	return fmt.Errorf("predicates were not running because pod or node was not found in cache")
 }
 
+// call volume binder to bind pod volumes if necessary,
+// internally, volume binder maintains a cache (podBindingCache) for pod volumes,
+// and before calling this, they should have been updated by FindPodVolumes and AssumePodVolumes.
+func (ctx *Context) BindPodVolumes(pod *v1.Pod) error {
+	if assumedPod, exist := ctx.schedulerCache.GetPod(string(pod.UID)); exist {
+		// assumePodVolumes is called when we do assumePod, why we call it again here?
+		// ideally this should not happen, we can simply do assumePodVolumes, based on the return value,
+		// see if pod needs the binding operation.
+		// However, the assumePodVolumes was done in scheduler-core, because these assumed pods are cached
+		// during scheduling process as they have directly impact to other scheduling processes.
+		// The actual binding happens in the shim right before binding to a node.
+		allBound, err := ctx.volumeBinder.Binder.AssumePodVolumes(assumedPod, pod.Spec.NodeName)
+		if err != nil {
+			log.Logger.Debug("AssumePodVolumes failed",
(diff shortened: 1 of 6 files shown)


pkg/cache/context.go
```diff
@@ -527,16 +527,23 @@ func (ctx *Context) IsPodFitNode(name string, node string) error {
...
-		if targetNode := ctx.schedulerCache.GetNode(node); targetNode != nil {
-			return ctx.schedulerCache.AssumePod(assumedPod)
+		if ctx.volumeBinder != nil {
+			// assume pod volumes before assuming the pod
+			// this will update scheduler cache with essential PV/PVC binding info
+			if _, err := ctx.volumeBinder.Binder.AssumePodVolumes(pod, node); err != nil {
+				return fmt.Errorf("failed to assume pod volumes")
+			}
+
+			// when add assumed pod, we make a copy of the pod to avoid
```
TaoYang526: When will ctx.volumeBinder be nil? Is it better to move the assumedPod logic out of this condition `if ctx.volumeBinder != nil {...}` ?

pkg/cache/external/scheduler_cache.go
```diff
@@ -293,6 +293,21 @@ func (cache *SchedulerCache) ForgetPod(pod *v1.Pod) error {
 	return nil
 }
 
+func (cache *SchedulerCache) BindPodVolumes(pod *v1.Pod) error {
+	if cache.volumeBinder != nil {
+		if err := cache.volumeBinder.Binder.BindPodVolumes(pod); err != nil {
```
TaoYang526: We should skip calling BindPodVolumes when all volumes are bound already, otherwise the process will fail inside it. Perhaps we need to remember pods whose volumes need to be bound in cache after calling AssumePodVolumes(returns whether all volumes of this pod are bound).

pkg/cache/task.go
```diff
@@ -42,14 +43,15 @@ type Task struct {
...
 	kubeClient     client.KubeClient
 	schedulerApi   api.SchedulerApi
+	schedulerCache *external.SchedulerCache
 	nodeName       string
 	sm             *fsm.FSM
 	lock           *sync.RWMutex
 }
 
-func newTask(tid string, app *Application, client client.KubeClient, schedulerApi api.SchedulerApi, pod *v1.Pod) Task {
+func newTask(tid string, app *Application, client client.KubeClient, schedulerApi api.SchedulerApi, cache *external.SchedulerCache, pod *v1.Pod) Task {
```
TaoYang526: cache is nil when calling this from task.go#createTaskFromPod.
```
task := newTask(string(pod.UID), app, client, scheduler, nil, pod)
```

pkg/plugin/predicates/predictor.go
```diff
@@ -201,6 +203,16 @@ func (p *Predictor) init() {
...
+
+	// GeneralPredicates are the predicates that are enforced by all Kubernetes components
+	// (e.g. kubelet and all schedulers)
+	p.RegisterFitPredicate(predicates.GeneralPred, predicates.GeneralPredicates)
+
+	// Fit is determined based on whether a pod can tolerate all of the node's taints
+	p.RegisterMandatoryFitPredicate(predicates.PodToleratesNodeTaintsPred, predicates.PodToleratesNodeTaints)
+
+	// Fit is determined based on whether a pod can tolerate unschedulable of node
+	p.RegisterMandatoryFitPredicate(predicates.CheckNodeUnschedulablePred, predicates.CheckNodeUnschedulablePredicate)
```
TaoYang526: Now Predictor#mandatoryFitPredicates is set but not used, should we support adding all mandatory-fit-predicates when initializing scheduler policy in predctor.go#NewPredictor ?

pkg/cache/context.go
yangwwei: volumeBinder might be nil when running some of UTs, I think it is safer to check nil first before calling it.

pkg/cache/external/scheduler_cache.go
yangwwei: > We should skip calling BindPodVolumes when all volumes are bound already

Ideally, we should do that, which means when all PVCs are bound, we skip this call ... but our logic is different from the default scheduler, it is not easy to achieve this. 

> otherwise the process will fail inside it
I don't think it will fail. Let's say a pod uses 3 PVCs, 2 of them are bound and 1 is not., then it still needs to call BindPodVolumes, I don't expect it will fail, correct? In such case, when 3 PVCs are all bound, it should not fail too, it just might cause a little bit overhead.

pkg/cache/task.go
yangwwei: this is fixed in latest patch

pkg/plugin/predicates/predictor.go
yangwwei: Yes, I think we should register all predicates in the init, in order to fully support all existing predicates. If we have anything missing, we need another ticket to add them back.

yangwwei:
@TaoYang526 please take a look if the updated patch looks good or not. thanks!

pkg/cache/task.go
```diff
@@ -42,14 +43,15 @@ type Task struct {
...
 	kubeClient     client.KubeClient
 	schedulerApi   api.SchedulerApi
+	volumeBinder   *volumebinder.VolumeBinder
 	nodeName       string
 	sm             *fsm.FSM
 	lock           *sync.RWMutex
 }
 
-func newTask(tid string, app *Application, client client.KubeClient, schedulerApi api.SchedulerApi, pod *v1.Pod) Task {
+func newTask(tid string, app *Application, client client.KubeClient, schedulerApi api.SchedulerApi, volumeBinder *volumebinder.VolumeBinder, pod *v1.Pod) Task {
```
wilfred-s: Instead of passing in three different references it would be much easier to pass in the `context` and do the calls using the `context` values. Cleans up the signatures of a number of calls.

pkg/cache/context.go
```diff
@@ -527,16 +529,23 @@ func (ctx *Context) IsPodFitNode(name string, node string) error {
...
-		assumedPod := pod.DeepCopy()
-		// assign the node name for pod
-		assumedPod.Spec.NodeName = node
-		if targetNode := ctx.schedulerCache.GetNode(node); targetNode != nil {
-			return ctx.schedulerCache.AssumePod(assumedPod)
+		if ctx.volumeBinder != nil {
+			// assume pod volumes before assuming the pod
+			// this will update scheduler cache with essential PV/PVC binding info
+			if _, err := ctx.volumeBinder.Binder.AssumePodVolumes(pod, node); err != nil {
+				return fmt.Errorf("failed to assume pod volumes")
```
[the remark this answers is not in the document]
yangwwei: Yeah, make sense. we can just return that err

pkg/cache/task.go
```diff
@@ -238,6 +242,21 @@ func (task *Task) postTaskAllocated(event *fsm.Event) {
...
+			if err := task.volumeBinder.Binder.BindPodVolumes(task.pod); err != nil {
+				errorMessage = fmt.Sprintf("bind pod volumes failed, name: %s, uid: %s, %#v",
+					task.pod.Name, task.pod.UID, err)
+				dispatcher.Dispatch(NewFailTaskEvent(task.applicationId, task.taskId, errorMessage))
+				events.GetRecorder().Eventf(task.pod,
+					v1.EventTypeWarning, "PodVolumesBindFailure", errorMessage)
+				return
+			}
+		}
+
```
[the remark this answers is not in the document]
yangwwei: Just looked at that PR, I think we don't need to worry about these changes as they are outside of the scheduler. Either we have this problem if the K8s version has this problem, or we don't have this problem if K8s is newer and get this fixed already, there is nothing we can do for it.

pkg/plugin/predicates/predictor.go
```diff
@@ -201,6 +203,16 @@ func (p *Predictor) init() {
 			return predicates.NewVolumeBindingPredicate(args.VolumeBinder)
 		},
 	)
+
+	// GeneralPredicates are the predicates that are enforced by all Kubernetes components
+	// (e.g. kubelet and all schedulers)
+	p.RegisterFitPredicate(predicates.GeneralPred, predicates.GeneralPredicates)
```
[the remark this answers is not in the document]
yangwwei: I did a cleanup on the predicates listing and kept it consistent with https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/algorithmprovider/defaults/register_predicates.go. The issue was we were using RegisterFit instead of RegisterMandatory earlier for `PodToleratesNodeTaintsPred` and `CheckNodeUnschedulablePred`, now I should get that fixed.

pkg/cache/task.go
```diff
@@ -238,6 +242,21 @@ func (task *Task) postTaskAllocated(event *fsm.Event) {
 		// task allocation UID is assigned once we get allocation decision from scheduler core
 		task.allocationUuid = allocUuid
 
+		// before binding pod to node, first bind volumes to pod
+		log.Logger.Debug("bind pod volumes",
+			zap.String("podName", task.pod.Name),
+			zap.String("podUID", string(task.pod.UID)))
+		if task.volumeBinder != nil {
+			if err := task.volumeBinder.Binder.BindPodVolumes(task.pod); err != nil {
```
TaoYang526: I think we should use cached pod instead of task pod here, since task pod didn't update its Spec.NodeName field and this call will fail when task.pod.Spec.NodeName is "".
I have tested successfully for pods with PVC but failed for pods without PVC, and I can't find where task.pod is updated for pods with PVC, can you pls figure out that?

pkg/cache/task.go
yangwwei: Fair point, changed to pass in context instead in latest commit.

pkg/cache/task.go
yangwwei: I am still looking at this. I think there might be some inconsistency in the cache somewhere, but could not figure out yet before my laptop battery dead... will update on this tomorrow.

yangwwei: I've updated a new commit. Hopefully, it can resolve this issue.
I used a workaround by calling AssumePodVolumes again right before BindPodVolumes, this is to get up-to-date volume bindings info in the volume-binder's cache, I am not sure if there is any other better approach to fix this.
@TaoYang526 could you please try this and let me know your thought? Thx

TaoYang526: Thanks @yangwwei for this update. I'm not sure if we need to call AssumePodVolumes again, I thought call BindPodVolumes with assumed pod should be fine. I'll try this again quickly then feedback a few minutes later.

TaoYang526: @yangwwei  I have tested this PR, it works fine after fixing a small issue as follows.
```
-               allBound, err := ctx.volumeBinder.Binder.AssumePodVolumes(assumedPod, pod.Spec.NodeName)
+               allBound, err := ctx.volumeBinder.Binder.AssumePodVolumes(assumedPod, assumedPod.Spec.NodeName)
```
I have also tested directly call BindPodVolumes without check if all volumes have been bound, it will fail for pods without unbound volume. So it's necessary to call AssumePodVolumes again and make sure not all volumes are bound before calling BindPodVolumes.

yangwwei: Thanks for the quick update! Just updated the code to fix this.

pkg/cache/context.go
```diff
@@ -527,12 +564,20 @@ func (ctx *Context) IsPodFitNode(name string, node string) error {
...
 		assumedPod := pod.DeepCopy()
+		// assume pod volumes, this will update bindings info in cache
+		if ctx.volumeBinder != nil {
+			// assume pod volumes before assuming the pod
+			// this will update scheduler cache with essential PV/PVC binding info
+			if _, err := ctx.volumeBinder.Binder.AssumePodVolumes(assumedPod, node); err != nil {
+				return err
+			}
+		}
+
```
wilfred-s: If the node does not exist we will have assumed the volumes on a node that does not exists. This does not make sense. It will also not keep the copied pod object around as it should be garbage collected when it is not added to the cache.
I think this all needs to be inside the check for the `targetNode`

pkg/cache/context.go
```diff
@@ -519,6 +524,38 @@ func (ctx *Context) IsPodFitNode(name string, node string) error {
...
+		// The actual binding happens in the shim right before binding to a node.
+		allBound, err := ctx.volumeBinder.Binder.AssumePodVolumes(assumedPod, assumedPod.Spec.NodeName)
+		if err != nil {
+			log.Logger.Debug("AssumePodVolumes failed",
+				zap.String("podName", pod.Name),
+				zap.Error(err))
+			return err
+		}
+
+		if allBound {
```
wilfred-s: I don't think this is correct, we should only call this once in the `AssumePod` method.
The result  should be stored in the `assumedPods` mapping and leveraged here. 
```
if allBound, ok := assumedPods[pod id]; ok {
    if !allBound {
        return bind result volumes
    }
    return nil
}
return an error pod not assumed
```

pkg/cache/context.go
yangwwei: If the node doesn't exist, then bindPodVolumes will fail. If that fails, this will trigger the release operation and by then forgetPod will be called. This should be consistent.
And for the copied pod, that is stored in `ctx.schedulerCache`, this is to avoid directly modifying the original reference of the pod.

pkg/cache/context.go
yangwwei: Unfortunately, this is the only option I come up with so far.
The reason I have commented in the code, please let me know if it makes sense

```
assumePodVolumes is called when we do assumePod, why we call it again here?
ideally this should not happen, we can simply do assumePodVolumes, based on the return value, see if pod needs the binding operation.
However, the assumePodVolumes was done in scheduler-core, because these assumed pods are cached during scheduling process as they have directly impact to other scheduling processes. The actual binding happens in the shim right before binding to a node.
```

wilfred-s: we're calling `AssumePodVolumes` twice on the same object:
1. first call in `AssumePod` in the context we deep copy the pod and then call `AssumePodVolumes`. After this we store the pod in the cache by calling `cache.AssumePod`. The cached version is thus the modified one.
1. the second call is in `BindPodVolumes` in the context, we pull the pod from the cache and call `AssumePodVolumes` again on the same object.

The two pods should thus be the same objects unless we keep overwriting the cached version which I would find strange and consider a bug. That is why I think we need to only call it once. The K8 scheduler also only calls it once.
It does require us to not ignore the result of the first `AssumePodVolumes` and pass it in to the `context.AssumePod` to set the boolean in cache assumedPods map.

pkg/cache/context.go
wilfred-s: `AssumePod` in the context is called on reconcile from the core, the node comes from core. Scheduling should fail at that point as the core is using a node that is not in the shim. We should retry scheduling or fail, not proceed as if nothing happened and then hope the volume bind fails later. What happens if we do not have volumes? 

I really don't think we should even call `AssumePodVolumes` if we know the node does not exist.

We also call `AssumePodVolumes` on the deep copy, when the node is not found that deep copy of the object is **not** added to the cache. We assume the volumes on a pod that we immediately drop. Bind on the original object will fail as the volumes have never been assumed.

Relying on a failure of `BindPodVolumes` to forget the pod is strange as we have not assumed the pod and never will.

yangwwei:
Had a discussion with @wilfred-s , we agree that calling AssumePodVolumes twice might be a bad idea. Instead, we decided to store the allBound info in schedulerCache. Please see my updates in the PR. Thanks!

pkg/cache/context.go
yangwwei: Make sense, fixed in latest PR.

pkg/cache/context.go
yangwwei: Got the point now, yeah, we should check the node existence for assumePodVolumes too, to keep them in sync.

TaoYang526:
Thanks @yangwwei for the update and @wilfred-s for the review. 
This PR looks better now.

wilfred-s:
this resolves #66

PR #73, opened 2020-01-23
Misc fixes and improvements

yangwwei:
Fixes include:
- Set pod condition to unscheduable in a delay-manner, in order to trigger auto-scaling.
- Add sanity check (about volumes) for each task before sending it for scheduling
- Improve the auto-app generation code to use [PodNamespace]_[PodName]_[Timestamp], this is to avoid potentially non-readable, too-long ID which could be rejected by admission controllers.
- Changing the image pull policy to be "Always", easier for rolling update


pkg/cache/task.go
```diff
@@ -196,6 +197,37 @@ func (task *Task) handleSubmitTaskEvent(event *fsm.Event) {
...
+	// the auto-scaling process.
+	// TODO improve this
+	// Note, this approach is suboptimal, when a task cannot be allocated, it doesn't always
+	// mean it needs the nodes to scale up. E.g task runs out of max capacity of the queue,
+	// user/app runs out of the limit etc. Ideally, we should add more interactions between
+	// core and shim to negotiate on when to set the state to unscheduable and trigger the
+	// auto-scaling appropriately.
+	go func(self *Task) {
+		time.Sleep(5*time.Second)
+		log.Logger.Info("updating pod state",
```
wilfred-s: The log comment should be checking pod state, we might not update at all.

pkg/cache/task.go
```diff
@@ -196,6 +197,37 @@ func (task *Task) handleSubmitTaskEvent(event *fsm.Event) {
...
+				&v1.PodCondition{
+					Type:    v1.PodScheduled,
+					Status:  v1.ConditionFalse,
+					Reason:  v1.PodReasonUnschedulable,
+					Message: "retrying scheduling",
+				}); err != nil {
+				log.Logger.Error("update pod condition failed",
+					zap.Error(err))
+			}
+			self.lock.Unlock()
```
wilfred-s: Can we change this to a `defer`

pkg/cache/task.go
```diff
@@ -196,6 +197,37 @@ func (task *Task) handleSubmitTaskEvent(event *fsm.Event) {
...
+	// after a small amount of time, if the task is still not able to get allocation,
+	// put the pod to unscheduable state. This will trigger the cluster-auto-scaler to launch
+	// the auto-scaling process.
+	// TODO improve this
+	// Note, this approach is suboptimal, when a task cannot be allocated, it doesn't always
+	// mean it needs the nodes to scale up. E.g task runs out of max capacity of the queue,
+	// user/app runs out of the limit etc. Ideally, we should add more interactions between
+	// core and shim to negotiate on when to set the state to unscheduable and trigger the
+	// auto-scaling appropriately.
+	go func(self *Task) {
```
wilfred-s: Using `self` is not really clear as the reference. Can we not just keep it at `task` ?

pkg/cache/context.go
```diff
@@ -503,6 +504,20 @@ func (ctx *Context) triggerReloadConfig() {
 	}
 }
 
+func (ctx *Context) UpdatePodCondition(pod *v1.Pod, condition *v1.PodCondition) error {
+	log.Logger.Info("Updating pod condition",
+		zap.String("namespace", pod.Namespace),
+		zap.String("name", pod.Name),
+		zap.String("", string(condition.Type)),
+		zap.String("status", string(condition.Status)),
+		zap.String("reason", condition.Reason))
```
wilfred-s: Should we not just log the whole condition here?
If not we need a description for the `Type` log

pkg/cache/application.go
```diff
@@ -235,14 +234,22 @@ func (app *Application) Schedule() {
...
-					NewSimpleTaskEvent(task.applicationID, task.taskID, events.InitTask)); err != nil {
-					// something goes wrong when transit task to PENDING state,
-					// this should not happen because we already checked the state
-					// before calling the transition. Nowhere to go, just log the error.
-					log.Logger.Warn("init task failed", zap.Error(err))
+				// for each new task, we do a sanity check before moving the state to Pending_Schedule
+				if err := task.sanityCheckBeforeScheduling(); err == nil {
+					// note, if we directly trigger submit task event, it may spawn too many duplicate
+					// events, because a task might be submitted multiple times before its state transits to PENDING.
+					if err := task.handle(
```
wilfred-s: don't shadow here

pkg/cache/context.go
```diff
@@ -503,6 +504,20 @@ func (ctx *Context) triggerReloadConfig() {
 	}
 }
 
+func (ctx *Context) UpdatePodCondition(pod *v1.Pod, condition *v1.PodCondition) error {
```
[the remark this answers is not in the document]
yangwwei: Done

pkg/cache/task.go
yangwwei: Moved the log to the inside of "if". And added more info.

pkg/cache/task.go
yangwwei: done

pkg/cache/task.go
yangwwei: Renamed to t,  because this is a go runtine func inside of task, so I avoid to naming it same as the parent refer.

pkg/cache/context.go
yangwwei: Changed it to zap.Any("podCondition", condition)

pkg/cache/application.go
yangwwei: Done

wilfred-s:
Thank you for the update, I'll merge this

PR #74, opened 2020-02-05
Identify spark pod with label spark-app-selector

yangwwei:
Native spark on K8s adds a label "spark-app-selector" to track spark app ID. We should directly read from this label.

### DIFF
--- a/pkg/dispatcher/dispatcher_test.go
+++ b/pkg/dispatcher/dispatcher_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package dispatcher
 
 import (
+	"fmt"
 	"runtime"
 	"strings"
 	"sync"
@@ -24,6 +25,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/utils"
 	"gotest.tools/assert"
 
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/events"
@@ -204,7 +206,7 @@ func TestDispatchTimeout(t *testing.T) {
 	handledChan := make(chan bool)
 	RegisterEventHandler(EventTypeApp, func(obj interface{}) {
 		if _, ok := obj.(events.ApplicationEvent); ok {
-			time.Sleep(2 * time.Second)
+			time.Sleep(30 * time.Second)
 			handledChan <- true
 		}
 	})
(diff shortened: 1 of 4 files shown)


yangwwei:
recheck

pkg/dispatcher/dispatcher_test.go
```diff
@@ -17,13 +17,15 @@ limitations under the License.
...
 import (
+	"fmt"
 	"runtime"
 	"strings"
 	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
 
+	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/utils"
```
wilfred-s: all yunikorn imports should be grouped together. I filed new issues to fix the linter setting but this should be with the events import (k8shim  #75)

pkg/dispatcher/dispatcher_test.go
```diff
@@ -201,31 +204,57 @@ func TestDispatchTimeout(t *testing.T) {
 	}()
 
 	// pretend to be an time-consuming event-handler
```
wilfred-s: This comment does not make sense any more as the sleep has been removed

pkg/dispatcher/dispatcher_test.go
```diff
@@ -201,31 +204,57 @@ func TestDispatchTimeout(t *testing.T) {
...
+	event2 := TestAppEvent{
+		appID:     "test-02",
+		eventType: events.RunApplication,
+		flag:      make(chan bool),
+	}
+
+	// dispatch all events
+	Dispatch(event0)
+	Dispatch(event1)
+	Dispatch(event2)
```
wilfred-s: Why unwrap the for loop? A loop makes the code much more readable and easier to maintain.

pkg/dispatcher/dispatcher_test.go
yangwwei: Done.

pkg/dispatcher/dispatcher_test.go
yangwwei: Fixed.

pkg/dispatcher/dispatcher_test.go
yangwwei: This is no longer needed. Earlier I need to access each of the events to verify something. But that is now removed. I just changed to use a simple loop. Thanks.

yangwwei:
Thanks @TaoYang526 for the review.
Since the change is trivial, I have merged the PR after addressing @wilfred-s' comment, because right now the UT is broken without this fix. @wilfred-s , if there is any more comment not been addressed, we can use a separate issue to track. Hope this makes sense.

PR #77, opened 2020-02-15
Recovery times out if the cluster initially has no nodes

yangwwei:
1. when there is no nodes in the cluster, recovery timesout
 2. support KUBECONFIG environment variable for local mode

### DIFF
--- a/pkg/common/resource.go
+++ b/pkg/common/resource.go
@@ -159,8 +159,9 @@ func CreateUpdateRequestForNewNode(node Node) si.UpdateRequest {
 func CreateUpdateRequestForUpdatedNode(node Node) si.UpdateRequest {
 	// Currently only includes resource in the update request
 	nodeInfo := &si.UpdateNodeInfo{
-		NodeID:              node.name,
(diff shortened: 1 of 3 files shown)


pkg/common/resource.go
```diff
@@ -159,8 +159,9 @@ func CreateUpdateRequestForNewNode(node Node) si.UpdateRequest {
 func CreateUpdateRequestForUpdatedNode(node Node) si.UpdateRequest {
 	// Currently only includes resource in the update request
 	nodeInfo := &si.UpdateNodeInfo{
-		NodeID:              node.name,
-		SchedulableResource: node.resource,
+		NodeID:               node.name,
+		Attributes:           make(map[string]string),
```
wilfred-s: This code has been the same for a long time and is only called within the mock scheduler (test only)
Currently without this change I do not see a nil pointer in the tests.

Please explain how this caused a nil pointer.

wilfred-s:
With the changes to the code the fix looks good. I'll commit soon.
Just for future reference it would be good to explain where the nil pointer came from.

PR #80, opened 2020-03-02
[YUNIKORN-11] release pod resources from the scheduler when its status is completed

yangwwei:
this fixes https://github.com/apache/incubator-yunikorn-core/issues/90


wilfred-s:
Linked to jira YUNIKORN-11
Left a comment in the jira about what we are trying to fix. The code looks good but before I approve and or commit anything we need to finish the discussion in the jira.

PR #82, opened 2020-03-10
[YUNIKORN-5]Change the contribution guide and add .asf.yaml

wilfred-s:
Updating the contribution guide to remove the last references to Cloudera.
Adding a new .asf.yaml file to turn off github issues.


wilfred-s:
NOTE: this should be committed after we have pushed the changes into the core so we can update the dependencies accordingly.

PR #84, opened 2020-03-18
[YUNIKORN-36] Cleanup leaked resources when deleting the helm chart

TaoYang526:
Please refer to issue: https://issues.apache.org/jira/browse/YUNIKORN-36

### DIFF
--- a/helm-charts/yunikorn/templates/cleanup.yaml
+++ b/helm-charts/yunikorn/templates/cleanup.yaml
@@ -20,14 +20,19 @@ spec:
     spec:
       serviceAccountName: {{ .Values.serviceAccount }}
       containers:
-        - name: kubectl
-          image: "{{ .Values.hyperkube_image.repository }}:{{ .Values.hyperkube_image.tag }}"
-          imagePullPolicy: {{ .Values.hyperkube_image.pullPolicy }}
+        - name: curl
+          image: "{{ .Values.curl_image.repository }}:{{ .Values.curl_image.tag }}"
+          imagePullPolicy: {{ .Values.curl_image.pullPolicy }}
           command:
             - /bin/sh
             - -c
             - >
-              kubectl delete configmap yunikorn-configs;
-              kubectl delete serviceaccount yunikorn-admin;
-              kubectl delete clusterrolebinding yunikorn-rbac;
+              APISERVER=https://kubernetes.default.svc.cluster.local;
+              SERVICEACCOUNT=/var/run/secrets/kubernetes.io/serviceaccount;
+              NAMESPACE=$(cat ${SERVICEACCOUNT}/namespace);
+              TOKEN=$(cat ${SERVICEACCOUNT}/token);
+              CACERT=${SERVICEACCOUNT}/ca.crt;
+              curl --cacert ${CACERT} --header "Authorization: Bearer ${TOKEN}" -X DELETE ${APISERVER}/api/v1/namespaces/${NAMESPACE}/configmaps/yunikorn-configs;
+              curl --cacert ${CACERT} --header "Authorization: Bearer ${TOKEN}" -X DELETE ${APISERVER}/api/v1/namespaces/${NAMESPACE}/serviceaccounts/yunikorn-admin;
+              curl --cacert ${CACERT} --header "Authorization: Bearer ${TOKEN}" -X DELETE ${APISERVER}/apis/rbac.authorization.k8s.io/v1beta1/clusterrolebindings/yunikorn-rbac;
       restartPolicy: OnFailure
--- a/helm-charts/yunikorn/templates/cleanup.yaml
+++ b/helm-charts/yunikorn/templates/cleanup.yaml
@@ -27,7 +27,7 @@ spec:
             - /bin/sh
             - -c
             - >
-              APISERVER=https://kubernetes.default.svc.cluster.local;
+              APISERVER=https://kubernetes.default.svc;
               SERVICEACCOUNT=/var/run/secrets/kubernetes.io/serviceaccount;
               NAMESPACE=$(cat ${SERVICEACCOUNT}/namespace);
               TOKEN=$(cat ${SERVICEACCOUNT}/token);
--- a/helm-charts/yunikorn/templates/cleanup.yaml
+++ b/helm-charts/yunikorn/templates/cleanup.yaml
@@ -1,0 +1,33 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: yunikorn-cleanup
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "3"
+    "helm.sh/hook-delete-policy": hook-succeeded
+  labels:
+    app: yunikorn
+    release: {{ .Release.Name }}
+spec:
+  template:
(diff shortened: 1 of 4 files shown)


wilfred-s:
I don't think that relying on hyperkube is the right thing as kubernetes has just merged the changes to stop releasing the hyperkube container image [kubernetes/kubernetes#88676](https://github.com/kubernetes/kubernetes/pull/88676)

yangwwei:
Hi @TaoYang526 

Can we leverage the yunikorn scheduler docker image which already has the script installed, we can do 
```
admission_util.sh delete
```
to clean up everything

TaoYang526:
Thanks @wilfred-s , @yangwwei  for the review.
> Can we leverage the yunikorn scheduler docker image which already has the script installed, we can do `admission_util.sh delete ` to clean up everything

I think we can't do that since the command will be called in PreStop hook which is called immediately before a container is terminated due to an API request or management event such as liveness probe failure, preemption, resource contention and others, not just before the final termination. (Please refer to https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks)

I haven't find a proper image to replace deprecated hyperkube image after a quick search, I'll find another image or another way to solve this problem. Please share your thoughts if you have any idea about this. Thanks.

TaoYang526:
Hi, @yangwwei @wilfred-s 
This PR has been updated to clean up things via curling to APIServer instead of relying on hyperkube which is removed from k8s base images, please help to review it again. Thanks.

TaoYang526:
Thanks @yangwwei for the feedback about the URL of APIServer.
I discovered that there's something wrong in the former image which may cause DNS can't resolve the service address correctly in search domains defined in /etc/resolv.conf, that's why it had to use the full service name with suffix '.cluster.local' to connect with API server.
Latest commit replaces the incorrect curl image with official curl image, and update the URL of APIServer with non-suffix service name: 'kubernetes.default.svc' just as described in the kubernetes official doc: https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/#accessing-the-kubernetes-api.

helm-charts/yunikorn/templates/cleanup.yaml
```diff
@@ -0,0 +1,38 @@
...
+            - >
+              APISERVER=https://kubernetes.default.svc;
+              SERVICEACCOUNT=/var/run/secrets/kubernetes.io/serviceaccount;
+              NAMESPACE=$(cat ${SERVICEACCOUNT}/namespace);
+              TOKEN=$(cat ${SERVICEACCOUNT}/token);
+              CACERT=${SERVICEACCOUNT}/ca.crt;
+              curl --cacert ${CACERT} --header "Authorization: Bearer ${TOKEN}" -X DELETE ${APISERVER}/api/v1/namespaces/${NAMESPACE}/configmaps/yunikorn-configs;
+              curl --cacert ${CACERT} --header "Authorization: Bearer ${TOKEN}" -X DELETE ${APISERVER}/api/v1/namespaces/${NAMESPACE}/serviceaccounts/yunikorn-admin;
+              curl --cacert ${CACERT} --header "Authorization: Bearer ${TOKEN}" -X DELETE ${APISERVER}/apis/rbac.authorization.k8s.io/v1beta1/clusterrolebindings/yunikorn-rbac;
+      restartPolicy: OnFailure
```
yangwwei, line 38: I feel we don't need to restart this on failure, if the execution failed, probably the second attempt will fail too.
Can we change this to Never?

helm-charts/yunikorn/templates/cleanup.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: yunikorn-cleanup
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-weight": "3"
```
yangwwei, line 8: My understanding the weight is giving us a determinable order when installing resources.
But why we have the weight=3 for the cleanup?
This is a post-delete hook, which only triggers when we delete the helm chart, correct?

TaoYang526: Yes, absolutely right. I will remove this annotation. Thanks for this reminder.

helm-charts/yunikorn/templates/cleanup.yaml
TaoYang526: Make sense.

TaoYang526:

> With this patch, does it mean we can delete all yunikorn created resources by a simple "helm delete" command?  do we still need the pre-stop hook?

We can clean up all resources that yunikorn created by "helm uninstall" command. However, I think the pre-stop hook is more suitable to do that for admission-controller, since the resources created for admission-controller should better be managed by admission-controller itself, otherwise if these resources changed we may need to update cleanup.yaml for helm-charts as well as admission_util.sh script for admission-controller. Thoughts?

yangwwei:
I did some tests, install/delete helm chart again and again.

I found an issue we need to fix. When I delete the helm-chart second time, the job fails because the job (with that name) already exists. I think we can fix this by

```
-  name: yunikorn-cleanup
+  generateName: yunikorn-cleanup
```

so every time the job will have a different name.

Another thing is when the job finished, the job is not deleted and the cleanup pod is left-over. Can we take a look at if https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#ttl-mechanism-for-finished-jobs can help the cleanup?

Other than these issues, it looks good.

TaoYang526:
Thanks @yangwwei for the feedback.

> When I delete the helm-chart second time, the job fails because the job (with that name) already exists.

Do you mean that "helm uninstall" command was executed again before the uninstall process was done? I think it's ok since the first execution couldn't be affected by the second one, right?

> when the job finished, the job is not deleted and the cleanup pod is left-over.

I haven't seen this happen, it should be cleaned up after the job is done since we have configured `"helm.sh/hook-delete-policy": hook-succeeded` in cleanup.yaml. Could you please elaborate how this may happen?

TaoYang526:
After discussed with @yangwwei offline, it seems that `"helm.sh/hook-delete-policy": hook-succeeded` config may not works well in helm v2 (at least for several specific versions: https://github.com/helm/helm/issues/1769), we can assume that it can work well in helm v3 and I have already tested it for many times in my local environment. 

Also, I have tested to add generateName but fails to uninstall since it's not available in helm(https://github.com/helm/helm/issues/3348), and TTL may not work well unless that both kube-apiserver and kube-controller-manager enable feature gate TTLAfterFinished according to https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/.

Now helm v3 is the stable and default version according to helm official website(https://helm.sh/docs/), and there are lots of changes between v2 and v3, we decided to do nothing more for this PR and update helm commands based on helm v3 in the User Guide document via another PR in core project: https://github.com/apache/incubator-yunikorn-core/pull/129

yangwwei:
I agree this sounds good to me. 
We can suggest people use helm3, for which I think most people are already doing that. Let's track the doc changes in another ticket. This one looks good to go. Thanks!

yangwwei:
BTW, this approach is following the doc: https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/#without-using-a-proxy

PR #88, opened 2020-03-26
[YUNIKORN-54] admission-controller should keep generated ID tight

yangwwei:
Simple change. We need a better format of generated appIDs. 
And I have filed https://issues.apache.org/jira/browse/YUNIKORN-55 to re-add that info back to allocation asks (instead of having them in the app ID).

### DIFF
--- a/pkg/plugin/admissioncontrollers/webhook/admission_controller.go
+++ b/pkg/plugin/admissioncontrollers/webhook/admission_controller.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strconv"
 	"time"
 
 	"go.uber.org/zap"
@@ -124,6 +125,15 @@ func updateSchedulerName(patch []patchOperation) []patchOperation {
 	})
 }
 
+// the generated ID is using [PodName]_[Timestamp] naming convention.
+// the generated ID should be unique even the pod name is same (differ on nano time),
+// and the max length of the ID is 63 chars.
+func generateAppID(podName string) string {
+	timestamp := strconv.FormatInt(time.Now().UnixNano(), 10)
+	generatedID := fmt.Sprintf("%.43s_%.19s", podName, timestamp)
+	return generatedID
+}
+
 func updateLabels(pod *v1.Pod, patch []patchOperation) []patchOperation {
 	log.Logger.Info("updating pod labels",
 		zap.String("podName", pod.Name),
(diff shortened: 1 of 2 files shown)


pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -138,25 +138,7 @@ func updateLabels(pod *v1.Pod, patch []patchOperation) []patchOperation {
...
-			// look for generateName instead
-			podName := "unknown"
-			if pod.Name != "" {
-				podName = pod.Name
-			} else if pod.GenerateName != "" {
-				podName = pod.GenerateName
-			}
-
-			generatedID := fmt.Sprintf("%s_%s_%d", podNamespace, podName, time.Now().Unix())
+			generatedID := fmt.Sprintf("__app_%d", time.Now().UnixNano())
```
wilfred-s: I do not understand the change. We are going from using the pod name or generate name combined with a timestamp to just `app` and a timestamp. It would make tracking pods on the YuniKorn side really difficult.

Can we not use a length limited label setup still leveraging the name or generate name?
* remove the podNamespace (not making it more unique)
* limit the size of podName to 50 char in the printed text
* use a `_` as separator
* use the millis since 1970 (13 characters) if seconds is not good enough
```
	timestamp := strconv.FormatInt(time.Now().UnixNano(), 10)
	generatedID := fmt.Sprintf("%.50s_%.13s", podName, timestamp)
```
This will give a guaranteed maximum of 64 chars. We can even tweak that down if you think that is needed.

yangwwei: One of the main reasons to have this change, not just to keep it < 64 char. Is the fact during the webhook phase, pod namespace, pod name could both be empty (in many cases). My suggestion is to use a general generated ID just with a timestamp, but get more info from attributes https://issues.apache.org/jira/browse/YUNIKORN-55.

wilfred-s: Why throw out all the info we have and then try to substitute it with some other info?
If we have the detail we should use it. The change will limit the label to a maximum of 64 char.

If the build up `podName` turns out to be empty we should substitute at that point instead of the  generic rule.

adamantal: To me @wilfred-s suggestion make sense, assuming podName is not going to change. 
It is much more informative from debugging perspective for instance to have IDs with more details in it.

yangwwei: Thanks for the review. I think that makes sense, we can do that.
In my latest commit, I use a very similar approach like @wilfred-s suggested, the only difference is to maintain a full timestamp to ensure the ID would be unique (pod name might be the same). UT added.  Please help to review, thanks!

pkg/plugin/admissioncontrollers/webhook/admission_controller_test.go
```diff
@@ -232,3 +258,26 @@ func TestValidateConfigMap(t *testing.T) {
 	assert.Equal(t, err.Error(), "required config 'queues.yaml' not found in this configmap")
 	// skip further validations which depends on the webservice of yunikorn-core
 }
+
+func TestGenerateAppID(t *testing.T) {
+	appID := generateAppID("this-is-a-pod-name")
+	t.Logf("generated appID %s", appID)
```
wilfred-s, line 264: These log messages show up as failures in the github runs. We don't need them to be logged for every test run, i'll remove on commit.

yangwwei:
Thanks @wilfred-s !

PR #91, opened 2020-03-28
[YUNIKORN-63] Handle occupied resources while registering a node.

yangwwei:


wilfred-s:
See my comment in the jira: why this change from not needing it in apache/incubator-yunikorn-core#108 to needing it now?

yangwwei:
> See my comment in the jira: why this change from not needing it in apache/incubator-yunikorn-core#108 to needing it now?

This is found while testing the recovery on a real cluster. When it tries to report the occupied resources to the scheduler-core, **the node is not yet registered**. Therefore, the latest patch ensures that when we register the node, we need to have the initial occupied resources counted correctly.

My previous fix (only updating the node capacity) might have the same issue, but I guess I missed that during the local test (not covering the recovery test).

PR #93, opened 2020-04-01
[YUNIKORN-64] fix zap encoder config

wilfred-s:
Zap logging requires an encoder to be set Duration and Time.
Duration encoder was not set causing a panic.
Cleanup of the init code of the logger.


wilfred-s:
This change does **not** depend on the core side changes for the same jira.

PR #94, opened 2020-04-03
[YUNIKORN-76] YuniKorn admission-controller should be installed on the same host as Yunikorn-scheduler

kingamarton:
Tested with the following steps:
- created a multi node cluster locally with Vagrant
- changed the image pull policy to Never in Helm chart
- built the image locally and made a tar.gz
- uploaded the tar.gz with the images to the hosts and load the images
- checked if the two pod were scheduled on the same node each time.

### DIFF
--- a/deployments/admission-controllers/scheduler/templates/server.yaml.template
+++ b/deployments/admission-controllers/scheduler/templates/server.yaml.template
@@ -15,6 +15,16 @@ spec:
       labels:
         app: yunikorn
     spec:
+      affinity:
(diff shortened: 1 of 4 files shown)


sunilgovind:
@yangwwei @kingamarton cud we confirm this change covers our helm install files as well'?

yangwwei:
Hi @sunilgovind 

It does for upstream. But thanks for the reminder, we also need a patch internally to get this work. Let's track that separately.

PR #96, opened 2020-04-05
[YUNIKORN-72] data race in unit test: TestTaskStateTransitions

yangwwei:
Fix the data-race in UT

### DIFF
--- a/pkg/cache/context_test.go
+++ b/pkg/cache/context_test.go
@@ -21,13 +21,13 @@ package cache
 import (
 	"testing"
 
+	"github.com/apache/incubator-yunikorn-k8shim/pkg/conf"
 	"gotest.tools/assert"
 	"k8s.io/api/core/v1"
 
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/appmgmt/interfaces"
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/client"
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/events"
-	"github.com/apache/incubator-yunikorn-k8shim/pkg/conf"
 )
 
 const fakeClusterID = "test-cluster"
@@ -36,17 +36,7 @@ const fakeClusterSchedulerName = "yunikorn-test"
 const fakeClusterSchedulingInterval = 1
 
 func initContextForTest() *Context {
-	configs := conf.SchedulerConf{
-		ClusterID:      fakeClusterID,
-		ClusterVersion: fakeClusterVersion,
-		SchedulerName:  fakeClusterSchedulerName,
-		Interval:       fakeClusterSchedulingInterval,
-		KubeConfig:     "",
-		TestMode:       true,
-	}
-
-	conf.Set(&configs)
-
+	conf.GetSchedulerConf().SetTestMode(true)
 	context := NewContext(client.NewMockedAPIProvider())
 	return context
 }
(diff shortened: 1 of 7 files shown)


yangwwei:
hi @wilfred-s 
please help to review this PR, thanks

pkg/conf/schedulerconf.go
```diff
@@ -45,7 +47,7 @@ const (
 	DefaultKubeBurst            = 1000
 )
 
-var configuration *SchedulerConf
+var pointer unsafe.Pointer
```
adamantal, line 50: Could you give rationale why the `unsafe` should be used here?

pkg/conf/schedulerconf.go
```diff
@@ -68,12 +70,12 @@ type SchedulerConf struct {
 }
 
 func GetSchedulerConf() *SchedulerConf {
-	return configuration
+	return (*SchedulerConf)(atomic.LoadPointer(&pointer))
 }
 
 // unit tests may need to override configuration
```
adamantal, line 76: The comment should be modified, because not only the tests use this function anymore.

pkg/conf/schedulerconf.go
```diff
@@ -68,12 +70,12 @@ type SchedulerConf struct {
...
 func Set(conf *SchedulerConf) {
```
adamantal, line 77: As far as I can see, we do not import this function - so we can use `set`.

pkg/conf/schedulerconf_test.go
```diff
@@ -37,3 +39,34 @@ func TestDefaultValues(t *testing.T) {
...
+
+// run some concurrent write/read
+// if the Get and Set function is not thread safe,
+// this will give data race warnings when running with -race flag
+func TestConcurrency(t *testing.T) {
+	wg := sync.WaitGroup{}
+
+	wg.Add(1)
+	go func() {
+		for i := 0; i < 10; i++ {
```
adamantal, line 51: Can you move the `10` to a constant and use it here and above in the other goroutine?

pkg/conf/schedulerconf.go
wilfred-s: I do not see any reason why we should use unsafe. Unsafe has its uses but I think  this is not one of them.
Unsafe is not stable and can change in any go release.

The other problem is with the `Set`.  `Init` will always be called before we set a new one. If we hold on to the previous pointer that we got via a `GetSchedulerConf()` then that could cause issues.
Atomic also does not provide any ordering guarantee, it just provides the correct value at that point in time. 

`Set` is only called in tests. Instead of replacing pointer we should update the content of the structure created in `init`. That means the pointer will never change. I would rather go to a full get/set structure for the config.

yangwwei: Thanks, @adamantal @wilfred-s 
this a fair point, I removed unsafe pkg used in this patch. Instead, using a lock and get/set approach.
Since the conf object is global, I think we should not replace the object anytime, just stick to update or get values.
Please see my latest commit. Thanks!

pkg/cache/context_test.go
```diff
@@ -21,13 +21,13 @@ package cache
 import (
 	"testing"
 
+	"github.com/apache/incubator-yunikorn-k8shim/pkg/conf"
```
wilfred-s, line 24: should be with the rest of the shim import

yangwwei:
Thanks @wilfred-s !

PR #99, opened 2020-04-08
[YUNIKORN-82] Fix wrong revision in docker image.

TaoYang526:
The target split in get-revision commands for yunikorn-core-revision and yunikorn-scheduler-interface-revision should be updated when the project names have changed from "yunikorn-xxx" to "incubator-yunikorn-xxx".


yangwwei:
Thanks @TaoYang526 !

PR #100, opened 2020-04-09
[YUNIKORN-84] Always waiting in recovery phase as long as rejected node exists

TaoYang526:
In recovery phase, scheduler keeps waiting until all nodes are recovered, but not all nodes are considered according to Context#recover in context_recovery.go, only nodes with Healthy or Draining state will be counted, nodes with Rejected state should be counted as well.

Updates:
- nodes with Rejected state are counted
- Improve UT TestNodesRecovery
- Remove unused node state: Unhealthy

@yangwwei @wilfred-s  please help to review this PR. Thanks.

### DIFF
--- a/pkg/cache/context_recovery_test.go
+++ b/pkg/cache/context_recovery_test.go
@@ -19,12 +19,15 @@
 package cache
 
 import (
+	"reflect"
+	"strconv"
 	"testing"
 	"time"
 
 	"gotest.tools/assert"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	apis "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/appmgmt/interfaces"
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/client"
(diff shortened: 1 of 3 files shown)


pkg/cache/context_recovery_test.go
```diff
@@ -106,83 +109,91 @@ func TestNodesRecovery(t *testing.T) {
...
+	}
 	apiProvide4test.SetNodeLister(nodeLister)
 
 	mockedAppRecover := test.NewMockedRecoverableAppManager()
-	if err := context.recover([]interfaces.Recoverable{mockedAppRecover}, 3*time.Second); err == nil {
+	if err := context.recover([]interfaces.Recoverable{mockedAppRecover}, 1*time.Second); err == nil {
 		t.Fatalf("expecting timeout here!")
 	} else {
 		t.Logf("context stays waiting for recovery, error: %v", err)
 	}
```
wilfred-s, line 147: do we really need to know the content of the error if we are expecting it?
I know it was not introduced in this change but lets remove it

pkg/cache/context_recovery_test.go
```diff
@@ -106,83 +109,91 @@ func TestNodesRecovery(t *testing.T) {
...
 	dispatcher.Dispatch(CachedSchedulerNodeEvent{
-		NodeID: "host0001",
+		NodeID: schedulerNodes[0].name,
 		Event:  events.NodeAccepted,
 	})
+	expectedStates[0] = events.States().Node.Healthy
+	if err := utils.WaitForCondition(func() bool {
+		return reflect.DeepEqual(getNodeStates(schedulerNodes), expectedStates)
+	}, 100*time.Millisecond, 3*time.Second); err != nil {
+		t.Fatal("unexpected node states")
```
wilfred-s, line 165: log the error as it could help find test failure

pkg/cache/context_recovery_test.go
```diff
@@ -106,83 +109,91 @@ func TestNodesRecovery(t *testing.T) {
...
+		Event:  events.NodeRejected,
+	})
+	expectedStates[1] = events.States().Node.Rejected
 	if err := utils.WaitForCondition(func() bool {
-		return sn1.getNodeState() == string(events.States().Node.Healthy) &&
-			sn2.getNodeState() == string(events.States().Node.Recovering)
-	}, time.Second, 5*time.Second); err != nil {
+		return reflect.DeepEqual(getNodeStates(schedulerNodes), expectedStates)
+	}, 100*time.Millisecond, 3*time.Second); err != nil {
 		t.Fatal("unexpected node states")
```
wilfred-s, line 177: same as above log the error to help find test failures

pkg/cache/context_recovery_test.go
yangwwei: Agree. it is fine to remove it.

TaoYang526:
Thanks @wilfred-s and @yangwwei for the review.
Already improved these logs in latest commit.

PR #102, opened 2020-04-09
[YUNIKORN-86] Set proper resource request and limit for YuniKorn pods

adamantal:


kingamarton:
I think we should add the resource requests to scheduler.yaml as well, to have the same behaviour in case of manual installation as well.

adamantal:
Thanks @kingamarton! Updated the PR accordingly.

Referring to the mail thread discussion: I updated the limits as @yangwwei suggested, and I made it configurable in the Helm chart as @TaoYang526 said.

I'll test it, but in the meantime please review it.

adamantal:
Tested locally (docker-desktop), and it looks good - see comment on jira.

yangwwei:
hi @adamantal mostly seems good to me. can we have the similar changes for `deployment/scheduler/scheduler-load.yaml`?

adamantal:
Updated `scheduler-load.yaml`.

PR #103, opened 2020-04-09
[YUNIKORN-87] Add predicate check for reservation

wilfred-s:
Add a flag to distinguish between allocation and reservation for
predicate checks. A reservation predicate run will not update the event
system in k8s. A new limited ist of predicates to run for reservations
only is created. The list of predicates for reservations must be
contained in the default and or configured list.

DO NOT COMMIT
go.mod changes must be added waiting for apache/incubator-yunikorn-scheduler-interface#11


yangwwei:
I have two concerns about this PR.

1. Predicates are configurable, meaning that we can specify a list of whitelist predicate names from CLI. Then we only run those predicates. We should continue to honor that.
2. The `newPredictorInternal()` function first set the `schedulerPolicy`, then it runs the init code: 
```
// init all predicates
p.init()
// generate predicate functions
p.populatePredicateFunc(*args)
// generate predicate meta producer
p.populatePredicateMetaProducer(*args)
```
this PR creates a new schedulerPolicy for reservation, but not run init code for them. I doubt this can work. 

Based on these 2 points, I think we can still have 1 `schedulerPolicy` like today. Then add a white-list name of predicates for reservation and stored that in a `[]string`. Then when we iterate all predicates, we only execute those functions that appear in this list.

wilfred-s:
The reservation predicate list already uses the configured policy and we do not want to re-init anything.

The list of reservation predicates is a part of the default policy and if the configured policy does not contain the reservation predicate it will not be called. The init of the policy sets up a list of functions during the registration and that same list is checked for the reservation predicates. This means that it already functions as a white list.

To make it clearer the reservation predicate list has been changed from a PredicatePolicy array to a string array. The behaviour has not changed.

PR #107, opened 2020-04-28
[YUNIKORN-55] Add pod labels and annotations to allocation ask attributes

adamantal:


pkg/common/resource.go
```diff
@@ -99,12 +103,38 @@ func getResource(resourceList v1.ResourceList) *si.Resource {
 	return resources.Build()
 }
 
-func CreateUpdateRequestForTask(appID, taskID string, resource *si.Resource) si.UpdateRequest {
+func createTagsForTask(pod *v1.Pod, taskID string) map[string]string {
+	tags := map[string]string{
+		"namespace": pod.Namespace,
```
yangwwei, line 108: I think we should add domain names for these tags.
For things from pod spec, I suggest we use the following keys:
- kubernetes.io/meta/podName
- kubernetes.io/meta/namespace
- kubernetes.io/label/some-label
- kubernetes.io/annotation/some-annotation

this is to give us the flexibility to add more tags in the future, we do have plan to introduce some yunikorn domain tags, and those will be like
- yunikorn.apache.org/group/some-tag
- yunikorn.apache.org/some-tag

this is using format `domain/group/key-name`. I do think we need to add `domain` and `group` as enum types in the scheduler-interface spec. So we can refer that in both core and the shim. Does this make sense to you?

pkg/common/resource.go
```diff
@@ -99,12 +103,38 @@ func getResource(resourceList v1.ResourceList) *si.Resource {
...
+		"podName": pod.Name,
+	}
+
+	// add Pod labels to Task tags
+	labels, err := json.Marshal(pod.Labels)
```
yangwwei, line 113: is `labels` a string?
I would prefer to continue to get key/value pairs and store them in the tags map.

pkg/common/resource.go
```diff
@@ -99,12 +103,38 @@ func getResource(resourceList v1.ResourceList) *si.Resource {
...
+	if err != nil {
+		log.Logger.Debug("Could not marshal pod labels", zap.String("taskID", taskID))
+	} else {
+		tags["labels"] = string(labels)
+	}
+
+	// add Pod annotations to Task tags
+	annotations, err := json.Marshal(pod.Annotations)
```
wilfred-s, line 121: This is dangerous: annotations have no limits on what they can contain as a value. They can contain large blobs of data. I am not sure that we just want to pull that in without some kind of vetting. 

I also doubt that we can do much with it when we get it into the scheduler.

pkg/common/resource.go
wilfred-s: Labels are key/value pairs with a limited char set content (alphanumerical, dash, underscore and dot)
key is string with a prefix part between 0 and 253 char followed by a slash and a name part which can be 63 char long.
value is a string of 63 char max

Adding them as a marshalled json means that we need to unmarshall on the other side. Keeping them as plain key value pairs is simpler but I do think we need to screen what we pull in.

pkg/common/resource.go
wilfred-s: Namespace is an application level tag already. I don't think we should allow different namespaces for pods within an application and thus not allow specifying the namespace on the ask. It will be really confusing.

The labels should also be screened and not just all added to the tags. The tags for an application and an allocation ask were added to the SI for placement and allocation support not to just provide further details on the object.

adamantal: Created a [PR|https://github.com/apache/incubator-yunikorn-scheduler-interface/pull/14] for the constants in scheduler-interface. It seems that these constants are not added to the package, so I also added them to common/constants.go.

adamantal: Removed the namespace from the allocation tags.

pkg/common/resource.go
```diff
@@ -19,11 +19,15 @@
 package common
 
 import (
+	"encoding/json"
+
+	"go.uber.org/zap"
```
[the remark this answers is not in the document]
adamantal: For me it passed.

Could you please provide the correct order of the imports?

pkg/common/resource.go
adamantal: Removed bulk add of the labels.

pkg/common/resource.go
adamantal: Removed.

pkg/common/resource.go
adamantal: Remove bulk add of all labels

adamantal:
Updated the PR with your review items.

Right now all the labels and annotation have been dropped off from the patch. This doesn't really make sense without at least a few of them, so @yangwwei @wilfred-s could you please list a few of them that we are going to need in the scheduler-core side?

Thanks!

adamantal:
Updated the PR with the offline comments. Please don't merge until YUNIKORN-135 and the SI part of this issue is pushed.

adamantal:
Using the constants from SI side. Please review https://github.com/apache/incubator-yunikorn-scheduler-interface/pull/20 first.

adamantal:
Recreated the diff because of YUNIKORN-153.

@yangwwei @wilfred-s could you please review?

yangwwei:
hi @adamantal  could you please rebase this as apache/incubator-yunikorn-scheduler-interface#20 now is merged.

adamantal:
Thanks @yangwwei for reminding me. Rebased the patch, pending on jenkins.

pkg/common/si_helper.go
```diff
@@ -19,16 +19,48 @@
...
+		tags["labels"] = string(labels)
+	}
+
+	// add Pod annotations to Task tags
+	annotations, err := json.Marshal(pod.Annotations)
+	if err != nil {
+		log.Logger.Debug("Could not marshal pod annotations", zap.String("taskID", taskID))
+	} else {
+		tags["annotations"] = string(annotations)
+	}
```
yangwwei, line 52: An earlier comment from @wilfred-s mentioned pulling in all annotations might be dangerous. Can we remove this?

pkg/common/si_helper.go
```diff
@@ -19,16 +19,48 @@
...
+		"podName": pod.Name,
+	}
+
+	// add Pod labels to Task tags
+	labels, err := json.Marshal(pod.Labels)
+	if err != nil {
+		log.Logger.Debug("Could not marshal pod labels", zap.String("taskID", taskID))
+	} else {
+		tags["labels"] = string(labels)
+	}
```
yangwwei, line 44: One of earlier comment I have was, instead of marshaling this to a single string, can we keep the label values as key-value pairs. So in tags, we are not expecting
```
labels:  <json values for all labels>
```

instead, we'd like to have
```
kubernetes.io/labe/<some-label-key-1>: <some-value-2>
kubernetes.io/labe/<some-label-key-2>: <some-value-2>
```

this is to differentiate other tags from YK itself.

pkg/common/si_helper.go
adamantal: All right. Removed it.

pkg/common/si_helper.go
adamantal: Fixed.

adamantal:
When reimplemented the patch, somehow some commits got lots (like use the constants from the SI change and to remove the annotations).
Readded the changes and fixed the review items.

adamantal:
Thanks for merging this @yangwwei

PR #110, opened 2020-05-09
[YUNIKORN-143] Travis CI integration for K8shim

yangwwei:


Makefile
```diff
@@ -48,7 +48,7 @@ endif
 # Image build parameters
 # This tag of the image must be changed when pushed to a public repository.
 ifeq ($(REGISTRY),)
-REGISTRY := yunikorn
+REGISTRY := apache
```
kingamarton, line 51: With this registry change we need to make sure that will cleanup the other images, otherwise it may cause a lot of misunderstandings.

Makefile
```diff
@@ -124,7 +124,7 @@ sched_image: scheduler
 	@coreSHA=$$(go list -m "github.com/apache/incubator-yunikorn-core" | cut -d "-" -f5) ; \
 	siSHA=$$(go list -m "github.com/apache/incubator-yunikorn-scheduler-interface" | cut -d "-" -f6) ; \
 	shimSHA=$$(git rev-parse --short=12 HEAD) ; \
-	docker build ./deployments/image/configmap -t ${REGISTRY}/yunikorn-scheduler-k8s:${VERSION} \
+	docker build ./deployments/image/configmap -t ${REGISTRY}/yunikorn:core-${VERSION} \
```
kingamarton, line 127: I think here we are building the scheduler image not the core. If we really need the rename please update the comment and the target name as well.

Makefile
```diff
@@ -149,18 +149,25 @@ admission: init
...
 
 # Build all images based on the production ready version
 .PHONY: image
 image: sched_image adm_image
 
+.PHONY: push
+push: image
+	@echo "push docker images"
+	echo "${DOCKER_PASSWORD}" | docker login -u "${DOCKER_USERNAME}" --password-stdin
+	docker push ${REGISTRY}/yunikorn:core-${VERSION}
```
kingamarton, line 163: Why are we renaming the scheduler to core?

Makefile
wilfred-s: We need to file a jira to move the main build from the shim to the core. Including this docker build. That requires a lot of work moving the entrypoint and other code from one repo to the other and update the startup code to check for an integrated k8shim. This is not something for 0.9 but it is a blocker for a 1.0 release.

The changes in the name are needed to allow us to use _yunikorn_ as a container under the apache hub account to store all our images

.travis.yml
```diff
@@ -0,0 +1,35 @@
...
+  - GO111MODULE=on
+
+install: true
+
+jobs:
+  include:
+    - stage: unit test
+      script:
+      - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.22.2
+      - make common-check-license
```
wilfred-s, line 25: this needs to move to the same naming as for the core, it also requires an update of the target
we should at least make the target name change so it does not need to be done later: license-check

.travis.yml
```diff
@@ -0,0 +1,35 @@
...
+  - 'docker'
+
+env:
+  - GO111MODULE=on
+
+install: true
+
+jobs:
+  include:
+    - stage: unit test
```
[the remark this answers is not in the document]
yangwwei: Make sense, let's rename to pre-commit checks

Makefile
yangwwei: Yes. We should clean up the non-apache one. Once we move everything to the apache repo.

Makefile
yangwwei: `yunikorn:scheduler-latest` might be a better name

.travis.yml
```diff
@@ -0,0 +1,35 @@
+language: go
```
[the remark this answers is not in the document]
yangwwei: fixed

.travis.yml
yangwwei: Renamed to `license-check`

Makefile
wilfred-s: @kingamarton please file a jira for that so we keep track and clean up

kingamarton: Created YUNIKORN-154 for this purpose.

PR #113, opened 2020-05-13
[YUNIKORN-117] Create event cache for queue and application events

adamantal:
A way to expose the `EventStore` to the shim


yangwwei:
I don't think we should expose EventStore to the shim. Instead, can we add another plugin: https://github.com/apache/incubator-yunikorn-core/blob/master/pkg/plugins/plugins.go?

PR #115, opened 2020-05-14
[YUNIKORN-125] Add missing license headers in k8shim

HuangTing-Yao:


go.mod
```diff
@@ -1,3 +1,21 @@
+//
```
sunilgovind, line 1: once we add a new dependency, this file will be auto changed by go modules. 
in my local testing, i did go mod tidy as well. looks ok so far.

wilfred-s: Mod files support comments in the file based on the specs. It should not remove any comment lines when you run a `go mod`
See: https://golang.org/cmd/go/#hdr-The_go_mod_file

.asf.yaml
```diff
@@ -1,4 +1,29 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 github:
+  description: "Apache YuniKorn K8shim"
+  homepage: https://yunikorn.apache.org/
+  labels:
```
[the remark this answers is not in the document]
wilfred-s: These were the labels I added as a test on the core, we can add more, file a follow up jira to set them please.

PR #116, opened 2020-05-14
[YUNIKORN-137] Update examples and docs to use docker images from ASF repo

kingamarton:


kingamarton:
> Changes look good except for the removal of the landmark directory. This directory is required for an off-line integration test we run and without that it will not work.
> We need to keep them around. Updates will follow as soon as we have the new naming committed here.

I forgot to add it back. Not it is there and made the changes related the naming change as well.

PR #117, opened 2020-05-14
YUNIKORN-158. Admission controller deployment file should use the same image version as the scheduler

yangwwei:
Fix the hardcoded value and make sure the version is always aligned

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,7 +150,7 @@ adm_image: admission
 	@echo "building admission controller docker images"
 	@cp ${ADMISSION_CONTROLLER_BIN_DIR}/${POD_ADMISSION_CONTROLLER_BINARY} ./deployments/image/admission
 	docker build ./deployments/image/admission -t ${REGISTRY}/yunikorn:admission-${VERSION} \
-	  --build-arg DOCKER_IMAGE_REGISTRY=${REGISTRY} DOCKER_IMAGE_VERSION=${VERSION}
+	  --build-arg DOCKER_IMAGE_REGISTRY=${REGISTRY} --build-arg DOCKER_IMAGE_VERSION=${VERSION}
 	@rm -f ./deployments/image/admission/${POD_ADMISSION_CONTROLLER_BINARY}
 
 # Build all images based on the production ready version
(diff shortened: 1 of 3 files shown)


yangwwei:
@TaoYang526 , @sunilgovind could you help to review this?

Makefile
```diff
@@ -149,7 +149,8 @@ admission: init
 adm_image: admission
 	@echo "building admission controller docker images"
 	@cp ${ADMISSION_CONTROLLER_BIN_DIR}/${POD_ADMISSION_CONTROLLER_BINARY} ./deployments/image/admission
-	docker build ./deployments/image/admission -t ${REGISTRY}/yunikorn:admission-${VERSION}
+	docker build ./deployments/image/admission -t ${REGISTRY}/yunikorn:admission-${VERSION} \
+	  --build-arg DOCKER_IMAGE_REGISTRY=${REGISTRY} DOCKER_IMAGE_VERSION=${VERSION}
```
TaoYang526, line 153: It seems that --build-arg can only support one argument so that we can add another "--build-arg" before "DOCKER_IMAGE_VERSION=${VERSION}".

Makefile
```diff
@@ -149,7 +149,8 @@ admission: init
 adm_image: admission
 	@echo "building admission controller docker images"
 	@cp ${ADMISSION_CONTROLLER_BIN_DIR}/${POD_ADMISSION_CONTROLLER_BINARY} ./deployments/image/admission
-	docker build ./deployments/image/admission -t ${REGISTRY}/yunikorn:admission-${VERSION}
+	docker build ./deployments/image/admission -t ${REGISTRY}/yunikorn:admission-${VERSION} \
```
TaoYang526, line 152: Just noticed that admission and scheduler are sharing the same repo and type(scheduler/admission) is included in the version, is this a common solution to make sure they are coming from the same version? 
I noticed repository is not consistent with the updated scheduler repo(${REGISTRY}/yunikorn:scheduler-${VERSION} ), we should update helm-charts/values.yaml as the scheduler repo changed, right?

Makefile
yangwwei: Ah, good catch. I only tested 1 arg I guess. Let me fix this, thanks!

Makefile
yangwwei: yes, we need to have them using same repo/version. we have some other PRs pending to merge for the repo updates. basically, we are migrating all docker images to ASF repo.

PR #120, opened 2020-05-21
[YUNIKORN-173] Generates one default application ID per namespace in the admission controller

yangwwei:
See more in https://issues.apache.org/jira/browse/YUNIKORN-173.

### DIFF
--- a/pkg/plugin/admissioncontrollers/webhook/admission_controller.go
+++ b/pkg/plugin/admissioncontrollers/webhook/admission_controller.go
@@ -157,7 +157,7 @@ func updateLabels(namespace string, pod *v1.Pod, patch []patchOperation) []patch
 		if _, ok := existingLabels[common.LabelApplicationID]; !ok {
 			// if app id not exist, generate one
 			// for each namespace, we group unnamed pods to one single app
-			// application ID convention: __${NAMESPACE}_app_default
+			// application ID convention: ${AUTO_GEN_PREFIX}-${NAMESPACE}-${AUTO_GEN_SUFFIX}
 			generatedID := generateAppID(namespace)
 			log.Logger.Debug("adding application ID",
 				zap.String("generatedID", generatedID))
--- a/pkg/plugin/admissioncontrollers/webhook/admission_controller.go
+++ b/pkg/plugin/admissioncontrollers/webhook/admission_controller.go
@@ -24,8 +24,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


yangwwei:
Tested locally, with or without namespace specified. the name generated will be like:

```
yunikorn-default-autogen
yunikorn-somenamespace-autogen
```

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -125,20 +129,23 @@ func updateSchedulerName(patch []patchOperation) []patchOperation {
...
 
-// the generated ID is using [PodName]_[Timestamp] naming convention.
-// the generated ID should be unique even the pod name is same (differ on nano time),
+// generate appID based on the namespace value,
 // and the max length of the ID is 63 chars.
-func generateAppID(podName string) string {
-	timestamp := strconv.FormatInt(time.Now().UnixNano(), 10)
-	generatedID := fmt.Sprintf("%.43s_%.19s", podName, timestamp)
-	return generatedID
+func generateAppID(namespace string) string {
```
wilfred-s, line 134: Do we want to keep both options? One based on podName and the other based on namespace?
We have a config object and can then switch giving us the choice between namespace based and just on pod name. Seems a waste to throw away that option.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -149,16 +156,9 @@ func updateLabels(pod *v1.Pod, patch []patchOperation) []patchOperation {
...
-			podName := "unknown"
-			if pod.Name != "" {
-				podName = pod.Name
-			} else if pod.GenerateName != "" {
-				podName = pod.GenerateName
-			}
-
-			generatedID := generateAppID(podName)
+			// for each namespace, we group unnamed pods to one single app
+			// application ID convention: __${NAMESPACE}_app_default
```
wilfred-s, line 160: This does not line up with the test and what we generate a `prefix`-${NAMESPACE}-`suffix`

yangwwei: Fixed

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
yangwwei: That will be a bit hard to accomplish. We don't have an option to config the admission controller's behavior today.

PR #122, opened 2020-05-21
[YUNIKORN-171] Add example and document about how to run kubeflow with yunikorn

HuangTing-Yao:
Add a tfjob example for yunikorn.

### DIFF
--- a/deployments/examples/README.md
+++ b/deployments/examples/README.md
@@ -45,7 +45,7 @@ Deployment files for the driver and executor:
 * [driver](spark30/driver.yaml).
 * [executor](spark30/executor.yaml).
 
-## tfjob
+## Tensorflow job
 A simple example for [kubeflow/tensorflow](./tfjob/tf-job-mnist.yaml) image.
 
 ## volumes
--- a/deployments/examples/README.md
+++ b/deployments/examples/README.md
@@ -47,6 +47,8 @@ Deployment files for the driver and executor:
 
 ## Tensorflow job
 A simple example for [kubeflow/tensorflow](./tfjob/tf-job-mnist.yaml) image.
+In this example it will run a distributed mnist model for e2e test, for full more detail see the [dist-mnist](https://github.com/kubeflow/tf-operator/tree/master/examples/v1/dist-mnist) section.
+As we can schedule tfjob mean we can support kubeflow.
 
 ## volumes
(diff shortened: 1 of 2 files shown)


deployments/examples/README.md
```diff
@@ -45,6 +45,9 @@ Deployment files for the driver and executor:
 * [driver](spark30/driver.yaml).
 * [executor](spark30/executor.yaml).
 
+## tfjob
```
kingamarton, line 48: Please use the full name instead of abbreviation: tfjob -> TensorFlow job, or Kubeflow

deployments/examples/README.md
```diff
@@ -45,6 +45,9 @@ Deployment files for the driver and executor:
...
+A simple example for [kubeflow/tensorflow](./tfjob/tf-job-mnist.yaml) image.
```
yangwwei, line 49: Can we add a bit more doc about what does this job do?

deployments/examples/tfjob/tf-job-mnist.yaml
```diff
@@ -0,0 +1,51 @@
...
+  namespace: kubeflow
+spec:
+  tfReplicaSpecs:
+    PS:
+      replicas: 2
+      restartPolicy: Never
+      template:
+        metadata:
+          labels:
+            applicationId: "tf_ps_20200521_001"
```
yangwwei, line 31: can we use the same `applicationId` for this job?
both ps and workers should belong to one same job.

HuangTing-Yao: Yes, we should use same `applicationId` for this job.

deployments/examples/README.md
HuangTing-Yao: I have already add doc on it, do we need more?

yangwwei:
> Overall looks good.
> If its possible, I would like to have a Readme or Application section in landing Read me doc about the running apps YuniKorn.
> Like Spark, Flink, Kubeflow etc called out explicitly

Yeah, that's a good suggestion. I do think our doc currently is not very easy to follow. Maybe we should have a separate section for this. We need to org our doc better than what we have now.

deployments/examples/README.md
yangwwei: Sorry I think I've missed that in last review.

PR #123, opened 2020-05-22
[YUNIKORN-176] schedulerCache might become inconsistent sometimes depending on the ordering of the events

yangwwei:
See more description in https://issues.apache.org/jira/browse/YUNIKORN-176

### DIFF
--- a/pkg/cache/external/scheduler_cache.go
+++ b/pkg/cache/external/scheduler_cache.go
@@ -93,10 +93,12 @@ func (cache *SchedulerCache) AddNode(node *v1.Node) {
 	_, ok := cache.nodesMap[node.Name]
 	if !ok {
 		n := schedulernode.NewNodeInfo()
-		// API call always returns nil, never an error
-		//nolint:errcheck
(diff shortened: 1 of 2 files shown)


yangwwei:
hmm, this is interesting. I did 1 line code fix, adding an extra UT file, but the coverage decreased ...

wilfred-s:
for the test coverage: yes. Files without tests do not show up in the coverage data and are thus not calculated as lines to include. Bit strange but it is like that.

yangwwei:
Hi @wilfred-s 

> The log of the error is dead code, the comment and lint message is clearer than providing a false sense that it returns an error. Can we at least document that this is the case to prevent looking for the message in the logs?

I have added more comments to explain this behavior. I prefer to keep the err check code around, in the case that in the future this API returns us an error. So we don't need to get back and revise this code.

yangwwei:
I think we are good on this one. Merging this so we can start to build a docker image. cc @wilfred-s , please let me know if you feel otherwise. Thanks!

PR #124, opened 2020-05-22
[YUNIKORN-158] Admission controller deployment file should use the same version as the scheduler.

yangwwei:
This is to fix https://issues.apache.org/jira/browse/YUNIKORN-158

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,12 +120,13 @@ sched_image: scheduler
 	@cp ${RELEASE_BIN_DIR}/${BINARY} ./deployments/image/configmap
 	@mkdir -p ./deployments/image/configmap/admission-controller-init-scripts
 	@cp -r ./deployments/admission-controllers/scheduler/*  deployments/image/configmap/admission-controller-init-scripts/
+	@sed -i -e 's@\$${REGISTRY}@'"${REGISTRY}"'@g' deployments/image/configmap/admission-controller-init-scripts/templates/server.yaml.template
+	@sed -i -e 's@\$${VERSION}@'"${VERSION}"'@g' deployments/image/configmap/admission-controller-init-scripts/templates/server.yaml.template
 	@sed -i'.bkp' 's/clusterVersion=.*"/clusterVersion=${VERSION}"/' deployments/image/configmap/Dockerfile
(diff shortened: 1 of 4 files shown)


yangwwei:
> Can we further simplify this by making the change before we copy the files into image?
> 
> ```
> sed -i -e  's/\${REGISTRY}/'"${REGISTRY}"'/' ./deployments/image/configmap/admission-controller-init-scripts/templates/server.yaml.template
> sed -i -e  's/\${VERSION}/'"${VERSION}"'/' ./deployments/image/configmap/admission-controller-init-scripts/templates/server.yaml.template
> ```
> 
> That will keep all changes down to the template and the Makefile.
> It will also centralise the locations where we make these version related changes instead of pushing them to multiple places.

hi @wilfred-s 

The problem with this approach is, every time we run a `make image`, it will modify the file: `./deployments/image/configmap/admission-controller-init-scripts/templates/server.yaml.template`, this is a source code file tracked by git. Which means we need to revert the changes after `make` is done. I don't think that is simpler.

Having this a configurable env variable in the docker file is simple, and we can pass in the values via standard docker build command, I think that is also very flexible.

wilfred-s:
> The problem with this approach is, every time we run a `make image`, it will modify the file: `./deployments/image/configmap/admission-controller-init-scripts/templates/server.yaml.template`, this is a source code file tracked by git. Which means we need to revert the changes after `make` is done. I don't think that is simpler.

That is not correct, in the makefile on the lines 121-122 we create the directory and copy the files from the tracked location:
```
	@mkdir -p ./deployments/image/configmap/admission-controller-init-scripts
	@cp -r ./deployments/admission-controllers/scheduler/*  deployments/image/configmap/admission-controller-init-scripts/
```
Then on line 135 we remove the directory with all its content again as soon we are done with the image creation:
```
	@rm -rf ./deployments/image/configmap/admission-controller-init-scripts/
```
There is thus no reason to do this in the docker build.

yangwwei:
> That is not correct, in the makefile on the lines 121-122 we create the directory and copy the files from the tracked location

You are right. We did a copy. Then that's fine. I just updated the patch for this change.

PR #125, opened 2020-05-23
[YUNIKORN-185] K8shim logs cleanup

yangwwei:
This PR does a few optimizations for the logs
- remove the predicate errors from the logging, we can leverage the event system, no need to log in the file
- for state transition logs, remove dup records
- leverage zap post-state hook to log event transitions, added unit tests

### DIFF
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -347,11 +347,9 @@ func (app *Application) handleCompleteApplicationEvent(event *fsm.Event) {
 }
 
 func (app *Application) enterState(event *fsm.Event) {
-	log.Logger.Info("app state",
-		zap.String("currentState", event.Src),
-		zap.String("targetState", event.Dst),
-		zap.String("event", event.Event),
-		zap.String("appID", app.applicationID),
-		zap.String("queue", app.queue),
-		zap.String("currentAppState", app.GetApplicationState()))
+	log.Logger.Debug("shim app state transition",
+		zap.String("app", app.applicationID),
+		zap.String("source", event.Src),
+		zap.String("destination", event.Dst),
+		zap.String("event", event.Event))
 }
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -123,18 +123,11 @@ func (app *Application) handle(ev events.ApplicationEvent) error {
 	//    because the lock is already held here.
 	app.lock.Lock()
 	defer app.lock.Unlock()
-	log.Logger.Debug("application state transition",
-		zap.String("appID", app.applicationID),
-		zap.String("preState", app.sm.Current()),
-		zap.String("pendingEvent", string(ev.GetEvent())))
 	err := app.sm.Event(string(ev.GetEvent()), ev.GetArgs()...)
 	// handle the same state transition not nil error (limit of fsm).
 	if err != nil && err.Error() != "no transition" {
 		return err
 	}
-	log.Logger.Debug("application state transition",
-		zap.String("appID", app.applicationID),
-		zap.String("postState", app.sm.Current()))
 	return nil
 }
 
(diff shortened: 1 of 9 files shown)


yangwwei:
I have to remove the unit tests for those event loggings
because of the race condition.

pkg/cache/application.go
```diff
@@ -355,7 +348,10 @@ func (app *Application) handleCompleteApplicationEvent(event *fsm.Event) {
 
 func (app *Application) enterState(event *fsm.Event) {
 	log.Logger.Info("app state",
+		zap.String("currentState", event.Src),
+		zap.String("targetState", event.Dst),
+		zap.String("event", event.Event),
 		zap.String("appID", app.applicationID),
 		zap.String("queue", app.queue),
-		zap.String("state", app.GetApplicationState()))
+		zap.String("currentAppState", app.GetApplicationState()))
```
wilfred-s, line 356: This is a double up: we have current and target state already (currentAppState)
Can we also bring this inline with the way this is done in the core? Log at a debug level and with minimal info and the same order: appID, source, destination, and event.

pkg/cache/application_test.go
```diff
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"gotest.tools/assert"
+
```
wilfred-s, line 26: spurious new line

pkg/cache/node.go
```diff
@@ -245,3 +239,12 @@ func (n *SchedulerNode) canHandle(ev events.SchedulerNodeEvent) bool {
...
 	return n.fsm.Can(string(ev.GetEvent()))
 }
+
+func (n *SchedulerNode) enterState(event *fsm.Event) {
+	log.Logger.Info("node state",
+		zap.String("currentState", event.Src),
+		zap.String("targetState", event.Dst),
+		zap.String("event", event.Event),
+		zap.String("nodeName", n.name),
+		zap.String("currentNodeState", n.getNodeState()))
```
wilfred-s, line 249: This is a double up: we have current and target state already (currentAppState).
Can we also bring this inline with the way this is done in the core? Log at a debug level and with same info order: node, source, destination, and event.

pkg/cache/task.go
```diff
@@ -415,8 +408,11 @@ func (task *Task) sanityCheckBeforeScheduling() error {
...
 func (task *Task) enterState(event *fsm.Event) {
 	log.Logger.Info("task state",
+		zap.String("currentState", event.Src),
+		zap.String("targetState", event.Dst),
+		zap.String("event", event.Event),
 		zap.String("appID", task.applicationID),
 		zap.String("taskID", task.taskID),
 		zap.String("taskAlias", task.alias),
-		zap.String("state", task.GetTaskState()))
+		zap.String("currentTaskState", task.GetTaskState()))
```
wilfred-s, line 417: This is a double up: we have current and target state already (currentAppState).
Log at a debug level and with same info order: taskID, source, destination, and event. If we really need the appID to track it as the task ID is not unique then we should add that back.

pkg/cache/task_test.go
```diff
@@ -23,13 +23,13 @@ import (
 
 	"gotest.tools/assert"
```
wilfred-s, line 25: remove the empty line

pkg/shim/scheduler.go
```diff
@@ -282,6 +268,14 @@ func (ss *KubernetesShim) run() {
 	ss.apiFactory.Start()
 }
 
+func (ss *KubernetesShim) enterState(event *fsm.Event) {
+	log.Logger.Info("scheduler shim state",
+		zap.String("currentState", event.Src),
+		zap.String("targetState", event.Dst),
+		zap.String("event", event.Event),
+		zap.String("currentSchedulerState", ss.GetSchedulerState()))
+}
```
wilfred-s, line 277: This is a double up: we have current and target state already (currentAppState).
Log at a debug level and with same info order: shim identification, source, destination and event.

pkg/cache/application.go
yangwwei: Make sense. Align with how the state transition is logged in core.

pkg/cache/task.go
yangwwei: We need appID, taskID and alias for tracking purpose, revised to

```
	log.Logger.Debug("shim task state transition",
		zap.String("app", task.applicationID),
		zap.String("task", task.taskID),
		zap.String("taskAlias", task.alias),
		zap.String("source", event.Src),
		zap.String("destination", event.Dst),
		zap.String("event", event.Event))
```

pkg/cache/application_test.go
yangwwei: Removed

pkg/cache/task_test.go
yangwwei: removed

PR #128, opened 2020-05-27
[YUNIKORN-117] Create event cache for queue and application events

adamantal:
Shim side PR

### DIFF
--- a/pkg/callback/scheduler_callback.go
+++ b/pkg/callback/scheduler_callback.go
@@ -20,6 +20,7 @@ package callback
 
 import (
 	"fmt"
+	"strings"
 
 	"go.uber.org/zap"
 	"k8s.io/api/core/v1"
@@ -146,30 +147,35 @@ func (callback *AsyncRMCallback) ReSyncSchedulerCache(args *si.ReSyncSchedulerCa
 }
 
 // this callback implement scheduler plugin interface EventPlugin.
-func (callback *AsyncRMCallback) SendEvent(event *si.EventMessage) error {
-	log.Logger.Debug("processing event")
-	reason := event.Reason
-	msg := event.Message
+func (callback *AsyncRMCallback) SendEvent(eventMessages []*si.EventMessage) error {
+	errors := make([]string, 0)
+	if len(eventMessages) > 0 {
+		log.Logger.Debug(fmt.Sprintf("processing %d events", len(eventMessages)))
+		for _, event := range eventMessages {
+			reason := event.Reason
+			msg := event.Message
 
-	switch event.Type {
-	case si.EventMessage_APP:
-		// until we don't have app CRD let's expose app event to all its pods (asks)
-		appId := event.ID
-		pods, err := cache.GetAllPods(callback.context, appId)
-		if err != nil {
-			return fmt.Errorf("could not find application: %s", appId)
+			switch event.Type {
+			case si.EventMessage_APP:
+				// until we don't have app CRD let's expose app event to all its pods (asks)
+				// pending on YUNIKORN-170
+				appId := event.ID
+				pods, err := cache.GetAllPods(callback.context, appId)
+				if err != nil {
+					errors = append(errors, fmt.Sprintf("could not find application: %s", appId))
+					continue
+				}
+				for _, pod := range pods {
+					log.Logger.Info("Emitting event", zap.String("pod", pod.ObjectMeta.Name))
+					events.GetRecorder().Event(pod, v1.EventTypeWarning, reason, msg)
+				}
+			case si.EventMessage_REQUEST:
+				// TODO search pod and emit event there
+				errors = append(errors, fmt.Sprintf("processing request event is not implemented yet: %s", event))
+			default:
+				errors = append(errors, fmt.Sprintf("could not process unknown event type: %s", event))
+			}
 		}
-		for _, pod := range pods {
-			log.Logger.Debug("Emitting ")
-			// TODO set event type: normal/warning
-			events.GetRecorder().Event(pod, v1.EventTypeWarning, reason, msg)
-		}
-	case si.EventMessage_REQUEST:
-
-		// TODO search pod and emit event there
-		return fmt.Errorf("processing request event is not implemented yet: %s", event)
-	default:
-		return fmt.Errorf("could not process unknown event type: %s", event)
 	}
-	return nil
+	return fmt.Errorf(strings.Join(errors, "\n"))
 }
--- a/pkg/callback/scheduler_callback.go
+++ b/pkg/callback/scheduler_callback.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


pkg/cache/external/scheduler_cache.go
```diff
@@ -65,6 +65,10 @@ func (cache *SchedulerCache) GetNodesInfoMap() map[string]*schedulernode.NodeInf
 	return cache.nodesMap
 }
 
+func (cache *SchedulerCache) GetPodsInfoMap() map[string]*v1.Pod {
+	return cache.podsMap
+}
+
```
yangwwei, line 71: Redundant?

pkg/callback/scheduler_callback.go
```diff
@@ -143,3 +145,48 @@ func (callback *AsyncRMCallback) ReSyncSchedulerCache(args *si.ReSyncSchedulerCa
 	}
 	return nil
 }
+
+// this callback implement scheduler plugin interface EventPlugin.
+func (callback *AsyncRMCallback) SendEvent(eventMessages []*si.EventMessage) error {
+	errors := make([]string, 0)
+	if len(eventMessages) > 0 {
+		log.Logger.Debug(fmt.Sprintf("processing %d events", len(eventMessages)))
+		for _, event := range eventMessages {
```
yangwwei, line 154: For better modulization. I propose to add a publishEvent function for `task.go` (and similarly we can add for `application.go` or `node.go` in the future). So we can simply find out the task/app/node, and call `publishEvent(eventMessage *si.EventMessage)`

pkg/callback/scheduler_callback.go
```diff
@@ -143,3 +145,48 @@ func (callback *AsyncRMCallback) ReSyncSchedulerCache(args *si.ReSyncSchedulerCa
...
+			reason := event.Reason
+			msg := event.Message
+
+			switch event.Type {
+			case si.EventMessage_REQUEST:
+				taskAppID := event.ID
+				parts := strings.Split(taskAppID, ",")
```
yangwwei, line 161: Instead of using a comma, can we use the naming convention like:
```
queueName/appID/taskID
```

this is more like a full-qualified name for an ID. If that makes sense.

adamantal: Modified accordingly.

pkg/cache/external/scheduler_cache.go
adamantal: Removed.

pkg/callback/scheduler_callback.go
adamantal: Done.

adamantal:
Fixed review items.

pkg/cache/task.go
```diff
@@ -415,3 +416,21 @@ func (task *Task) enterState(event *fsm.Event) {
 		zap.String("destination", event.Dst),
 		zap.String("event", event.Event))
 }
+
+func PublishTaskEvent(taskID, reason, msg string, app interfaces.ManagedApp) error {
+	log.Logger.Info("taskID", zap.String("taskID", taskID))
```
sunilgovind, line 421: I am in double mind here.
1. taskID alone is not enough here, We may need more context here for helping to debug later.
2. May be its a debug log itself?

pkg/cache/task.go
```diff
@@ -415,3 +416,21 @@ func (task *Task) enterState(event *fsm.Event) {
...
+	pod := task.GetTaskPod()
+	if pod == nil {
+		return fmt.Errorf("could not obtain %s task's pod", taskID)
+	}
+
+	// TODO remove this
+	log.Logger.Debug("Emitting event", zap.String("pod name", pod.ObjectMeta.Name), zap.String("reason", reason), zap.String("message", msg))
+	events.GetRecorder().Event(pod, v1.EventTypeWarning, reason, msg)
+
+	return nil
```
sunilgovind, line 435: Do we need to return here explicitly?

pkg/callback/scheduler_callback.go
```diff
@@ -143,3 +144,35 @@ func (callback *AsyncRMCallback) ReSyncSchedulerCache(args *si.ReSyncSchedulerCa
 	}
 	return nil
 }
+
+// this callback implement scheduler plugin interface EventPlugin.
+func (callback *AsyncRMCallback) SendEvent(eventRecords []*si.EventRecord) error {
```
sunilgovind, line 149: Name is a bit confusing, is it SendEvent here?

pkg/callback/scheduler_callback.go
```diff
@@ -143,3 +144,35 @@ func (callback *AsyncRMCallback) ReSyncSchedulerCache(args *si.ReSyncSchedulerCa
...
+			case si.EventRecord_APP:
+				// until we don't have app CRD let's expose app event to all its pods (asks)
+				// pending on YUNIKORN-170
+				errors = append(errors, fmt.Sprintf("processing app event is not implemented yet: %s", record))
+			default:
+				errors = append(errors, fmt.Sprintf("could not process unknown event type: %s", record))
+			}
+		}
+	}
+	return fmt.Errorf(strings.Join(errors, "\n"))
```
sunilgovind, line 177: A naive question, sorry for that. If there are no errors here, how the return will look like?

pkg/cache/task.go
adamantal: I used for debugging this patch, but it makes sense to print this information in debug level.

adamantal: Actually a bit later, we already have a debug log. I added a more meaningful message to that.

pkg/cache/task.go
adamantal: We have to return an `error` object, and in the good case it's a `nil` object. So I think, we should return here indeed.

pkg/callback/scheduler_callback.go
adamantal: This function is part of the scheduler callback, it sends the `EventRecords` object to the shim. I think the `SendEvent` is concise and descriptive. What do you suggest?

pkg/callback/scheduler_callback.go
adamantal: It will return an empty string. I'll create a UT to cover that case.

pkg/callback/scheduler_callback.go
```diff
@@ -143,3 +144,35 @@ func (callback *AsyncRMCallback) ReSyncSchedulerCache(args *si.ReSyncSchedulerCa
...
+
+				err := cache.PublishTaskEvent(taskID, reason, msg, app)
+				if err != nil {
+					errors = append(errors, err.Error())
+				}
+				log.Logger.Debug("event emitted")
+			case si.EventRecord_APP:
+				// until we don't have app CRD let's expose app event to all its pods (asks)
+				// pending on YUNIKORN-170
+				errors = append(errors, fmt.Sprintf("processing app event is not implemented yet: %s", record))
```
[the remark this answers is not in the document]
yangwwei: agree

PR #132, opened 2020-06-02
[YUNIKORN-200] K8shim reports resource quota based on namespace annotations

yangwwei:


pkg/cache/context.go
```diff
@@ -426,6 +428,28 @@ func (ctx *Context) NotifyTaskComplete(appID, taskID string) {
...
+	}
+
+	nsLister := ctx.apiProvider.GetAPIs().NamespaceInformer.Lister()
+	namespaceObj, err := nsLister.Get(namespace)
+	if err != nil {
+		// every app should belong to a namespace,
+		// if we cannot list the namespace here, probably something is wrong
+		// log an error here and do not add the app to cache
+		log.Logger.Error("failed to get app namespace", zap.Error(err))
+		return nil
```
wilfred-s, line 447: 1. I do not see the error being logged when the namespace is empty and we also return a nil there. Does that mean we need a message there too?
2. I also do not see how returning a nil at this point distinguishes from no tags being found and returning a nil. You cannot stop adding the app based on a nil being returned

pkg/common/constants.go
```diff
@@ -29,6 +29,9 @@ const LabelApplicationID = "applicationId"
 const LabelQueueName = "queue"
 const ApplicationDefaultQueue = "root.sandbox"
 const DefaultPartition = "default"
+const AppTagNamespace = "namespace"
+const AppTagNamespaceResourceQuota = "namespace.resourcequota"
```
wilfred-s, line 33: These two should be moved to the SI as we use them in both the core and shim

pkg/common/resource.go
```diff
@@ -81,6 +83,37 @@ func GetNodeResource(nodeStatus *v1.NodeStatus) *si.Resource {
...
+	}
+
+	result := NewResourceBuilder()
+	if cpuStr != "" {
+		if vcore, err := resource.ParseQuantity(cpuStr); err == nil {
+			result.AddResource(CPU, vcore.MilliValue())
+		} else {
+			log.Logger.Error("failed to parse cpu resource",
+				zap.String("cpuStr", cpuStr),
+				zap.Error(err))
```
wilfred-s, line 100: If we fail to parse we should not return a resource object but a nil to clearly show that we're not enforcing the quota. Returning half the quota set is going to give inconsistent behaviour.

Negative values and zeros do not have to stop parsing at this point.

pkg/cache/context.go
```diff
@@ -434,6 +458,19 @@ func (ctx *Context) AddApplication(request *interfaces.AddApplicationRequest) in
 
 	ctx.lock.Lock()
 	defer ctx.lock.Unlock()
+
+	// add resource quota info as a app tag
+	log.Logger.Info("### retrieve namespace info 1")
+	if ns, ok := request.Metadata.Tags[common.AppTagNamespace]; ok {
+		log.Logger.Info("### retrieve namespace info 2", zap.String("namespace", ns))
+		resourceQuota := ctx.getNamespaceResourceQuota(ns)
+		if resourceQuota != nil && !common.IsZero(resourceQuota) {
```
wilfred-s, line 467: This does not capture the case that we set `mem=0` and `cpu=10` but we cannot apply that on the core side as it would stop scheduling for that namespace.
I think we need to catch this earlier than the core side.

pkg/common/utils/utils.go
```diff
@@ -22,12 +22,15 @@ import (
 	"strings"
 	"time"
 
+	"github.com/apache/incubator-yunikorn-k8shim/pkg/log"
```
wilfred-s, line 25: ordering of imports incorrect

pkg/common/utils/utils.go
```diff
@@ -102,6 +105,22 @@ func GetApplicationIDFromPod(pod *v1.Pod) (string, error) {
 		pod.Spec.String())
 }
 
+func GetNamespaceQuotaFromAnnotation(namespaceObj *v1.Namespace) *si.Resource {
+	log.Logger.Info("### GetNamespaceQuotaFromAnnotation")
+	// retrieve resource quota info from annotations
+	cpuQuota := namespaceObj.Annotations["yunikorn.apache.org/queue.max.cpu"]
+	memQuota := namespaceObj.Annotations["yunikorn.apache.org/queue.max.memory"]
```
wilfred-s, line 112: These should be referring to namespace not to queue. The fact that we map that in the core to a queue is beside the point. On the k8s side it is applying to a namespace.

pkg/common/utils/utils.go
```diff
@@ -102,6 +105,22 @@ func GetApplicationIDFromPod(pod *v1.Pod) (string, error) {
...
+
+	log.Logger.Info("### GetNamespaceQuotaFromAnnotation", zap.String("cpu", cpuQuota), zap.String("mem", memQuota))
+
+	// no quota found
+	if cpuQuota == "" && memQuota == "" {
+		return nil
+	}
+
+	return common.ParseResource(cpuQuota, memQuota)
```
wilfred-s, line 121: This is the point we need to build in sanity checks with clear logging as to what is checked and failed.:
- zero values for either setting
- negative values for either setting

Only return a resource that has both values set to a value larger than 0 or if only one is specified that one needs to be larger than 0. A nil should be returned in all other cases.

pkg/cache/context.go
yangwwei: > I do not see the error being logged when the namespace is empty and we also return a nil there. Does that mean we need a message there too?
Yes, I have added a DEBUG level message when the namespace is empty. 
We are not expecting this to happen, as each pod is always added in a namespace, `default` if not explicitly set. And I do not think an empty string is a valid namespace name. So when this happens, we should prompt a warning. 

> I also do not see how returning a nil at this point distinguishes from no tags being found and returning a nil. You cannot stop adding the app based on a nil being returned
For both cases, a nil means the resource quota is not set in the annotation. We only skip adding this info the app tag:
```
request.Metadata.Tags[common.AppTagNamespaceResourceQuota] = string(quotaStr)
```
but this won't stop us adding the app for scheduling.

pkg/common/constants.go
yangwwei: Yes, I am adding this in another PR.

pkg/common/utils/utils.go
yangwwei: Makes sense.

pkg/common/utils/utils.go
yangwwei: After offline discussion, we agree to have such sanity checks in core alone. We do not duplicate them in shim because currently there is no way to send feedback back to the client  event it fails.

pkg/common/resource.go
yangwwei: To make this be more consistent with the yunikorn-core, I think we can do that. Just updated the code for this. Thanks.

pkg/cache/context.go
wilfred-s: > but this won't stop us adding the app for scheduling.
Which means that the comment is wrong:
```
		// if we cannot list the namespace here, probably something is wrong
		// log an error here and do not add the app to cache
```
It says that you do not add the app to the cache, which is contradictory to what the behaviour is.

Beside that comment this part looks good

yangwwei: Fixed that comment.

yangwwei:
Logged https://issues.apache.org/jira/browse/YUNIKORN-206 for the constants refactoring follow up

PR #134, opened 2020-06-08
[YUNIKORN-170] Create YuniKorn App CRD

HuangTing-Yao:
Some code are generated by code-generator `release-1.14`


pkg/apis/yunikorn.apache.org/update-codegen.sh
```diff
@@ -0,0 +1,31 @@
...
+# generate the code with:
+# --output-base    because this script should also be able to run inside the vendor dir of
+#                  k8s.io/kubernetes. The output-base is needed for the generators to output into the vendor dir
+#                  instead of the $GOPATH directly. For normal projects this can be dropped.
+GO111MODULE=off
+bash k8s.io/code-generator/generate-groups.sh "all" \
+  github.com/apache/incubator-yunikorn-k8shim/pkg/client github.com/apache/incubator-yunikorn-k8shim/pkg/apis \
+  "yunikorn.apache.org:v1alpha1" \
+  #--output-base "$(dirname "${BASH_SOURCE[0]}")/../../.." \
+  #--go-header-file  ./pkg/apis/apache.org/boilerplate.go.txt
```
yangwwei, line 28: is this the script to run the code gen? Can you pls add some documents about how to run this?
what's the reason to comment out these 2 lines? For the header file, if this is commented out, how the license headers were added to those generated files?

pkg/apis/yunikorn.apache.org/update-codegen.sh
```diff
@@ -0,0 +1,31 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+#limitations under the License.
+#
+
+# generate the code with:
+# --output-base    because this script should also be able to run inside the vendor dir of
+#                  k8s.io/kubernetes. The output-base is needed for the generators to output into the vendor dir
+#                  instead of the $GOPATH directly. For normal projects this can be dropped.
+GO111MODULE=off
```
yangwwei, line 23: We should enable go mod, I am expecting this value to be true.
All of our repos are managed by go mod, with GO111MODULE set to true.

pkg/apis/yunikorn.apache.org/v1alpha1/type.go
```diff
@@ -0,0 +1,72 @@
...
+	SubmittedState        ApplicationStateType = "SUBMITTED"
+	RunningState          ApplicationStateType = "RUNNING"
+	CompletedState        ApplicationStateType = "COMPLETED"
+	FailedState           ApplicationStateType = "FAILED"
+	FailedSubmissionState ApplicationStateType = "SUBMISSION_FAILED"
+	PendingRerunState     ApplicationStateType = "PENDING_RERUN"
+	InvalidatingState     ApplicationStateType = "INVALIDATING"
+	SucceedingState       ApplicationStateType = "SUCCEEDING"
+	FailingState          ApplicationStateType = "FAILING"
+	UnknownState          ApplicationStateType = "UNKNOWN"
```
yangwwei, line 59: For the application state, currently, shim and core maintain different states. But moving on we should get them to be consistent. And the core should be the source of truth.
Please refer to https://github.com/apache/incubator-yunikorn-core/blob/f4ea4502a52e06e64e5abd0c47fedf2e08fe12e3/pkg/cache/application_state.go#L53-L61 to get the supported states by far.

pkg/apis/yunikorn.apache.org/v1alpha1/type.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+type ApplicationStatus struct {
+	AppStatus ApplicationStateType `json:"applicationState,omitempty"`
```
yangwwei, line 63: I think we need to include 2 more fields for app status:

```
type ApplicationStatus struct {
	State ApplicationStateType
	Message string // some arbitrary message to explain what happens
         LastUpdate time // last app state update time
}
```

yangwwei:
Can the code-gen be integrated into the Makefile?
I think we should try that, and add some doc directly in the Makefile if whoever wants to re-run the code gen.

pkg/apis/yunikorn.apache.org/update-codegen.sh
HuangTing-Yao: According https://github.com/kubernetes/code-generator/issues/69
I turn GO111MODULE off.
But I think we can turn on the GO111MODULE

pkg/apis/yunikorn.apache.org/update-codegen.sh
HuangTing-Yao: About the license headers, code generator will read the `boilerplate.go.txt` which in `k8s.io/code-generator/hack` first. Base on my test, the `--go-header-file` won't work.
The `--output-base` part, base on the description "because this script should also be able to run inside the vendor dir of k8s.io/kubernetes. The output-base is needed for the generators to output into the vendor dir instead of the $GOPATH directly. For normal projects this can be dropped." I don't think we need it.

HuangTing-Yao:
@sunilgovind  I just refer from https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/tree/master/pkg/apis/sparkoperator.k8s.io

pkg/apis/yunikorn.apache.org/v1alpha1/type.go
```diff
@@ -0,0 +1,71 @@
...
+	Queue             string `json:"queue"`
+	MaxPendingSeconds int32  `json:"maxPendingSeconds,omitempty"`
+}
+
+//Status part
+
+type ApplicationStateType string
+
+const (
+	NewApplicationState ApplicationStateType = "iota"
```
kingamarton, line 49: The string representation for New state should be "New". iota used here: https://github.com/apache/incubator-yunikorn-core/blob/f4ea4502a52e06e64e5abd0c47fedf2e08fe12e3/pkg/cache/application_state.go#L53-L61 is a  keyword to represent successive integer constants

pkg/apis/yunikorn.apache.org/v1alpha1/type.go
yangwwei: I think we can try to move the definition to the scheduler-interface, then let this reuse that.
However, I am not sure if this is supported by the code-gen, this needs further investigation.

pkg/apis/yunikorn.apache.org/update-codegen.sh
yangwwei: > About the license headers, code generator will read the boilerplate.go.txt which in k8s.io/code-generator/hack first. Base on my test, the --go-header-file won't work.

do you mean that specifying a customized file e.g /path/to/asf.header doesn't work with this command? Then how you add those license headers?

pkg/apis/yunikorn.apache.org/update-codegen.sh
yangwwei: We can turn this on for now. We can revisit this if we see some problems.

Makefile
```diff
@@ -165,6 +165,12 @@ push: image
 	docker push ${REGISTRY}/yunikorn:scheduler-${VERSION}
 	docker push ${REGISTRY}/yunikorn:admission-${VERSION}
 
+#Generate the CRD code with code-generator (release-1.14)
+.PHONY: code_gen
+code_gen:
+	@echo "Generating CRD code"
+	bash ${GOPATH}/src/$(REPO)/apis/yunikorn.apache.org/update-codegen.sh
```
yangwwei, line 172: We are using `go mod` to manage dependencies. That doesn't require ${GOPATH} to be set.
Can we use the relative path to this file, `pkg/apis/yunikorn.apache.org/update-codegen.sh` to run this script?
In this case, I think we will need to specify the output dir, pls see the other comment I added in the `update-codegen.sh`

pkg/apis/yunikorn.apache.org/update-codegen.sh
```diff
@@ -0,0 +1,29 @@
...
+#limitations under the License.
+#
+
+# generate the code with:
+# --output-base    because this script should also be able to run inside the vendor dir of
+#                  k8s.io/kubernetes. The output-base is needed for the generators to output into the vendor dir
+#                  instead of the $GOPATH directly. For normal projects this can be dropped.
+
+bash ${GOPATH}/src/k8s.io/code-generator/generate-groups.sh "all" \
+  github.com/apache/incubator-yunikorn-k8shim/pkg/client github.com/apache/incubator-yunikorn-k8shim/pkg/apis \
```
yangwwei, line 25: I think we need to add `--output-base` param for this.
Please try to set up the repo outside of $GOPATH.  As we are using `go mod` to manage dependencies, the repo can live outside the GOPATH.

Makefile
```diff
@@ -165,6 +165,16 @@ push: image
...
 
+#Generate the CRD code with code-generator (release-1.14)
+
+# If you want to re-run the code-generator to generate code,
+# Please make sure the directory structure must be the example.
+# ex: github.com/apache/incubator-yunikorn-k8shim
+.PHONY: code_gen
+code_gen:
+	@echo "Generating CRD code"
+	bash pkg/apis/yunikorn.apache.org/update-codegen.sh
```
yangwwei, line 176: NIT: do not use bash to run executable, some Linux box may not have bash installed, we do not need to depend on that. ./path/to/script.sh is enough. 

Another thing is, can we move this script out of pkg folder? under pkg, that is expecting all .go src files. For script, I think we can create a top level folder: scripts

pkg/apis/yunikorn.apache.org/custom-boilerplate.go.txt
```diff
@@ -0,0 +1,17 @@
+/*
```
yangwwei, line 1: can we also move this file to project_root/scripts/

pkg/apis/yunikorn.apache.org/update-codegen.sh
```diff
@@ -0,0 +1,29 @@
...
+#limitations under the License.
+#
+
+# If you want to re-run the code-generator to generate code,
+# Please make sure the directory structure must be the example.
+# ex: github.com/apache/incubator-yunikorn-k8shim
+
+SCRIPT_ROOT=$(dirname ${BASH_SOURCE})
+
+bash pkg/apis/yunikorn.apache.org/generate-groups.sh "all" \
```
yangwwei, line 25: NIT: remove bash

pkg/apis/yunikorn.apache.org/update-codegen.sh
```diff
@@ -0,0 +1,29 @@
...
+  github.com/apache/incubator-yunikorn-k8shim/pkg/client github.com/apache/incubator-yunikorn-k8shim/pkg/apis \
+  "yunikorn.apache.org:v1alpha1" \
+  --go-header-file "${SCRIPT_ROOT}"/custom-boilerplate.go.txt \
+  --output-base "$(dirname "${BASH_SOURCE[0]}")/../../../../../.."
```
yangwwei, line 29: suggest to move this script to project_root/scripts

Makefile
HuangTing-Yao: If we use`./` to run sh file, the `dirname` will fail, then the output-base setting will also fail.

Makefile
```diff
@@ -165,6 +165,16 @@ push: image
 	docker push ${REGISTRY}/yunikorn:scheduler-${VERSION}
 	docker push ${REGISTRY}/yunikorn:admission-${VERSION}
 
+#Generate the CRD code with code-generator (release-1.14)
+
+# If you want to re-run the code-generator to generate code,
+# Please make sure the directory structure must be the example.
+# ex: github.com/apache/incubator-yunikorn-k8shim
```
kingamarton, line 172: Is it possible to recognise where the repository is checked out instead of using a hardcoded directory structure? I tried to generate the code and in order to have it generated I have to change all my directory structure, what is a bit painfull. 
It will also fail, if GOPATH is not set. It would be good to set it in the script, or at least document it that we need it set. For me a simple " export GOPATH="$HOME/go"  solved the issue.

HuangTing-Yao: If we change the structure of directory, it may affect the import path which in the generated file.
About the GOPATH, if GOPATH is empty, it will export GOPATH in generate-groups.sh, which set in sub-shell, won't bring back to parent shell.The GOPATH in environment should be set by user.

kingamarton: Thank you for the explanation. It makes sense

PR #143, opened 2020-06-25
[YUNIKORN-256] remove error as a return from SendEvent

adamantal:


wilfred-s:
I have moved this to draft as it will a go.mod change after we have changed the interface and committed it

PR #146, opened 2020-06-29
[YUNIKORN-260] Handle node events in shim side

adamantal:


adamantal:
Tested with a dummy event emitted in core. Unfortunately the event did not appear in the output of the `kubectl describe` command, but `kubectl get events` listing showed it, so the event is in Kubernetes.

pkg/cache/context.go
```diff
@@ -606,6 +606,17 @@ func (ctx *Context) PublishEvents(eventRecords []*si.EventRecord) {
 						zap.String("taskID", taskID),
 						zap.String("event", record.String()))
 				}
+			case si.EventRecord_NODE:
+				nodeID := record.ObjectID
+				node, err := ctx.apiProvider.GetAPIs().NodeInformer.Lister().Get(nodeID)
```
yangwwei, line 611: this will cause the shim to query API server once the `PublishEvents` is called. this could be adding a big overhead to api-server. Instead of doing this, we should leverage the cached nodes in `pkg/cache/external/scheduler_cache.go`. we should be able to get the node object by `ctx.schedulerCache.GetNode(node).Node()`

adamantal: Agreed, fixed.

pkg/cache/context.go
```diff
@@ -606,6 +606,17 @@ func (ctx *Context) PublishEvents(eventRecords []*si.EventRecord) {
 						zap.String("taskID", taskID),
 						zap.String("event", record.String()))
 				}
+			case si.EventRecord_NODE:
+				nodeID := record.ObjectID
+				node := ctx.schedulerCache.GetNode(nodeID).Node()
```
yangwwei, line 611: we need to handle the case where `ctx.schedulerCache.GetNode` could be nil.
once this is fixed, I think this can be merged.

adamantal: Indeed. Fixed

adamantal:
Fixed @yangwwei 's review items, merged master and since I realized that we're not actually depending on #148, I also added unit tests. Could you check the patch again Weiwei, if the added unit tests are good?

PR #147, opened 2020-06-30
[YUNIKORN-263] fix lint issues introduced in YUNIKORN-170

HuangTing-Yao:
Add `--skip-dirs` to the lint run.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,9 +18,13 @@
 # options for analysis running
 run:
   issues-exit-code: 1
-  skip-dirs-use-default: true
   modules-download-mode: readonly
   deadline: 5m
+  skip-dirs:
+    - pkg/client/clientset
+    - pkg/client/informers
+    - pkg/client/listers
+    - pkg/apis/yunikorn.apache.org/v1alpha1/zz_generated.deepcopy.go
 
 # settings of specific linters
 linters-settings:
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -18,9 +18,13 @@
 # options for analysis running
 run:
   issues-exit-code: 1
-  skip-dirs-use-default: true
```
wilfred-s, line 21: This is independent of what we specify in the `skip-dirs`: it specifies that the check should not be run against dir like vendor etc. We should leave it, just to not depend on the defaults from the code.

.golangci.yml
```diff
@@ -18,9 +18,13 @@
...
   modules-download-mode: readonly
   deadline: 5m
+  skip-dirs:
+    - pkg/client/clientset
+    - pkg/client/informers
+    - pkg/client/listers
+    - pkg/apis/yunikorn.apache.org/v1alpha1/zz_generated.deepcopy.go
```
wilfred-s, line 27: I tried this added an extra new line to the file and it did not get skipped.
Should this not be under `skip-files` (https://golangci-lint.run/usage/configuration/)?

HuangTing-Yao:
@wilfred-s  I have already changed it as request, please help to review. Thanks.

PR #148, opened 2020-06-30
[YUNIKORN-253] Add unit tests for the shim side API

adamantal:


adamantal:
Ideally this should be merged in first before #146 so that I can add unit test coverage for the node types as well.

pkg/cache/context_test.go
```diff
@@ -460,3 +464,60 @@ func TestRemoveTask(t *testing.T) {
 	// now there is no task left
 	assert.Equal(t, len(app.GetNewTasks()), 0)
 }
+
+func TestPublishEvents(t *testing.T) {
```
yangwwei, line 468: we should have another UT to cover the case where app or task doesn't exist in the context.

pkg/cache/context_test.go
```diff
@@ -460,3 +464,60 @@ func TestRemoveTask(t *testing.T) {
...
+			log.Logger.Info(event)
+			if strings.Contains(event, reason) && strings.Contains(event, message) {
+				found = true
+			}
+		default:
+			if !found {
+				t.Fatal("the event is supposed to be published")
+			}
+		}
+	}
```
yangwwei, line 522: this could be hanging if something goes wrong. can you please add a timeout there?

adamantal: You're right, created a constructor with`util.waitFor`.

pkg/cache/context_test.go
adamantal: right. added a negative test

pkg/cache/context_test.go
```diff
@@ -20,20 +20,25 @@ package cache
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/utils"
```
yangwwei, line 27: pls format the import

pkg/cache/context_test.go
```diff
@@ -460,3 +465,100 @@ func TestRemoveTask(t *testing.T) {
...
+	// check that the event has been published
+	select {
+	case event := <-recorder.Events:
+		log.Logger.Info(event)
+		if strings.Contains(event, reason) && strings.Contains(event, message) {
+			t.Fatal("event should not be published if the pod does not exist")
+		}
+	default:
+		break
+	}
```
yangwwei, line 506: fix the potential issue that this may get stuck

adamantal:
Rebased and fixed the review items - sorry for the delay. Could you please review it @yangwwei ?

yangwwei:
hi @adamantal 

Changes look good. Can you fix the remaining lint error?
Next time you can run `make lint` locally before submitting the PR if you want to prevent such errors before creating the PR.

PR #149, opened 2020-07-01
[YUNIKORN-231] Add ability to update pod condition based on scheduling state

yangwwei:


pkg/cache/context.go
```diff
@@ -295,16 +295,25 @@ func (ctx *Context) triggerReloadConfig() {
 	}
 }
 
-func (ctx *Context) updatePodCondition(pod *v1.Pod, condition *v1.PodCondition) error {
-	log.Logger.Info("Updating pod condition",
+// update pod condition if the condition has changed, it returns a bool value to indicate
+// if the update is made or skipped, return an error if it is unable to make the change when it is needed to.
+func (ctx *Context) updatePodCondition(pod *v1.Pod, condition *v1.PodCondition) (bool, error) {
```
wilfred-s, line 300: This function should be merged into the caller, or the caller should be merged into this one. There are two layers of functions, two checks for errors and there is a 1:1 relation between `updatePodConditionFn` and `updatePodCondition`.
Too complex, does not make sense and uses a number of extra lines.

yangwwei: Make sense to me, I've removed this function and merged it to the caller.

pkg/cache/context.go
```diff
@@ -632,6 +620,59 @@ func (ctx *Context) PublishEvents(eventRecords []*si.EventRecord) {
 	}
 }
 
+// this function handles the pod scheduling failures with respect to the different causes,
+// and update the pod condition accordingly. the cluster autoscaler depends on the certain
+// pod condition in order to trigger auto-scaling.
+func (ctx *Context) HandleContainerStateUpdate(request *si.UpdateContainerSchedulingStateRequest) {
+	updatePodConditionFn := func(task *Task, podConditionReason, podConditionMsg, eventMessage string) {
```
wilfred-s, line 627: Can we move this big function out of the handle function into its own function?
Make a new function `ctx.updatePodCondition()` with the same parameters.

PR #150, opened 2020-07-07
[YUNIKORN-246] Travis CI: Build the docker image locally for e2e tests.

yangwwei:
Ensure e2e tests are running against the latest codebase.

### DIFF
--- a/scripts/run-e2e-tests.sh
+++ b/scripts/run-e2e-tests.sh
@@ -55,14 +55,19 @@ function install_cluster() {
   k8s_cluster_name=$1
   kind_node_image=$2
 
-  echo "step 1/5: installing helm-v3"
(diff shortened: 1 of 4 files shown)


scripts/run-e2e-tests.sh
```diff
@@ -0,0 +1,207 @@
...
+    exit 1
+    ;;
+esac
+done
+
+echo "action: ${action}"
+check_opt "${action}"
+echo "kind cluster name: ${cluster_name}"
+check_opt "${cluster_name}"
+echo "kind node image version ${cluster_version}"
```
ayubpathan, line 184: since "cluster_version" is not passed when this script is invoked, it will be always empty string!

wangdatan:
Thanks @yangwwei , patch looks good, the new structure looks much better than before. +1

PR #152, opened 2020-07-10
[YUNIKORN-274] app definition and example added

kingamarton:


deployments/yunikorn-application/application-definition.yaml
```diff
@@ -0,0 +1,71 @@
...
+    shortNames:
+      - app
+  validation:
+    openAPIV3Schema:
+      type: object
+      properties:
+        spec:
+          type: object
+          properties:
+            minMember:
```
yangwwei, line 53: 1. this is an optional field
2. it should be an integer

deployments/yunikorn-application/application-definition.yaml
```diff
@@ -0,0 +1,71 @@
...
+    plural: applications
+    # singular name to be used as an alias on the CLI and for display
+    singular: application
+    # kind is normally the CamelCased singular type. Your resource manifests use this.
+    kind: Application
+    # shortNames allow shorter string to match your resource on the CLI
+    shortNames:
+      - app
+  validation:
+    openAPIV3Schema:
```
yangwwei, line 47: we need to add more validation based on the schema
currently, it only defines a type, that's not enough

deployments/yunikorn-application/application-definition.yaml
```diff
@@ -0,0 +1,71 @@
...
+      type: object
+      properties:
+        spec:
+          type: object
+          properties:
+            minMember:
+              type: integer
+            queue:
```
yangwwei, line 55: 1. this is also an optional field (because the app-CRD is namespace scoped, user can create the CRD under a certain namespace and then leverage our placement rule to place the app to the corresponding queue)
2. it should be complying with the same regex like what defined here: https://github.com/apache/incubator-yunikorn-core/blob/06d2bd9a8b27211749f40caf2d568a0a3ea383d0/pkg/common/configs/configvalidator.go#L43
3. the example should give a full-qualified name, such as `root.default`.

deployments/yunikorn-application/application-definition.yaml
```diff
@@ -0,0 +1,71 @@
...
+              type: string
+            maxPendingSeconds:
+              type: integer
```
yangwwei, line 58: 1. this is also an optional field
2. it is an integer
3. min value 1 (if defined the app will be pending at most X seconds; if not defined, it can be pending forever)

deployments/yunikorn-application/sample-yunikorn-app.yaml
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: "yunikorn.apache.org/v1alpha1"
+kind: Application
+metadata:
+  name: example
```
yangwwei, line 21: we need to be careful here. will this field be used as our unique `applicationID`?
I guess this name can same across namespaces, correct? so most likely it cannot avoid dup.
And in most cases, users may not want to input appID, we should be able to generate one for them. How can we achieve that? Maybe use the object UID?
I think we need to think more about these problems before moving forward.

deployments/yunikorn-application/application-definition.yaml
kingamarton: by default it is optional. And it is integer, see line 54

deployments/yunikorn-application/sample-yunikorn-app.yaml
kingamarton: Adding a name is mandatory, and yes it will be unique only in the namespace. We can introduce a new field as ApplicationID and if it is not filled, we can generate it. I would export in some common place and use the same generator, used by the admission controller. What do you think?

deployments/yunikorn-application/application-definition.yaml
kingamarton: 1. By default it is optional
2, 3 added the requested changes

deployments/yunikorn-application/application-definition.yaml
kingamarton: 1. by default it is optional
2. it is an integer, see line 59
3. done

deployments/yunikorn-application/sample-yunikorn-app.yaml
kingamarton: An another idea is to build the applicationID from the namespace + name. When querying the application in case of a status update we need to provide the namespace and the name, but from the core side we will get only the applicationID. If we don't have the name then I would need to list all the apps and find the one with a specific applicationID, but if I can parse the name from the applicationID, it can be queried for a given namespace and name.

yangwwei: hi @kingamarton 

According to https://kubernetes.io/docs/concepts/overview/working-with-objects/names/. Object UID should be unique across the cluster. Maybe we can use that as the `applicationID`. That means the user only needs to specify a namespace-unique name for each app-CRD. Also, metadata has `generateName` field, so if the user doesn't specify any name, K8s will generate one with that as the prefix. What do you think about this?

yangwwei:
> According to https://kubernetes.io/docs/concepts/overview/working-with-objects/names/. Object UID should be unique across the cluster. Maybe we can use that as the applicationID. That means the user only needs to specify a namespace-unique name for each app-CRD. Also, metadata has generateName field, so if the user doesn't specify any name, K8s will generate one with that as the prefix. What do you think about this?

We can track these discussions later. It is not a blocker for this PR now. Let's get this merged. Thanks @kingamarton

PR #154, opened 2020-07-14
[YUNIKORN-279] Admission controller post-install needs to clean up key files to avoid leaking sensitive info.

yangwwei:
Post the admission-controller install, we should delete all the temp files, including the keys and certs, to avoid leaking sensitive info.


yangwwei:
> We should either fix it there or if we want to shorten the time that the files exist do it as soon as we have called kubectl create secret (i.e. move it to line 116). After that point we should never need the files anymore as the values come out of the secret.

Good catch, thanks. I have moved the `rm` command to the place after line 116. And I've removed the dup line (the incorrect) rm line.  Please take another look. Thanks!

PR #155, opened 2020-07-14
[YUNIKORN-268] When deleting a deployment the application is not deleted from the shim cache

HuangTing-Yao:
improve `RemoveApplication()` and `TestRemoveApplication()`

### DIFF
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -234,14 +234,10 @@ func (app *Application) getTasks(state string) []*Task {
 
 func (app *Application) getNonTerminatedTaskAlias() []string {
 	var nonTerminatedTaskAlias []string
-re:
 	for _, task := range app.taskMap {
-		for _, states := range events.States().Task.Terminated {
-			if states == task.GetTaskState() {
-				continue re
-			}
+		if !task.isTerminated() {
+			nonTerminatedTaskAlias = append(nonTerminatedTaskAlias, task.alias)
 		}
-		nonTerminatedTaskAlias = append(nonTerminatedTaskAlias, task.alias)
 	}
 	return nonTerminatedTaskAlias
 }
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -25,7 +25,7 @@ import (
 
 	"github.com/looplab/fsm"
 	"go.uber.org/zap"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 
 	"github.com/apache/incubator-yunikorn-core/pkg/api"
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/appmgmt/interfaces"
@@ -230,6 +230,24 @@ func (app *Application) getTasks(state string) []*Task {
 	})
 
 	return taskList
+}
+
+func (app *Application) getNonTerminatedTaskAlias() []string {
+	var nonTerminatedTaskAlias []string
+	terminatedStates := events.States().Task.Terminated
+	terminated := false
+	for _, task := range app.taskMap {
+		for _, states := range terminatedStates {
+			if states == task.GetTaskState() {
+				terminated = true
+			}
+		}
+		if !terminated {
+			nonTerminatedTaskAlias = append(nonTerminatedTaskAlias, task.alias)
+		}
+		terminated = false
+	}
+	return nonTerminatedTaskAlias
 }
 
 // only for testing
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -234,18 +234,14 @@ func (app *Application) getTasks(state string) []*Task {
 
 func (app *Application) getNonTerminatedTaskAlias() []string {
 	var nonTerminatedTaskAlias []string
-	terminatedStates := events.States().Task.Terminated
-	terminated := false
+re:
 	for _, task := range app.taskMap {
-		for _, states := range terminatedStates {
+		for _, states := range events.States().Task.Terminated {
 			if states == task.GetTaskState() {
-				terminated = true
+				continue re
 			}
 		}
-		if !terminated {
-			nonTerminatedTaskAlias = append(nonTerminatedTaskAlias, task.alias)
-		}
-		terminated = false
+		nonTerminatedTaskAlias = append(nonTerminatedTaskAlias, task.alias)
 	}
 	return nonTerminatedTaskAlias
 }
--- a/pkg/cache/application_test.go
+++ b/pkg/cache/application_test.go
@@ -222,3 +222,62 @@ func assertAppState(t *testing.T, app *Application, expectedState string, durati
 		}
 	}
 }
+
+func TestGetNonTerminatedTaskAlias(t *testing.T) {
+	context := initContextForTest()
+	app := NewApplication("app00001", "root.a", "testuser", map[string]string{}, newMockSchedulerAPI())
+	context.applications["app00001"] = app
+	// app doesn't have any task
+	res := app.getNonTerminatedTaskAlias()
+	assert.Equal(t, len(res), 0)
+
+	pod1 := &v1.Pod{
+		TypeMeta: apis.TypeMeta{
+			Kind:       "Pod",
+			APIVersion: "v1",
+		},
+		ObjectMeta: apis.ObjectMeta{
+			Name: "test-00001",
+			UID:  "UID-00001",
+		},
+	}
+	pod2 := &v1.Pod{
+		TypeMeta: apis.TypeMeta{
+			Kind:       "Pod",
+			APIVersion: "v1",
+		},
+		ObjectMeta: apis.ObjectMeta{
+			Name: "test-00002",
+			UID:  "UID-00002",
+		},
+	}
+	// set two task to non-terminated states
+	task1 := NewTask("task01", app, context, pod1)
+	app.taskMap["task01"] = task1
+	task1.sm.SetState(events.States().Task.Pending)
+	task2 := NewTask("task02", app, context, pod2)
+	app.taskMap["task02"] = task2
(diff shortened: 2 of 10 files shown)


HuangTing-Yao:
@yangwwei Please help to review. Thanks.

pkg/cache/context.go
```diff
@@ -499,10 +499,49 @@ func (ctx *Context) GetApplication(appID string) interfaces.ManagedApp {
 func (ctx *Context) RemoveApplication(appID string) error {
 	ctx.lock.Lock()
 	defer ctx.lock.Unlock()
-	if _, exist := ctx.applications[appID]; exist {
+	if app, exist := ctx.applications[appID]; exist {
```
yangwwei, line 502: This can be simplified. We can group a set of states into a group, e.g **Terminated**. That includes state: `Rejected`, `Killed`, `Failed`, `Completed`. If we found any task whose state is not in this group, then we fail with some error. I think you can do something similar to https://github.com/apache/incubator-yunikorn-k8shim/blob/1c46df413f290941616a4cb701e6ddd291851667/pkg/common/events/states.go#L75.

pkg/cache/context.go
```diff
@@ -499,10 +499,49 @@ func (ctx *Context) GetApplication(appID string) interfaces.ManagedApp {
...
+		var states = events.States().Task
+		for _, task := range app.taskMap {
+			switch taskState := task.GetTaskState(); taskState {
+			case states.New:
+				log.Logger.Warn("Can't remove application beacuse task states is New",
+					zap.String("taskID", task.GetTaskID()))
+				return fmt.Errorf("application %s still has task in non-terminated", appID)
```
yangwwei, line 509: Instead of failing once an error is seeing, can we change this to scan all the tasks? And then filter out the ones that not in the terminated state, then give a detailed error message. Something like: `application % cannot be removed because it has non-terminated tasks. Tasks: task1 [Killed], ... taskN [state]`.

pkg/cache/context.go
```diff
@@ -499,10 +499,49 @@ func (ctx *Context) GetApplication(appID string) interfaces.ManagedApp {
...
+			}
+		}
 		delete(ctx.applications, appID)
 		log.Logger.Info("app removed",
 			zap.String("appID", appID))
+		// send the update request to scheduler core
+		rr := common.CreateUpdateRequestForRemoveApplication(app.applicationID, app.partition)
+		if err := ctx.apiProvider.GetAPIs().SchedulerAPI.Update(&rr); err != nil {
+			log.Logger.Debug("failed to send remove application request to core", zap.Error(err))
```
yangwwei, line 543: Change `Debug` to `Error`.
And move line 537 after this. Do `delete(ctx.applications, appID)` after the `UpdateRequest` has been sent.

pkg/cache/context_test.go
```diff
@@ -119,42 +119,69 @@ func TestGetApplication(t *testing.T) {
...
+	assert.Assert(t, err != nil)
 
-	// remove application
+	app := context.GetApplication("app00001")
+	assert.Assert(t, app != nil)
+
+	// remove application 2 which have terminated task
 	// this should be successful
-	err := context.RemoveApplication("app00001")
+	err = context.RemoveApplication("app00002")
```
yangwwei, line 172: after the app is removed. make sure the app is removed from the cache.
can you pls add a check for that?

pkg/cache/context_test.go
```diff
@@ -119,42 +119,69 @@ func TestGetApplication(t *testing.T) {
...
+	app = context.GetApplication("app00002")
 	assert.Assert(t, app == nil)
 
-	// try remove again
-	// this should fail
-	err = context.RemoveApplication("app00001")
+	//try remove again
+	//this should fail
+	err = context.RemoveApplication("app00002")
 	assert.Assert(t, err != nil)
```
yangwwei, line 181: once you have improved the error message, pls add the verification for that too.
make sure the error message contains enough info we want.

pkg/common/si_helper.go
```diff
@@ -162,3 +162,17 @@ func CreateUpdateRequestForDeleteNode(node Node) si.UpdateRequest {
 	}
 	return request
 }
+
+func CreateUpdateRequestForRemoveApplication(appID, partition string) si.UpdateRequest {
```
yangwwei, line 166: We need UT for this function too

pkg/cache/context_test.go
HuangTing-Yao: I think I had already check in Line 175. 
Do I misunderstanding anything?

yangwwei: Sorry, this is already handled. You can ignore this.

pkg/cache/context.go
```diff
@@ -499,10 +499,31 @@ func (ctx *Context) GetApplication(appID string) interfaces.ManagedApp {
...
+		var terminatedStates = events.States().Task.Terminated
+		terminatedNum := 0
+		taskmessage := "Task: "
+		for _, task := range app.taskMap {
+			for _, states := range terminatedStates {
+				if states == task.GetTaskState() {
+					terminatedNum++
+				}
+			}
+			taskmessage = taskmessage + task.alias + ", "
```
yangwwei, line 512: this will leave a "," in the end of the string. This can be fixed by using https://golang.org/pkg/strings/#Join
another thing is we can simplify the code in `context.go` by introducing a function in `application.go`, like the following

```
// returns a array of task alias that in non-terminated state
nonTerminatedTaskAlias:= app.getNonTerminatedTaskAlias()
if len(nonTerminatedTasks) > 0 {
  return fmt.Errorf("failed to remove application %s because it still has tasks in non-terminated state, tasks:  %s", appID, strings.Join(nonTerminatedTaskAlias, ","))
}
 
```

yangwwei: if this makes sense to you, pls add UT for `app.getNonTerminatedTaskAlias()` too. So we can cover cases like:
- app doesn't have any task
- app has some tasks, all in the terminated state
- app has some tasks, all in non-terminated state
- app has some tasks that in both terminated or non-terminated state

pkg/cache/application.go
```diff
@@ -232,6 +232,24 @@ func (app *Application) getTasks(state string) []*Task {
 	return taskList
 }
 
+func (app *Application) getNonTerminatedTaskAlias() []string {
+	var nonTerminatedTaskAlias []string
```
yangwwei, line 236: looks like this can be simplified like the following:

```
var nonTerminatedTaskAlias []string
	for _, task := range app.taskMap {
		for _, states := range events.States().Task.Terminated {
			if states == task.GetTaskState() {
				continue
			}
		}
             nonTerminatedTaskAlias = append(nonTerminatedTaskAlias, task.alias)
	}
	return nonTerminatedTaskAlias
```

HuangTing-Yao: Maybe we can add continue label `re`  ahead of the for of  app.taskMap and use `continue re ` to avoid append the terminated task.

pkg/cache/application_test.go
```diff
@@ -222,3 +222,62 @@ func assertAppState(t *testing.T, app *Application, expectedState string, durati
 		}
 	}
 }
+
+func TestGetNonTerminatedTaskAlias(t *testing.T) {
+	context := initContextForTest()
+	app := NewApplication("app00001", "root.a", "testuser", map[string]string{}, newMockSchedulerAPI())
```
kingamarton, line 228: Please extract the appId to a variable

pkg/cache/application_test.go
```diff
@@ -222,3 +222,62 @@ func assertAppState(t *testing.T, app *Application, expectedState string, durati
...
+			APIVersion: "v1",
+		},
+		ObjectMeta: apis.ObjectMeta{
+			Name: "test-00002",
+			UID:  "UID-00002",
+		},
+	}
+	// set two task to non-terminated states
+	task1 := NewTask("task01", app, context, pod1)
+	app.taskMap["task01"] = task1
```
kingamarton, line 256: Please extract the taskId into a variable

pkg/cache/application_test.go
```diff
@@ -222,3 +222,62 @@ func assertAppState(t *testing.T, app *Application, expectedState string, durati
...
+	task1.sm.SetState(events.States().Task.Pending)
+	task2 := NewTask("task02", app, context, pod2)
```
kingamarton, line 258: Please extract the taskId into a variable

pkg/cache/context_test.go
```diff
@@ -119,42 +119,71 @@ func TestGetApplication(t *testing.T) {
...
-	// add 2 applications
+	// add 3 applications
 	context := initContextForTest()
-	context.AddApplication(&interfaces.AddApplicationRequest{
-		Metadata: interfaces.ApplicationMetadata{
-			ApplicationID: "app00001",
-			QueueName:     "root.a",
-			User:          "test-user",
-			Tags:          nil,
+	app1 := NewApplication("app00001", "root.a", "testuser", map[string]string{}, newMockSchedulerAPI())
```
kingamarton, line 124: extract this appIDs into variable since they are used in multiple places

pkg/shim/scheduler_test.go
```diff
@@ -133,6 +133,9 @@ partitions:
 	cluster.waitAndAssertApplicationState(t, "app0001", events.States().Application.Failed)
```
kingamarton, line 133: please extract the appId

HuangTing-Yao:
@kingamarton this PR only handle the update request should send to core.
Doesn't check the core part.

yangwwei:
We can do this incrementally. We don't have to check the core side changes in this PR, it's fine to do that later when we hook things up together. The core side removeApp has UT covered, that should work. The changes in this PR looks good, I'll merge shortly.

PR #156, opened 2020-07-14
[YUNIKORN-281] update core dependency

wilfred-s:
changes to dependency for branch-0.9


wilfred-s:
merging the change no review needed just dependency update

PR #163, opened 2020-07-22
[YUNIKORN-177] SchedulerName doesn't need to be configurable

adamantal:


yangwwei:
hi @adamantal could you pls check why the e2e testing is failing?

yangwwei:
It might because this line:
https://github.com/apache/incubator-yunikorn-k8shim/blob/590438fde5bd451744bc59b9074b91014d2938e3/.travis.yml#L58
it always installs the latest version of `ginkgo `. the CLI syntax might be changing. cc @ayubpathan for this issue.

adamantal:
Since I only substituted a string with a constant I don't think this is related to the patch itself. Should we file a jira for this e2e test failure?

adamantal:
Removed the changes from the e2e tests. I don't know what has caused this, seemed unrelated to me, but I'd rather not touch that part of the code.

yangwwei:
Weird, it is still failing..

kingamarton:
I think the test failure is not related. It failed with the following error: 'error: timed out waiting for the condition on deployments/yunikorn-scheduler'. However I tried to deploy it in the same way as the e2e test is doing, and it got deployed successfully. I would suggest to create a Jira to review the e2e test.

yangwwei:
hi @kingamarton @adamantal 

If you look at the docker file we used to build the scheduler, you'll see: https://github.com/apache/incubator-yunikorn-k8shim/blob/e11a7e068da3e8b024ceb86abe6c9a945308aca7/deployments/image/configmap/Dockerfile#L54. The `schedulerName` is part of the startup option. The PR here removes this option from the executable,  which causes the scheduler could not be started. I think the docker image also needs to be updated, to remove that option.

adamantal:
Yes, it is much clear now. Thanks for the comments!

I checked and actually the `make run` command was also affected (could not start the scheduler with the patch). Removed two occurences of he `-name` flag, and hopefully I caught each of them.

yangwwei:
I just did a minor update that removes SCHEDULER_NAME env var in the docker file, rest looks good to me.  Merging now. Thanks @adamantal .

adamantal:
Thanks @yangwwei !

PR #168, opened 2020-07-28
YUNIKORN-318: Adding support to read image pull secrets for admission controller

ayubpathan:


ayubpathan:
@yangwwei  Please review.

deployments/admission-controllers/scheduler/templates/server.yaml.template
```diff
@@ -15,6 +15,7 @@ spec:
       labels:
         app: yunikorn
     spec:
+      imagePullSecrets:
```
yangwwei, line 18: hi @ayubpathan 

When `imagePullSecrets` is empty, this will leave a field like the following:

```
spec:
      imagePullSecrets:
```

will this break the helm install? As it might expecting something for this field. Could you please verify this?

ayubpathan: The deployment wont break. The empty spec vars are ignored!

PR #170, opened 2020-07-29
[YUNIKORN-204] Basic controller logic 

kingamarton:
This is a WIP PR. There are a couple of missing thins to to: adding tests, fix the TODO's, but I am creating it because some feedback it would be very helpful regarding the main idea. (especially because there are some things what smells a bit hacky for me, but I had no better idea).
@wilfred-s, @yangwwei, @HuangTing-Yao can you please check it when you will have some time?


pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+// this implements AppManagementService interface
+func (appMgr *AppManager) Start() error {
+	appMgr.controller = NewController(appMgr)
+	appMgr.controller.Run()
+	return nil
+}
+
+// this implements AppManagementService interface
+func (appMgr *AppManager) Stop() {
+	// noop
```
yangwwei, line 79: this is not a noop. it needs to stop the controller

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+import (
+	"os"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+
+	"go.uber.org/zap"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
```
yangwwei, line 30: do we really need to depend on this package?

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+	}
+}
+
+type Controller struct {
+	kubeClient       kubernetes.Interface
+	extensionsClient apiextensionsclient.Interface
+	appClient        applicationclient.Interface
+	informer         cache.SharedIndexInformer
+	lister           applister.ApplicationLister
```
yangwwei, line 118: is it possible to move most of these things to `apifactory.go`? 
the `APIProvider` is supposed to provide a convenient way to access client libraries.

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	informerFactory := appinformers.NewSharedInformerFactory(appClient, time.Minute*1)
+	informer := informerFactory.Apache().V1alpha1().Applications()
+	informer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    appMgr.addApp,
+		UpdateFunc: appMgr.updateApp,
+		DeleteFunc: appMgr.deleteApp,
+	})
+	informerFactory.Start(wait.NeverStop)
+
+	utilruntime.Must(appv1.AddToScheme(appscheme.Scheme))
```
yangwwei, line 141: what is this for?

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+		extensionsClient: apiextensionsClient,
+		appClient:        appClient,
+		informer:         informer.Informer(),
+		lister:           informer.Lister(),
+	}
+}
+
+func (c *Controller) Run() {
+	//TODO: handle the case when the crd is not registered: register it programatically
+	defer utilruntime.HandleCrash()
```
yangwwei, line 153: I am not quite sure what does this mean.
why do we need to handle the case the CRD is not registered? If the CRD is not registered, I think we should not even start the scheduler. And also, what does the `HandleCrash()` do here?

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	_, err := appMgr.controller.appClient.ApacheV1alpha1().Applications(appCRD.Namespace).UpdateStatus(copy)
+	if err != nil {
+		log.Logger.Error("Failed to update application CRD",
+			zap.String("AppId", copy.Name))
+		return
+	}
+}
+
+func (appMgr *AppManager) getAppMetadata(app *appv1.Application) (interfaces.ApplicationMetadata, bool) {
+	appID := app.Name
```
yangwwei, line 254: Like I commented here: https://github.com/apache/incubator-yunikorn-k8shim/pull/152#discussion_r453930011.
> According to https://kubernetes.io/docs/concepts/overview/working-with-objects/names/. Object UID should be unique across the cluster. Maybe we can use that as the applicationID. That means the user only needs to specify a namespace-unique name for each app-CRD. Also, metadata has generateName field, so if the user doesn't specify any name, K8s will generate one with that as the prefix. 

I think we should let `appID` = spec.UID, not the name, to ensure it is unique.
We can put the appName in the tag, in the format of `namespace/appName`

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+		return appv1.WaitingState
+	case corecache.Rejected.String():
+		return appv1.RejectedState
+	case corecache.Completed.String():
+		return appv1.CompletedState
+	case corecache.Killed.String():
+		return appv1.KilledState
+	default:
+		return "Undefined"
+	}
```
yangwwei, line 294: I think this is one of the main things we need to fix.
We need to make shim side app state in sync with the core side states, pls take a look at the core side status: http://yunikorn.apache.org/docs/next/design/scheduler_object_states#application-state.
I would even suggest creating a separate JIRA to get this fixed first.
Ideally, all the state transition are done in the core side, and core leverages UpdateApplication API to send the state updates to the shim, and that triggers the transition of shim side app-state-machine.

pkg/controller/application/app_controller.go
kingamarton: this utilruntime.Must call is handling the errors. Actually it will panic if the scheme registration is failing. I've seen this pattern used in other places, when starting the controller, so I followed this one as well.

pkg/controller/application/app_controller.go
kingamarton: Right now we have two places where we can enable the CRD: the scheduler conf and the helm chart. In my opinion we should reduce it to only one place, the scheduler conf. If we will kepp this two places, I think it will be a frequent user error that they will enable it in one place and forget to enable it in the helm chart for example, or will forget to register it manually. I think in this case it will be useful to register it from the code.

pkg/controller/application/app_controller.go
kingamarton: I completely agree with you that we should keep the states in synch. For me this one seems a bigger refactoring task. And also we have different statuses for the shim side and different ones for the core side:
https://github.com/apache/incubator-yunikorn-k8shim/blob/master/pkg/cache/application.go#L71-L101
https://github.com/apache/incubator-yunikorn-core/blob/master/pkg/cache/application_state.go#L70-L100
If we want refactor it now, after then it will be easier to handle the AppCRD status as well, than it is now.

yangwwei: Based on the offline discussion, we will not do the refactoring for now given the complexity of this problem.

pkg/appmgmt/sparkoperator/spark.go
```diff
@@ -104,6 +105,12 @@ func (os *Manager) Stop() {
 	os.stopCh <- struct{}{}
 }
 
+func (os *Manager) HandleCallbackEvents() func(obj interface{}) {
+	return func(obj interface{}) {
+		//noop
+	}
+}
+
```
yangwwei, line 113: this can be removed once the handler function is moved to `AppManagementService`.

pkg/cache/application_events.go
```diff
@@ -47,6 +47,35 @@ func (st SimpleApplicationEvent) GetApplicationID() string {
 	return st.applicationID
 }
 
+// ------------------------
+// ApplicationStatusChangeEvent updates the status in the application CRD
+// ------------------------
+type ApplicationStatusChangeEvent struct {
+	applicationID string
+	event         events.ApplicationEventType
+	State         string
```
yangwwei, line 56: for a state change, it needs to have 2 fields: `source` and `destination` state.
if we do not have this in the scheduler-interface, I think we need to add that too. Sorry for missing this earlier.

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	"go.uber.org/zap"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clientcmd"
+
+	corecache "github.com/apache/incubator-yunikorn-core/pkg/cache"
```
yangwwei, line 36: please do not depend on this package

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+		ApplicationID: appID,
+		QueueName:     app.Spec.Queue,
+		User:          "default",
+		Tags:          tags,
+	}, true
+}
+
+func convertShimAppStateToAppCRDState(status string) appv1.ApplicationStateType {
+	switch status {
+	case corecache.New.String():
```
yangwwei, line 276: we should not depend on the core side package directly,
based on current limitations, I think we will just do the parse with strings...

pkg/shim/scheduler.go
```diff
@@ -110,10 +111,21 @@ func newShimSchedulerInternal(ctx *cache.Context, apiFactory client.APIProvider,
 	dispatcher.RegisterEventHandler(dispatcher.EventTypeTask, ctx.TaskEventHandler())
 	dispatcher.RegisterEventHandler(dispatcher.EventTypeNode, ctx.SchedulerNodeEventHandler())
 	dispatcher.RegisterEventHandler(dispatcher.EventTypeScheduler, ss.SchedulerEventHandler())
+	ss.registerAppManagerHandler(constants.AppManagerHandlerName)
```
yangwwei, line 114: Like the previous comment, please replace this with 

```
dispatcher.RegisterEventHandler(dispatcher. EventTypeAppStateUpdate, am.ApplicationStateUpdateEventHandler())
```

pkg/shim/scheduler.go
```diff
@@ -110,10 +111,21 @@ func newShimSchedulerInternal(ctx *cache.Context, apiFactory client.APIProvider,
...
+func (ss *KubernetesShim) registerAppManagerHandler (appMgrName string) {
+	ss.context.AddSchedulingEventHandlers()
+	appMgr := ss.appManager.GetManagerByName(appMgrName)
+	if appMgr != nil {
+		dispatcher.RegisterEventHandler(dispatcher.EventTypeAppStatus, appMgr.HandleCallbackEvents())
+	} else {
+		log.Logger.Debug("App manager not registered",
+		zap.String("app manager name", appMgrName))
+	}
+}
```
yangwwei, line 128: this can be removed

pkg/controller/application/app_controller.go
yangwwei: Could you please fix this based on the comment here: https://issues.apache.org/jira/browse/YUNIKORN-340?focusedCommentId=17169061&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17169061

pkg/controller/application/app_controller.go
yangwwei: OK, if this is the common pattern, then that is fine.

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	appClient := applicationclient.NewForConfigOrDie(config)
+
+	informerFactory := appinformers.NewSharedInformerFactory(appClient, time.Minute*1)
+	informer := informerFactory.Apache().V1alpha1().Applications()
+	informer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    appMgr.addApp,
+		UpdateFunc: appMgr.updateApp,
+		DeleteFunc: appMgr.deleteApp,
+	})
+	informerFactory.Start(wait.NeverStop)
```
yangwwei, line 139: you have a `Run()` function, we should move the start of the informerFactory stuff to there.

pkg/controller/application/app_controller.go
yangwwei: Got it, thx

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	log.Logger.Info("Waiting cache to be synced.")
+	// Handle timeout for syncing.
+	timeout := time.NewTimer(time.Second * 30)
+	timeoutCh := make(chan struct{})
+	go func() {
+		<-timeout.C
+		timeoutCh <- struct{}{}
+	}()
+	if ok := cache.WaitForCacheSync(timeoutCh, c.informer.HasSynced); !ok {
+		log.Logger.Error("Timeout expired during waiting for caches to sync. Skip starting the application controller")
```
yangwwei, line 164: this log is wrong, it is not skipping the controller start, everything has been started. this only means the controller is running without the up-to-date sync'd cache.

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	// No need to update if ResourceVersion is not changed
+	if newApp.ResourceVersion == oldApp.ResourceVersion {
+		log.Logger.Info("No need to update because app object is not modified")
+		return
+	}
+
+	// handle spec change
+	if !equality.Semantic.DeepEqual(oldApp.Spec, newApp.Spec) {
+		//TODO: handle update, discuss what should be updated since only the queue is the common property yet
+	}
```
yangwwei, line 191: I'd prefer to leave it empty, with a comment

```
// not implement today
```

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+/*
+Remove the application from the scheduler as well
+*/
+func (appMgr *AppManager) deleteApp(obj interface{}) {
+	app, ok := obj.(*appv1.Application)
+	if !ok {
+		log.Logger.Error("obj is not an Application")
```
yangwwei, line 200: forgot a return after the error logging here?

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	appID := app.Name
+	err := appMgr.amProtocol.RemoveApplication(appID)
+	if err != nil {
+		log.Logger.Error("Application removal failed",
+			zap.String("appID", appID),
+			zap.Error(err))
```
yangwwei, line 207: I think a `return` needs to be added after the error log

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	//TODO: delete related pods and deployments
+}
+
+/*
+Add application to scheduler
+*/
+func (appMgr *AppManager) addApp(obj interface{}) {
+	appCRD, ok := obj.(*appv1.Application)
+	if !ok {
+		log.Logger.Error("obj is not an Application")
```
yangwwei, line 220: I think a return needs to be added after the error log

pkg/controller/application/app_controller.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	if appMeta, ok := appMgr.getAppMetadata(appCRD); ok {
+		app := appMgr.amProtocol.GetApplication(appMeta.ApplicationID)
+		if app == nil {
+			appMgr.amProtocol.AddApplication(&interfaces.AddApplicationRequest{
+				Metadata: appMeta,
+				//TODO: debug recovery
+				//TODO: try to define the owner reference for the submitted pods to this CRD and try of we delete the
+				//CRD the pods will be deleted as well
```
yangwwei, line 229: We need a separate JIRA to track how to implement the recovery, pls file one

pkg/dispatcher/dispatcher.go
```diff
@@ -179,7 +180,11 @@ func Start() {
 			case event := <-dispatcher.eventChan:
 				switch v := event.(type) {
 				case events.ApplicationEvent:
-					getEventHandler(EventTypeApp)(v)
+					if v.GetEvent() == events.AppStateChange {
```
yangwwei, line 183: I feel this could be misleading.
Can we define another interface for the app state change? In `pkg/common/events/events.go`, create another interface specifically for handling the app-state-updates from the core to the shim.

pkg/controller/application/app_controller.go
kingamarton: Can you please elaborate on what exactly do you expect about stopping the controller? Closing the channel, or something else?

pkg/controller/application/app_controller.go
kingamarton: I will remove kubeclient and extensionsClient from here, since I need them only for initialising the other variables. The other ones are all application related and I would keep them here, since we need them to be created only if the CRD feature is turned on.

pkg/controller/application/app_controller.go
kingamarton: Since using UID is more complicated, I will use namespace-name as applicationID

yangwwei: Make sense to me, thanks

pkg/controller/application/app_controller.go
kingamarton: It is not a common pattern, it is just easy to forget to enable it in both places. I created a different Jira for it: YUNIKORN-346

pkg/cache/application_events.go
kingamarton: From the application point of view we don't need the source state, because when this event is generated the transition is already done, so we just need to update it in the CRD. The destination field will be unused. I suggest to extend it with the destination only if we will need it.

pkg/controller/application/app_controller.go
kingamarton: If we will not implement the update, I would rather delete it than leaving it empty

pkg/controller/application/app_controller.go
kingamarton: Filed YUNIKORN-347

pkg/appmgmt/interfaces/appmgr.go
```diff
@@ -42,4 +42,6 @@ type AppManager interface {
 	// if there is some go routines running in start, properly stop them while
 	// the stop() function is called.
 	Stop()
+
+	HandleCallbackEvents() func(obj interface{})
```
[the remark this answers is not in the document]
kingamarton: done

pkg/controller/application/app_controller.go
kingamarton: removed

pkg/controller/application/app_controller.go
kingamarton: removed

pkg/controller/application/app_controller.go
kingamarton: I am thinking about removing this line, because it seems to be enough to start the informers

pkg/dispatcher/dispatcher.go
kingamarton: Created a new interface: events.ApplicationStatusEvent

kingamarton:
@yangwwei I have addressed your review comments. I still have to add the unit tests and try to set the ownerReferrence for the pods belonging to an application.
@wilfred-s  can you please take a look to this changes as well?

pkg/client/clients.go
```diff
@@ -67,7 +71,8 @@ func (c *Clients) WaitForSync(interval time.Duration, timeout time.Duration) err
 			c.PVInformer.Informer().HasSynced() &&
 			c.StorageInformer.Informer().HasSynced() &&
 			c.ConfigMapInformer.Informer().HasSynced() &&
-			c.NamespaceInformer.Informer().HasSynced()
+			c.NamespaceInformer.Informer().HasSynced() &&
+			c.ApplicationInformer.Informer().HasSynced()
```
yangwwei, line 75: Seems this could be causing a NPE when app-CRD is not enabled.
Could you please add a nil check here?

kingamarton: Good catch. This is valid. I will fix it

yangwwei:
hi @kingamarton changes look good to me. But could you please check the travis job failure? Thanks.

kingamarton:
@yangwwei, the actual failure is because I need to change the go.mod file to point to the latest interface and also to the latest core. For updating it we need to merge the core side changes first otherwise the shim and the core will point to different interface versions.

yangwwei:
hi @kingamarton  can you at least fix the `codecov/project` error? Thanks

yangwwei:
hi, @kingamarton thanks for getting this done. Now it looks good.
The latest job shows only the codecov/project is failing, but I don't think that is caused by your patch. We will need to improve the code coverage in the shim repo to get those places, such as `application_events.go`. We can do that in follow up tasks, I have created https://issues.apache.org/jira/browse/YUNIKORN-383 for this.

PR #174, opened 2020-08-07
[YUNIKORN-349] Remove installCRD option during helm install for e2e test

HuangTing-Yao:
remove installCRD option during helm install.


HuangTing-Yao:
@yangwwei please review.

yangwwei:
Change looks good, but I guess the e2e test fails because we haven't updated the helm chart package in the helm hub repo.
Let's get that done first and then I will trigger the job again. Thanks @HuangTing-Yao

PR #178, opened 2020-08-10
[YUNIKORN-354] Update YK base docker images to UBI

adamantal:
Updated the base Docker image for the container

Questions:
- [ ] where should be changed the helm charts?
- [ ] should the base image for the admission controller be changed?
- [ ] add license for UBI image if needed

Tested:
locally using `make clean run` on  docker-desktop

### DIFF
--- a/deployments/image/configmap/Dockerfile
+++ b/deployments/image/configmap/Dockerfile
@@ -15,7 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.12.13-alpine3.10
+FROM registry.redhat.io/ubi8/go-toolset:1.12.8-45
 
 # admission controller bundles
 RUN apk add curl


kingamarton:
@adamantal The helm charts are located in the release repository: https://github.com/apache/incubator-yunikorn-release/tree/master/helm-charts/yunikorn

yangwwei:
Looks like the e2e test is failing due to:
> Step 1/34 : FROM registry.redhat.io/ubi8/go-toolset:1.12.8-45
Get https://registry.redhat.io/v2/ubi8/go-toolset/manifests/1.12.8-45: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials. Further instructions can be found here: https://access.redhat.com/RegistryAuthentication

adamantal:
Since the RedHat Docker registry requires authentication which may conflict with the ASF policies we decided to abandon this patch and try a downstream solution.

wilfred-s:
UBIs from RedHat are set to be moved into authenticated access only registries. That will require a RH account and thus acceptance of the terms that come with it. We would still need further entitlements to be able to add packages etc in the base image. 
I don't think that is a good idea for an Apache release.

I am a -1 on this change.

PR #179, opened 2020-08-12
[YUNIKORN-24] Fix the UT compatibility issue with GO 1.13.x

yangwwei:


wilfred-s:
I am not to happy about adding `testing.Init()`
The other option that was discussed and used in a number of places is to move the `flag.Parse()` call into `main()`

Why did you choose to keep the parse in the init of the config and not move it into main?

yangwwei:
> I am not to happy about adding `testing.Init()`
> The other option that was discussed and used in a number of places is to move the `flag.Parse()` call into `main()`
> 
> Why did you choose to keep the parse in the init of the config and not move it into main?

hi @wilfred-s , thanks for the review. Actually, that was an easier solution (a workaround). 
Moving the `flag.Parse()` call into `main()` is not enough. I just uploaded a complete solution, that removes all `init()` calls in the `conf`, `dispatcher`, and `log` package. Replacing them with `sync.once` through a getter. Lots of files got affected, because the change of `log.Logger` to `log.Logger()`. But I think this is necessary if we want to have something better than the previous workaround.
What do you think?

PR #182, opened 2020-08-18
[YUNIKORN-366] Add rest API to update queue configuration

kingamarton:


pkg/cache/context.go
```diff
@@ -765,3 +767,62 @@ func (ctx *Context) SchedulerNodeEventHandler() func(obj interface{}) {
...
+			ykconf = c
+			foundConf = true
+			break
+		}
+	}
+	if !foundConf {
+		return nil, fmt.Errorf("configmap with name %s not found", common.DefaultConfigMapName)
+	}
+	return ykconf, nil
+}
```
yangwwei, line 799: this can be simplified to:

```
for _, c := range configMaps {
  if c.Name == common.DefaultConfigMapName {
    return c
  }
}
return nil, fmt.Errorf("configmap with name %s not found", common.DefaultConfigMapName)
```
because range handles the case when the slice is nil

pkg/cache/context.go
```diff
@@ -765,3 +767,62 @@ func (ctx *Context) SchedulerNodeEventHandler() func(obj interface{}) {
...
+func createConfigMapSelector() (labels.Selector, error) {
+	slt := labels.NewSelector()
+	appVals := [] string {"yunikorn"}
+	req, err := labels.NewRequirement(common.LabelApp, selection.Equals, appVals)
+	if err != nil {
+		return  nil, err
+	}
+	slt = slt.Add(*req)
+	return slt, nil
+}
```
yangwwei, line 780: Can you pls check if this can be simplified by

```
labels.SelectorFromSet()
```

if that can work, seems like we do not need a separate function for this.

pkg/cache/context.go
```diff
@@ -765,3 +767,62 @@ func (ctx *Context) SchedulerNodeEventHandler() func(obj interface{}) {
...
+	configMaps, err := ctx.apiProvider.GetAPIs().ConfigMapInformer.Lister().List(slt)
+	if err != nil {
+		return "", err
+	}
+	ykconf, err := findYKConfigMap(configMaps)
+	if err != nil {
+		return "", err
+	}
+
+	newConfData := map[string] string {"queues.yaml":strings.ReplaceAll(data, "\r\n", "\n")}
```
yangwwei, line 818: not sure why the replaceAll is needed here. 
Why we have `\r\n` in the fist place?

pkg/cache/context.go
```diff
@@ -765,3 +767,62 @@ func (ctx *Context) SchedulerNodeEventHandler() func(obj interface{}) {
...
+	newConf := ykconf.DeepCopy()
+	oldConfData := ykconf.Data["queues.yaml"]
+	newConf.Data = newConfData
+	_, err = ctx.apiProvider.GetAPIs().KubeClient.GetClientSet().CoreV1().ConfigMaps(ykconf.Namespace).Update(newConf)
+	if err != nil {
+		return "", err
+	}
+	log.Logger().Debug("ConfigMap updated successfully")
```
yangwwei, line 826: this should be the `INFO` level

pkg/cache/context.go
```diff
@@ -21,6 +21,8 @@ package cache
 import (
 	"encoding/json"
 	"fmt"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"strings"
```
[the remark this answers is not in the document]
kingamarton: I gofmt-ed the file and also goimport-ed.

pkg/cache/context.go
kingamarton: When I did some testing using Advanced REST client, without this replace, the saved configuration in the Kubernetes dashboard wasn't parsed correctly.  Instead of breaking the lines, \r\n was showed. By replacing "\r\n'' with "\n", the new configuration will be shown in the similar way, as before this feature.

pkg/cache/context.go
kingamarton: Changed to INFO level

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -179,8 +182,37 @@ func updateLabels(namespace string, pod *v1.Pod, patch []patchOperation) []patch
...
+		return false
+	}
+	if allowed {
+		return true
+	}
+	if strings.Contains(userInfo, "yunikorn-admin") {
+		return true
+	}
+	return false
+}
```
yangwwei, line 200: Need to add UT for this.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -179,8 +182,37 @@ func updateLabels(namespace string, pod *v1.Pod, patch []patchOperation) []patch
...
+		log.Logger().Error("Failed to parse ENABLE_CONFIG_HOT_REFRESH value",
+			zap.String("ENABLE_CONFIG_HOT_REFRESH", hotRefreshEnabled))
+		return false
+	}
+	if allowed {
+		return true
+	}
+	if strings.Contains(userInfo, "yunikorn-admin") {
+		return true
+	}
```
yangwwei, line 198: this can be combined to 

```
if allowed || strings.Contains(userInfo, "yunikorn-admin") {
  return true
}
```

pkg/cache/context.go
kingamarton: It works, I changed it

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
kingamarton: added

pkg/conf/schedulerconf.go
```diff
@@ -146,6 +147,9 @@ func initConfigs() {
 		"log encoding, json or console.")
 	logFile := flag.String("logFile", "",
 		"absolute log file path")
+	enableConfigHotRefresh := flag.Bool("enableConfigHotRefresh", false, "Flag for enabling configuration reload " +
```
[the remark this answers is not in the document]
kingamarton: done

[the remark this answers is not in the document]
kingamarton: done

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
kingamarton: done

kingamarton:
The e2e tests failed because by default I disabled the config hot-refresh and made the API available

kingamarton:
@yangwwei, I don't understand why the coverage report is changed between the last 2 commits, with the previous commit it passed and now it failed. In the last commit I changed only the hot-refresh flag. Also the coverage reports some misses for the scheduler_mock.

yangwwei:
Looks good now. I'll merge this shortly.

PR #183, opened 2020-08-25
[YUNIKORN-382]Update kubectl in dockerfile to comply with the support policy

HuangTing-Yao:
Update the kubectl in Dockerfile from 1.14.7 to 1.16.13.

### DIFF
--- a/deployments/image/configmap/Dockerfile
+++ b/deployments/image/configmap/Dockerfile
@@ -21,7 +21,7 @@ FROM golang:1.12.13-alpine3.10
 RUN apk add curl
 RUN apk add jq
 RUN apk add --update openssl
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.14.7/bin/linux/amd64/kubectl
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.16.13/bin/linux/amd64/kubectl
 RUN chmod +x ./kubectl
 RUN mv ./kubectl /usr/local/bin/kubectl
 COPY admission-controller-init-scripts/admission_util.sh /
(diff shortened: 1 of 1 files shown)


wilfred-s:
Please hold of committing until we have it clear that we drop 1.14 from support in the next release. Officially we cannot support 1.14 if we use kubectl 1.16.

yangwwei:
Thanks for raising this up. Right, according to https://kubernetes.io/docs/setup/release/version-skew-policy/#kubelet. If we update kubelet to 1.16, that won't work with K8s 1.14/1.15. However, if we use kubelet 1.14, that is supposed to work with 1.14/1.15/1.16.
@HuangTing-Yao did you see any issue while using 1.14 kubelet working with K8s 1.16?

HuangTing-Yao:
@yangwwei Here is talking about kubectl, not kubelet.
And I haven't try kubelet 1.14 in k8s 1.16, both my kubelet and kubectl is 1.16.
according to [kubectl](https://kubernetes.io/docs/setup/release/version-skew-policy/#kubectl), if we use kubectl 1.16 that can only support to k8s 1.15.
Or should we modify kubectl in dockerfile to 1.15.x ?

yangwwei:
> @yangwwei Here is talking about kubectl, not kubelet.
Oh yes, thanks for the correction.

> And I haven't try kubelet 1.14 in k8s 1.16, both my kubelet and kubectl in cluster is 1.16, the other kubectl in scheduler image is 1.14.7. It doesn't show any big issue.
> according to [kubectl](https://kubernetes.io/docs/setup/release/version-skew-policy/#kubectl), if we use kubectl 1.16 that can only support to k8s 1.15.
> Or should we modify kubectl in dockerfile to 1.15.x ?

Yep. I think we should try 1.15. That looks to be safer.

wilfred-s (approved):
lgtm, committing the change

PR #184, opened 2020-08-25
[YUNIKORN-376] Update K8s dependency to 1.16 in k8shim repo

yangwwei:
This is a refined PR after #181 
Trying to minimize the changes in this PR. 
- update the go.mod file to depend on K8s 1.16.13 as the dependency
- re-generate the code for app CRD
- remove deprecated predicates functions that no longer supported in this new version


wilfred-s (approved):
lgtm, we have discussed this as part of the prometheus changes for the core and SI already.
The current cleanup looks good

PR #185, opened 2020-08-28
[YUNIKORN-386] Retrieve application ID from pod annotation

kingamarton:
@yangwwei, @TaoYang526  can you please helm me with the review?


pkg/common/utils/utils.go
```diff
@@ -82,24 +82,30 @@ func GetQueueNameFromPod(pod *v1.Pod) string {
...
 
 func GetApplicationIDFromPod(pod *v1.Pod) (string, error) {
-	for name, value := range pod.Labels {
-		// if a pod for spark already provided appID, reuse it
-		if name == constants.SparkLabelAppID {
+	// application ID can be defined in annotations
+	for name, value := range pod.Annotations {
+		if name == constants.LabelApplicationID {
+			return value, nil
+		}
```
yangwwei, line 89: these lines should not be here, this is already covered by line 98 - 100, I think we can remove them

kingamarton: Is not the same. Here we check the annotations and in lines 89-100 we check the labels

yangwwei: I don't think we need to check `constants.LabelApplicationID` here in this case.
When we ask users to specify annotations, that should follow the form: "yunikorn.apache.org/app-id".

kingamarton: I removed constants.LabelApplicationID from the annotations part and also renamed the SparkAnnotationAppID to AnnotationApplicationID, since we will use it for other kind of apps as well.

yangwwei: Thx

yangwwei:
hi @kingamarton  could u pls fix the remaining issue in this PR?

kingamarton:
> hi @kingamarton could u pls fix the remaining issue in this PR?

Sure, I deleted the failing TestGetApplicationIDFromPod from the application_test, because there is a test case for that method in utils_test.go. That test case belongs to the utils_test and not to application_test, so I kept this last one.

yangwwei (approved):
LGTM +1

PR #186, opened 2020-09-01
[YUNIKORN-389] Replace vanity import for vbom.ml

wilfred-s:
vbom.ml is an indirect dependency which seems to cause issues while
resolving as an import. The project has recently removed the vanity import
for the full github reference.
Even the latest versions of kubernetes seem to rely on this import.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -61,3 +61,5 @@ replace (
 	k8s.io/metrics => k8s.io/metrics v0.16.13
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.16.13
 )
+
+replace vbom.ml/util => github.com/fvbommel/util v0.0.0-20160121211510-db5cfe13f5cc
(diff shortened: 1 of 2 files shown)


wilfred-s:
The part of the test that we wanted to pass was the make this worked.
Based on that the change is good.

the e2e test that failed is know and really covered by YUNIKORN-317 as the root cause

adamantal (approved):
LGTM, this is exactly what the guys mentioned in the PR you commented on the issue as a workaround.

wilfred-s:
@yangwwei I explicitly did not put the replace for `vbom.ml` inside the other replace set. The replace that are part of the set are all K8s version dependencies and they will change with the version of K8s. The `vbom.ml` is not linked to K8s version, it is linked to the go version that we support in the build. I should have put a comment with it to make that clear.

It is now not really clear why we did this as it is hidden in the other replace tags.

PR #190, opened 2020-09-11
[YUNIKORN-410] Pod state change may cause incorrect update on SchedulerNode#occupied

TaoYang526:
Details please refer to [YUNIKORN-410](https://issues.apache.org/jira/browse/YUNIKORN-410)

### DIFF
--- a/pkg/cache/node_coordinator_test.go
+++ b/pkg/cache/node_coordinator_test.go
@@ -248,3 +248,45 @@ func TestDeleteTerminatedPod(t *testing.T) {
 	coordinator.deletePod(pod2)
 	assert.Equal(t, executed, false)
 }
+
+func TestUpdatePodPendingRunningAlternatively(t *testing.T) {
+	mockedSchedulerAPI := newMockSchedulerAPI()
+	nodes := newSchedulerNodes(mockedSchedulerAPI, NewTestSchedulerCache())
+	host1 := utils.NodeForTest("HOST1", "10G", "10")
+	nodes.addNode(host1)
+	coordinator := newNodeResourceCoordinator(nodes)
+
+	// init pod (pod1) and changed pod (pod2)
+	pod1 := utils.PodForTest("pod1", "1G", "500m")
+	pod1.SetUID("UID-01")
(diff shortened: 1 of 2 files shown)


yangwwei:
hi @TaoYang526 

I am trying to understand how this could happen. In the code, we already did a check

```
oldPod.Status.Phase != newPod.Status.Phase
```

that means when we try to update the occupied resource, that means the pod is transiting from `Pending` to `Running`. 

Pod phases are: `Pending`, `Running`, `Succeeded`, `Failed`, `Unknown`. Among them, `Succeed` and `Failed` are terminated state, which means they cannot transit to `Running` at all. So the only possibility is from `Pending` to `Running`, or `Unknown` to `Running`. The case you are seeing, is it from `Unknown` to `Running`?

pkg/cache/node_coordinator_test.go
```diff
@@ -248,3 +248,45 @@ func TestDeleteTerminatedPod(t *testing.T) {
...
+	pod1.Status.Phase = v1.PodPending
+	pod2.Status.Phase = v1.PodRunning
+	pod1.Spec.NodeName = "HOST1"
+	pod2.Spec.NodeName = "HOST1"
+	coordinator.updatePod(pod1, pod2)
+	node1 := coordinator.nodes.nodesMap["HOST1"]
+	assert.Equal(t, node1.occupied.Resources[constants.Memory].Value, int64(1000))
+	assert.Equal(t, node1.occupied.Resources[constants.CPU].Value, int64(500))
+
+	// pod state changed from running to pending,
```
yangwwei, line 275: Can pod state change from `Running` to `Pending`?

wilfred-s:
I echo the question from @yangwwei on the fact that a pod status should not be going back to pending.

Based on the response that is part of an old [kubernetes issue](https://github.com/kubernetes/kubernetes/issues/10319), and the [pod documentation](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/) a pod should never go back to a pending state.

The status is a one way change from pending -> running -> (succeeded || failed)

PR #192, opened 2020-09-21
[YUNIKORN-420]Add TaskGroup field in app-CRD

HuangTing-Yao:
Add taskGroup and schedulingPolicy to application-definition.yaml and yunikorn.apache.org/type.go 
And also re-generate the crd code with code-generator.
@yangwwei Pls review. Thanks.


deployments/yunikorn-application/application-definition.yaml
```diff
@@ -50,14 +50,40 @@ spec:
...
+                    type: object
+                    properties:
+                      cpu:
+                        type: integer
+                        format: int32
+                        minimum: 1
+                      memory:
+                        type: integer
+                        format: int32
+                        minimum: 1
```
yangwwei, line 86: this should support other resource types, other than cpu and memory
could u pls check how resources is being set in `pod` and I think we should do the similar thing here.
the format is not supposed to be `int32`, see more in https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-units-in-kubernetes.

deployments/yunikorn-application/application-definition.yaml
```diff
@@ -50,14 +50,40 @@ spec:
...
+              type: string
+            policy:
+              type: object
+              properties:
+                policy:
+                  type: string
+                timeout:
+                  type: integer
+                retrySecond:
+                  type: integer
```
yangwwei, line 63: I think this should be a map that supports arbitrary key-value pairs.
because different policies will have different parameter names, and values

yangwwei (changes requested):

deployments/yunikorn-application/sample-yunikorn-app.yaml
```diff
@@ -20,5 +20,19 @@ kind: Application
 metadata:
   name: example
 spec:
+  appId: "app0001"
```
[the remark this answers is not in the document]
kingamarton: I agree. But I think we should publish the appID. Right now, the definition does not have any appID field. I think it will be very useful to publish it. Right now the used doesn't know the appId after submitting the CRD. He/she will have to check the webUI or to recall it from the docs/code that it is in the form of namespace-name.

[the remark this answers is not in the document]
HuangTing-Yao: Or we should move `AppID` from spec field to status field ?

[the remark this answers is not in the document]
yangwwei: Yes, I think we can do that.

pkg/apis/yunikorn.apache.org/v1alpha1/type.go
```diff
@@ -33,15 +34,42 @@ type Application struct {
...
+
+type PolicyParam string
+
+const (
+	Timeout     PolicyParam = "Timeout"
+	RetrySecond PolicyParam = "RetrySecond"
+)
+
+type Task struct {
+	Name        string                       `json:"taskName"`
```
yangwwei, line 67: this should be `groupName`
as we are declaring multiple gang groups here

pkg/apis/yunikorn.apache.org/v1alpha1/type.go
```diff
@@ -33,15 +34,42 @@ type Application struct {
...
-	MaxPendingSeconds int32  `json:"maxPendingSeconds,omitempty"`
+	Policy    SchedulePolicy `json:"schedulingPolicy"`
+	Queue     string         `json:"queue"`
+	TaskGroup []Task         `json:"taskGroups"`
 }
 
-//Status part
+type SchedulePolicy struct {
+	Policy    SchedulingPolicy      `json:"name"`
+	Parameter map[PolicyParam]int32 `json:"parameter,omitempty"`
```
yangwwei, line 47: 2 comments for this:
- this should be `Parameters`, as we allow to set multiple options
- I prefer to define the value as `string` type, use int will not give us enough flexibility

pkg/apis/yunikorn.apache.org/v1alpha1/type.go
```diff
@@ -33,15 +34,42 @@ type Application struct {
...
+	TryReserve SchedulingPolicy = "TryReserve"
+	TryPreempt SchedulingPolicy = "TryPreempt"
+)
+
+type PolicyParam string
+
+const (
+	Timeout     PolicyParam = "Timeout"
+	RetrySecond PolicyParam = "RetrySecond"
+)
```
yangwwei, line 64: Every policy will have different parameters. We do not need to pre-define them here.
It can be just a map of strings.

yangwwei (changes requested):

pkg/apis/yunikorn.apache.org/v1alpha1/type.go
HuangTing-Yao: done.

pkg/apis/yunikorn.apache.org/v1alpha1/type.go
HuangTing-Yao: done.

pkg/apis/yunikorn.apache.org/v1alpha1/type.go
HuangTing-Yao: done.

yangwwei (approved):
Thanks for fixing all the comments, LGTM.

PR #193, opened 2020-09-28
[YUNIKORN-426] Refactor Spark deployment demo to remove some hardcode

yanghua:
jira detail: https://issues.apache.org/jira/browse/YUNIKORN-426


yanghua:
cc @yangwwei

deployments/examples/spark30/cmd/run.sh
```diff
@@ -16,12 +16,40 @@
...
+
+if [ -f $SPARK_BINARY_FILE_PATH ]; then
+  echo "The binary file $SPARK_BINARY_FILE_NAME has been cached!"
+else
+  echo "The binary file $SPARK_BINARY_FILE_NAME did not exist, try to download."
+  wget http://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/${SPARK_BINARY_FILE_NAME} -O ${SPARK_BINARY_FILE_PATH}
+fi
+
+# check signature to verify the completeness
+if [[ $(cat $SPARK_BINARY_CHECKSUM_FILE_PATH) == $(shasum -a 512 $SPARK_BINARY_FILE_PATH | awk '{print $1}') ]]; then
```
wilfred-s, line 45: If you want to check the checksum you must also pull down the checksum from the source.
Don't store the checksum in our code. If someone decides to change the version to 3.0.1 in the script this will break.

deployments/examples/spark30/cmd/run.sh
```diff
@@ -16,12 +16,40 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# replace these env var values accordingly
-SPARK_HOME=/Users/example/repository/spark
+SCRIPT_PATH=$(cd `dirname $0`; pwd)
+# set up root directory
+WORK_SPACE_ROOT=`dirname $SCRIPT_PATH`
```
wilfred-s, line 21: This forces a run from within the checked out tree. Which means you remove the flexibility from the user to have or use multiple inputs and setups.
The user should have the flexibility to run to run this from wherever they want on the system.

deployments/examples/spark30/cmd/run.sh
```diff
@@ -16,12 +16,40 @@
...
+
+SPARK_VERSION=3.0.0
+SPARK_HADOOP_VERSION=2.7
```
wilfred-s, line 24: See below, since you do not download the checksum file this is really still fixed to one specific version.
We should also be able to pick up these values from outside the script, for both spark and hadoop. Something like this:
```
if [ "$SPARK_VERSION" = "" ]; then
    echo using default Spark version;
    SPARK_VERSION=3.0.0
fi
```

deployments/examples/spark30/cmd/run.sh
```diff
@@ -16,12 +16,40 @@
...
 K8S_ENDPOINT=http://localhost:8001
 SPARK_EXECUTOR_NUM=1
 SPARK_DOCKER_IMAGE=yunikorn/spark:latest
-SPARK_EXAMPLE_JAR=local:///opt/spark/examples/jars/spark-examples_2.12-3.0.0-SNAPSHOT.jar
+
+if [ -f $SPARK_BINARY_FILE_PATH ]; then
+  echo "The binary file $SPARK_BINARY_FILE_NAME has been cached!"
+else
+  echo "The binary file $SPARK_BINARY_FILE_NAME did not exist, try to download."
+  wget http://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/${SPARK_BINARY_FILE_NAME} -O ${SPARK_BINARY_FILE_PATH}
```
wilfred-s, line 41: checksum and binary if you want to auto download

deployments/examples/spark30/cmd/run.sh
```diff
@@ -16,12 +16,40 @@
...
+fi
+
+# check signature to verify the completeness
+if [[ $(cat $SPARK_BINARY_CHECKSUM_FILE_PATH) == $(shasum -a 512 $SPARK_BINARY_FILE_PATH | awk '{print $1}') ]]; then
+  echo "The checksum is matched!"
+  rm -rf $WORK_SPACE_ROOT/spark-${SPARK_VERSION}-bin-hadoop${SPARK_HADOOP_VERSION}
+  tar -xvzf $SPARK_BINARY_FILE_PATH -C $WORK_SPACE_ROOT
+else
+  echo "The checksum is not matched, Removing the incompleted file, please download it again."
+  rm -f ./binarycache/${SPARK_BINARY_FILE_NAME}
```
wilfred-s, line 51: This is the first mention of a `binarycache` directory, this will never work.

deployments/examples/spark30/cmd/run.sh
```diff
@@ -16,12 +16,40 @@
...
+  echo "The binary file $SPARK_BINARY_FILE_NAME has been cached!"
+else
+  echo "The binary file $SPARK_BINARY_FILE_NAME did not exist, try to download."
+  wget http://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/${SPARK_BINARY_FILE_NAME} -O ${SPARK_BINARY_FILE_PATH}
+fi
+
+# check signature to verify the completeness
+if [[ $(cat $SPARK_BINARY_CHECKSUM_FILE_PATH) == $(shasum -a 512 $SPARK_BINARY_FILE_PATH | awk '{print $1}') ]]; then
+  echo "The checksum is matched!"
+  rm -rf $WORK_SPACE_ROOT/spark-${SPARK_VERSION}-bin-hadoop${SPARK_HADOOP_VERSION}
```
wilfred-s, line 47: If we have an unpacked version already why do we not detect that and use it instead of checking for the compressed archive?

wilfred-s (changes requested):
The idea is good but it has implementation issues at the moment.

deployments/examples/spark30/cmd/run.sh
yanghua: @wilfred-s I have tried to download the checksum from the official website. But there are two issues:

- We can not make sure the checksum file is completeness.
- The official sha512 file is more complex, e.g. blank and capital and small letter, we need to spend more time to verify them. Considering that we will have more framework demos in the future, each checksum file is different.

It seems we can just copy the checksum code and store it into a variable. I saw Flink did this. WDYT?

deployments/examples/spark30/cmd/run.sh
yanghua: We can give a parameter to let the user specify the location.

deployments/examples/spark30/cmd/run.sh
yanghua: I thought about this problem, but we cannot guarantee whether the existing folder has been changed. If it is unpredictable, why not just delete it based on the same standard file first. The decompression process does not consume more time.

Here, I think our goal is to let new users experience it. If it is not sufficiently automated and requires too high learning costs, then it will bring some rebellious emotions.

WDYT?

deployments/examples/spark30/cmd/run.sh
yanghua: But we have `spark24` demo dir. Maybe we can merge these two isolated demos through variables?

wilfred-s: Sure that would be nice, just one version to maintain.

deployments/examples/spark30/cmd/run.sh
wilfred-s: The offical spark checksum is generated using this command:
```
gpg --print-md sha512 spark-3.0.1-bin-hadoop2.7.tgz
spark-3.0.1-bin-hadoop2.7.tgz: F4A10BAE C5B8FF18 41F10651 CAC2C4AA 39C162D3
                               029CA180 A9749149 E6060805 B5B5DDF9 287B4AA3
                               21434810 172F8CC0 534943AC 005531BB 48B6622F
                               BE228DDC
```
There are really only two versions: `gpg` and `shasum`

It is also really simple to generate the format you want using:
```
cat spark-3.0.1-bin-hadoop2.7.tgz.sha512 | tr -d " \t\n\r" | awk -F: '{print $2 "  " $1}' > converted.sha512
```
That output is the same as the shasum output. Redirect into a file and use `shasum -c -a 512 converted.sha512` and you are good to go

deployments/examples/spark30/cmd/run.sh
wilfred-s: happy with that too

deployments/examples/spark30/cmd/run.sh
wilfred-s: Make the behaviour clear, always using a clean unpacked version is OK but the user needs to be made aware that you are doing it.

deployments/examples/spark30/cmd/run.sh
yanghua: OK, let me try.

yanghua:
@wilfred-s I have addressed most of the suggestions. Please review.

yanghua:
@wilfred-s Any opinion?

wilfred-s (approved):
lgtm
merging the change

PR #195, opened 2020-09-30
[YUNIKORN-429] Set default logging level to INFO

adamantal:
Updated the simple change, will test this and will update this PR with further log level optimizations.


yangwwei:
hi @adamantal  I think the default value in here: https://github.com/apache/incubator-yunikorn-k8shim/blob/1299259cc5899adc9c6fb4c2bf8055c45a952985/pkg/conf/schedulerconf.go#L40 also needs to be updated.

adamantal:
Thank you for checking this review @kingamarton !

The default value is `0`, which is the `INFO` level, so no need to change it there. I think YK always overrides that value, so I think that is fine. Even when we run locally against docker-desktop we override that value with DEBUG, but I think for local development it's ok to keep it at that level.

I also checked the helm charts, but since we're just using the Docker containers as they are, no need to make changes there.

adamantal:
Analyzed the logs of a locally run scheduler with some load and I suggest the following two minor changes:
- `updatePodCondition` does not seem to be an important log entry, I decreased to DEBUG level from INFO
- removed one INFO level log from node part of `context_recovery.go`. It has been displayed already a few lines above

yangwwei:
> The default value is 0, which is the INFO level, so no need to change it there. I think YK always overrides that value, so I think that is fine.

My bad. This sounds good.

@adamantal  are you going to track adding/removing INFO level logs in another JIRA, or you are going to make those changes also here? Just want to understand when we can merge this.

adamantal:
I have two PRs regarding the log level harmonization, one for the core (https://github.com/apache/incubator-yunikorn-core/pull/211) and another for the shim (this one). I think these two depend on each other, so they can be both handled under YUNIKORN-429.

yangwwei (approved):
LGTM

yangwwei:
Since this PR is only changing log lines, there is no need to add more UT for this. We can ignore the UT coverage warnings.

PR #197, opened 2020-10-03
[YUNIKORN-435] Use serviceAccountName from env for admission_controller pod

bmv126:
The admission controller pod goes into pending state as it is created with default service account.
The yunikorn-admin serviceAccount is not used for this pod creation.

Reference: https://issues.apache.org/jira/browse/YUNIKORN-435


yangwwei (approved):
LGTM. @kingamarton pls give another look at this. THX

kingamarton (approved):

PR #198, opened 2020-10-08
[YUNIKORN-438] Wait for sync after apiFactory start

kingamarton:


pkg/client/apifactory.go
```diff
@@ -203,6 +206,10 @@ func (s *APIFactory) Start() {
 	// launch clients
 	if !s.IsTestingMode() {
 		s.clients.Run(s.stopChan)
+		if err := s.WaitForSync(); err != nil {
```
wilfred-s, line 209: We already know we're not in test mode why are we not calling `s.clients.WaitForSync(time.Second, 30*time.Second)` directly?

wilfred-s (commented):
The other point to make (not really part of this change) we wrap returning a boolean value in a function without locking `IsTestingMode()` and use it internally. That seems just overhead for nothing inside the factory.
I can understand that from an external point of view but not for internal use.

pkg/client/apifactory.go
kingamarton: Changed it.

wilfred-s (approved):
change looks good

PR #202, opened 2020-10-26
[YUNIKORN-449] Added MatchNodeSelectorPred to reservation predicates list

kingamarton:


yangwwei:
hi @kingamarton the change looks good to me. Could you please try to verify this in on your real K8s setup to ensure it works as expected? Thanks!

kingamarton:
> hi @kingamarton the change looks good to me. Could you please try to verify this in on your real K8s setup to ensure it works as expected? Thanks!

@yangwwei, I tested it in my local setup and it worked only if I passed in the predicates, so I extended the predicate list for the default scheduling policy as well. After the change the reservation is not created if the nodeLabels are not matching and the pod remains in Pending state.

yangwwei:
Hi @kingamarton thanks for testing this out. 
> so I extended the predicate list for the default scheduling policy as well.
If this is added to the `defaultSchedulerPolicy` list, it will be overlapping with `predicates.GeneralPred`. If you follow the code traces, you'll see `predicates.GeneralPred` calls `PodMatchNodeSelector`. That means while it tries to `predicatesAllocate()`, it runs the node-selector twice.

I guess we need to fix some other way round. The problem you met might be caused by the CLI parser. Please take a look at https://github.com/apache/incubator-yunikorn-k8shim/pull/64. It adds the support to give a configurable list of predicates in CLI options. I think the fix here is, when the CLI is not providing any input, the default ones should come from both `defaultSchedulerPolicy` and `reservationPredicates`. Please let me know your thought about this.

Thanks

kingamarton:
> Hi @kingamarton thanks for testing this out.
> 
> > so I extended the predicate list for the default scheduling policy as well.
> 
> If this is added to the `defaultSchedulerPolicy` list, it will be overlapping with `predicates.GeneralPred`. If you follow the code traces, you'll see `predicates.GeneralPred` calls `PodMatchNodeSelector`. That means while it tries to `predicatesAllocate()`, it runs the node-selector twice.
> 
> I guess we need to fix some other way round. The problem you met might be caused by the CLI parser. Please take a look at #64. It adds the support to give a configurable list of predicates in CLI options. I think the fix here is, when the CLI is not providing any input, the default ones should come from both `defaultSchedulerPolicy` and `reservationPredicates`. Please let me know your thought about this.
> 
> Thanks

@yangwwei I thought about how we can solve this issue. What we discussed offline: to have an another list with the allocationPredicates might not work, and registering the reservations as well, might not work, because if there are some predicates passed in the CLI, we might end in checking some extra predicates for a reservation than for the allocation. 

Instead of this, I expanded the GeneralPredicate and added each predicate separately. I skipped one: PodFitsPorts, what has been replaced by PodFitsHostPorts, and since we don't support K8s 1.0, I think we don't need that.

yangwwei:
> Instead of this, I expanded the GeneralPredicate and added each predicate separately. I skipped one: PodFitsPorts, what has been replaced by PodFitsHostPorts, and since we don't support K8s 1.0, I think we don't need that.

Hi @kingamarton I am not sure if it is a good idea to replace `GeneralPred`, I just checked K8s 1.17 code, https://github.com/kubernetes/kubernetes/blob/a42a619a5f2bde896bcc63a990fb3b42d55dffe3/pkg/scheduler/algorithmprovider/defaults/defaults.go#L40-L54. It still uses `GeneralPred`, not the separate ones.

One difference I can see is in `GeneralPred`: https://github.com/kubernetes/kubernetes/blob/a42a619a5f2bde896bcc63a990fb3b42d55dffe3/pkg/scheduler/algorithm/predicates/predicates.go#L1144-L1189. It runs a bunch of things:
1. PodFitsResources
2. PodFitsHost
3. PodFitsHostPorts
4. PodMatchNodeSelector

what you have in the PR:
1. PodFitsResourcesPred
2. PodFitsHostPortsPred
3. MatchNodeSelectorPred

where misses `PodFitsHost`. Even you add this one, I still have some concern there might be something out of our radar, could you please take a closer look at the predicates functions and let me know if this is a safe change? Predicates are invoked in every scheduling cycle, we need to be extremely careful while making changes.

kingamarton:
> where misses `PodFitsHost`. Even you add this one, I still have some concern there might be something out of our radar, could you please take a closer look at the predicates functions and let me know if this is a safe change? Predicates are invoked in every scheduling cycle, we need to be extremely careful while making changes.

Oh, you are right. I mixed it with the `PodFitsPorts`. I added the missing predicate and also checked the functions. The same functions are called. I found only one minor difference. With this approach, we will return in case of the first predicate failure, and while using GeneralPredicates all the predicates are checked and all the failures are collected. However I don't consider this one an important difference, because in case of the other predicates the behaviour is the same: only the first failure is returned.

yangwwei:
hi @kingamarton  another difference I noticed:

```
// GeneralPredicates checks whether noncriticalPredicates and EssentialPredicates pass. noncriticalPredicates are the predicates
// that only non-critical pods need and EssentialPredicates are the predicates that all pods, including critical pods, need
```

if we remove `GeneralPred`, that means we are running all these predicates for every pod, no matter they are critical or non-critical, this could be a behavior change.

yangwwei:
hi @kingamarton 

Thanks for the patch, I've reviewed it but I found it is not very easy to read. I guess we need some refactoring. Here is my thought on how we can make the structure more readable:

1. Change the predicate related configs to the following

```
type SchedulerConf struct {
  ...
  AllocationPredicates            string        `json:"allocationPredicates"`
  ReservationPredicates        string        `json:"reservationPredicates"`
  ...
}
```

this allows us to config both allocating/reserving phase predicates separately. In the future, this will be more flexible.

2. Move the default list of to-be-registered predicates to `schedulerconf.go`

```
// current the default value is "", we can change this to the values we want to include
// for both allocating and reserving phase predicates
predicateList := flag.String("predicates", "",
		fmt.Sprintf("comma-separated list of predicates, valid predicates are: %s, "+
			"the program will exit if any invalid predicates exist.", predicates.Ordering()))
```

this means users can configure what kind of options they want to run for each of the phases separately. We do not need any hardcoded values anymore. 

3. Modify the `parseConfiguredSchedulerPolicy()` to parse predicate list from both configs. We can define a new local struct like:

```
type predicatePolicies struct {
  allocatePredMap  map[string]PredicatePolicy. // predicates function for allocating phase
  reservePredMap   map[string]PredicatePolicy // predicates functions for reserving phase
}
```
all these policies will be registered by `p.populatePredicateFunc(*args)`.

4. then we just need to slightly modify the `predicatesAllocate()` and `predicatesReserve()`.  We can do

```
func (p *Predictor) predicatesAllocate(...) error {
	for _, predicateKey := range predicates.Ordering() {
           if p, ok := predicatePolicies.allocatePredMap[predicateKey]; ok {
               //....
           }
	}
	return nil
}

func (p *Predictor) predicatesReserve(...) error {
	for _, predicateKey := range predicates.Ordering() {
           if p, ok := predicatePolicies.reservePredMap[predicateKey]; ok {
               //....
           }
	}
	return nil
}
```

please let me know if such refactoring makes sense to you or not. Thanks

kingamarton:
Hi @yangwwei thank you for the review. I agree that the actual approach is hard to read and it is a kind of hack. 
I find your suggestion a dangerous one, because of the following reasons:
- the client might configure some resource related predicates for the reservation list, what will always fail, because when the reservations are created it means that there aren't enough resources in the cluster
- also the client might configure different predicates for the reservation that for the allocation, what also might cause unwanted behaviours.

After analysing all this aspects I still think that the cleaner and safer solution is to expand the `GeneralPred` into the 4 predicates. I also checked the code again and the comment  and the implementation of the generalPredicate function:
`// GeneralPredicates checks whether noncriticalPredicates and EssentialPredicates pass` - I think we misunderstood this comment, because here it clearly says that both predicate types are checked. The rest of the comment is just a small explanation what this predicates are.

I also checked the ordering, and there will be no difference either. So I will revert this last change set and keep the expanded predicate list.

yangwwei:
Hi @kingamarton 

Thanks for the efforts for the deep analysis, your comment makes sense to me.
I think we can take your approach, let's run enough tests to make sure nothing breaks.
Change looks good to me. +1

yangwwei (approved):

PR #204, opened 2020-10-29
[YUNIKORN-439]Parse taskGroup info from pod annotation

HuangTing-Yao:
Parse taskGroups info from pod annotation.
Please review @yangwwei 
Thanks.

PR #205, opened 2020-10-30
[YUNIKORN-451]Fix application CRD definition

HuangTing-Yao:
Fix application CRD yaml which accord  to v1alpha1/type.go 
Please review @yangwwei


yangwwei:
hi @HuangTing-Yao . I have merged #204 . For this one, could you please rebase and re-submit?
Please double-check how to cover this in our e2e test, thanks

HuangTing-Yao:
Modify the app e2e test, so we can apply application with kubectl, and let application-definition.yaml do the schema check.

yangwwei:
hi @HuangTing-Yao overall it looks good. Please see my comments below:

1. I think we can wrap the `os.exec` commands to a helper method in https://github.com/apache/incubator-yunikorn-k8shim/blob/master/test/e2e/framework/helpers/k8s/k8s_utils.go. So it can be reused easily in the future.
2. We need some more negative tests for app CRD, in order to validate the schema check. Things can be checked including:
  - incorrect field names
  - incorrect minMember type
  - incorrect resource values
  - incorrect format of node-selector and tolerations
please define cases like the above in separate files and verify the creation should be failed
thanks

HuangTing-Yao:
The e2e test will fail is caused by the old version of application-definition.yaml in release repo.

yangwwei (approved):
+1

PR #206, opened 2020-11-20
[YUNIKORN-458]Retrieve app taskGroups from pod spec

HuangTing-Yao:
Retrieve taskGroup and taskGroupName from pod annotations.
And set them to application.
Please review @yangwwei . Thanks.


yangwwei (approved):
+1

PR #207, opened 2020-11-23
[YUNIKORN-459]Implement the placeholder cleanup in PlaceholderManager

HuangTing-Yao:
Clean up all placeholder pod and task.
Please review @yangwwei 
Thanks.

### DIFF
--- a/pkg/cache/placeholder_manager.go
+++ b/pkg/cache/placeholder_manager.go
@@ -70,10 +70,28 @@ func (mgr *PlaceholderManager) createAppPlaceholders(app *Application) error {
 	return nil
 }
 
-// recycle all the placeholders for an application
-func (mgr *PlaceholderManager) Recycle(appID string) {
-	log.Logger().Info("start to recycle app placeholders",
-		zap.String("appID", appID))
+// clean up all the placeholders for an application
+func (mgr *PlaceholderManager) CleanUp(app *Application) {
+	log.Logger().Info("start to clean up app placeholders",
+		zap.String("appID", app.GetApplicationID()))
+	for taskID, task := range app.taskMap {
+		if task.GetTaskPlaceholder() {
+			// remove pod
+			err := mgr.clients.KubeClient.Delete(task.pod)
+			if err != nil {
+				log.Logger().Error("failed to clean up placeholder pod",
+					zap.Error(err))
+			}
+			// remove task from application
+			err = app.removeTask(taskID)
+			if err != nil {
+				log.Logger().Error("failed to remove task from application",
+					zap.Error(err))
+			}
+		}
+	}
+	log.Logger().Info("finish to clean up app placeholders",
+		zap.String("appID", app.GetApplicationID()))
 }
 
 // this is only used in testing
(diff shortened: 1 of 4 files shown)


pkg/cache/placeholder_manager.go
```diff
@@ -70,10 +70,28 @@ func (mgr *PlaceholderManager) createAppPlaceholders(app *Application) error {
...
+func (mgr *PlaceholderManager) CleanUp(app *Application) {
+	log.Logger().Info("start to clean up app placeholders",
+		zap.String("appID", app.GetApplicationID()))
+	for taskID, task := range app.taskMap {
+		if task.GetTaskPlaceholder() {
+			// remove pod
+			err := mgr.clients.KubeClient.Delete(task.pod)
+			if err != nil {
+				log.Logger().Error("failed to clean up placeholder pod",
+					zap.Error(err))
```
yangwwei, line 83: When a deletion fails, if we do not do anything, this will create an "orphan" pod, which will leak some resources on the cluster, no good. I've created a JIRA: https://issues.apache.org/jira/browse/YUNIKORN-469 to handle such failures.

pkg/cache/placeholder_manager.go
```diff
@@ -70,10 +70,28 @@ func (mgr *PlaceholderManager) createAppPlaceholders(app *Application) error {
...
+			}
+			// remove task from application
+			err = app.removeTask(taskID)
+			if err != nil {
+				log.Logger().Error("failed to remove task from application",
+					zap.Error(err))
+			}
```
yangwwei, line 90: second thought, let's remove this. We do not need to remove the task from the app while cleaning up the placeholders. Task needs to be cleaned up separately, the normal process involves the core side to release corresponding resources in the scheduler.

pkg/cache/placeholder_manager_test.go
```diff
@@ -86,3 +88,67 @@ func TestCreateAppPlaceholders(t *testing.T) {
...
+			UID:  "UID-02",
+		},
+	}
+	taskID1 := "task01"
+	task1 := NewTask(taskID1, app, mockedContext, pod1)
+	task1.placeholder = true
+	app.taskMap[taskID1] = task1
+	taskID2 := "task02"
+	task2 := NewTask(taskID2, app, mockedContext, pod2)
+	task2.placeholder = true
```
yangwwei, line 132: Can we add another task with placeholder flag = false?

pkg/cache/placeholder_manager.go
HuangTing-Yao: OK, I have add `orphanpod` map in placeholder manager. 
And leverage it in YUNIKORN-469

pkg/cache/placeholder_manager.go
HuangTing-Yao: OK, done.

pkg/cache/placeholder_manager_test.go
HuangTing-Yao: OK, done.

yangwwei (approved):
+1

PR #212, opened 2020-12-04
[YUNIKORN-477] Include K8s 1.18 to e2e test matrix

yangwwei:


kingamarton (approved):
LGTM, +1. The coverage check is broken I think, we will need to check why it is complaining about a file what is not even touched by the PR.

PR #213, opened 2020-12-16
[YUNIKORN-478]Handle app completion at the shim side

HuangTing-Yao:
Delete pod when receive `ReleaseAllocation` response, and delete application from context when receive `UpdatedApplication` response.
Please review @yangwwei 
Thanks.

### DIFF
--- a/pkg/cache/application_events.go
+++ b/pkg/cache/application_events.go
@@ -167,15 +167,15 @@ func (fe FailApplicationEvent) GetApplicationID() string {
 // ------------------------
 // Release application allocations
 // ------------------------
-type ReleaseApplicationEvent struct {
+type ReleaseAppAllocationEvent struct {
 	applicationID   string
 	allocationUUID  string
 	terminationType string
 	event           events.ApplicationEventType
 }
 
-func NewReleaseApplicationEvent(appID string, allocTermination si.AllocationRelease_TerminationType, uuid string) ReleaseApplicationEvent {
-	return ReleaseApplicationEvent{
+func NewReleaseAppAllocationEvent(appID string, allocTermination si.AllocationRelease_TerminationType, uuid string) ReleaseAppAllocationEvent {
+	return ReleaseAppAllocationEvent{
 		applicationID:   appID,
 		allocationUUID:  uuid,
 		terminationType: si.AllocationRelease_TerminationType_name[int32(allocTermination)],
@@ -183,17 +183,17 @@ func NewReleaseApplicationEvent(appID string, allocTermination si.AllocationRele
 	}
 }
 
-func (re ReleaseApplicationEvent) GetApplicationID() string {
+func (re ReleaseAppAllocationEvent) GetApplicationID() string {
 	return re.applicationID
 }
 
-func (re ReleaseApplicationEvent) GetArgs() []interface{} {
+func (re ReleaseAppAllocationEvent) GetArgs() []interface{} {
 	args := make([]interface{}, 2)
 	args[0] = re.allocationUUID
 	args[1] = re.terminationType
 	return args
 }
 
-func (re ReleaseApplicationEvent) GetEvent() events.ApplicationEventType {
+func (re ReleaseAppAllocationEvent) GetEvent() events.ApplicationEventType {
 	return re.event
 }
--- a/pkg/cache/application_events.go
+++ b/pkg/cache/application_events.go
@@ -18,7 +18,9 @@
 
 package cache
 
-import "github.com/apache/incubator-yunikorn-k8shim/pkg/common/events"
+import (
+	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/events"
+)
 
 // ------------------------
 // SimpleApplicationEvent simples moves application states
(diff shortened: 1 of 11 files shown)


yangwwei:
hi @HuangTing-Yao  there is some issue here. First, we need to update go.mod file in order to use the latest binary from scheduler-interface repo, where we get the new `UpdateResponse#AllocationRelease`. And one more thing I forgot to mention, from the doc:
```
 // Released allocations, this could be either ack from scheduler when RM asks to terminate some allocations.
  // Or it could be decision made by scheduler (such as preemption or timeout).
  repeated AllocationRelease releasedAllocations = 3;
```
we need to handle this differently based on the different TERMINATION_TYPE, 

```
 enum TerminationType {
    STOPPED_BY_RM = 0;          // Stopped or killed by ResourceManager (created by RM)
    TIMEOUT = 1;                // Timed out based on the executionTimeoutMilliSeconds (created by core)
    PREEMPTED_BY_SCHEDULER = 2; // Preempted allocation by scheduler (created by core)
    PLACEHOLDER_REPLACED = 3;   // Placeholder allocation replaced by real allocation (created by core)
  }
```

if `STOPPED_BY_RM`,  there is no operation needed in the shim side (today's behavior)
if otherwise i.e `TIMEOUT`, `PREEMPTED_BY_SCHEDULER` or `PLACEHOLDER_REPLACED`, we will need to trigger the pod delete.

pkg/callback/scheduler_callback.go
```diff
@@ -118,15 +118,34 @@ func (callback *AsyncRMCallback) RecvUpdateResponse(response *si.UpdateResponse)
 	for _, release := range response.ReleasedAllocations {
 		log.Logger().Debug("callback: response to released allocations",
 			zap.String("UUID", release.UUID))
+		// delete the allocated pod
+		app := callback.context.GetApplication(release.ApplicartionID)
```
yangwwei, line 122: Need to check if app is nil or not, to avoid nil pointer exception

pkg/callback/scheduler_callback.go
```diff
@@ -118,15 +118,34 @@ func (callback *AsyncRMCallback) RecvUpdateResponse(response *si.UpdateResponse)
...
+		for _, task := range app.taskMap {
+			if task.allocationUUID == release.UUID {
+				// TerminationType 0 mean STOPPED_BY_RM
+				if release.TerminationType != si.AllocationRelease_STOPPED_BY_RM {
```
yangwwei, line 126: we need to handle this check before iterating the task map in order to make this more efficient

yangwwei (changes requested):

pkg/callback/scheduler_callback.go
```diff
@@ -118,15 +118,36 @@ func (callback *AsyncRMCallback) RecvUpdateResponse(response *si.UpdateResponse)
...
+				for _, task := range app.GetTaskMap() {
+					if task.GetTaskAllocationUUID() == release.UUID {
+						err := task.DeleteTaskPod(task.GetTaskPod())
+						if err != nil {
+							log.Logger().Error("failed to delete pod", zap.Error(err))
+						}
+					}
+				}
+			}
+		}
```
yangwwei, line 135: let's use event handling to handle this, so we do not need to expose `GetApplicationInternal ()`,  `GetTaskMap()` functions. We can change this to:

```
ev := cache.NewReleaseAppAllocationEvent(app.GetApplicationID(), toReleaseAllocations, events.ReleaseAppAllocation)
dispatcher.Dispatch(ev)
```

in `application.go`, we just need to add another event `ReleaseAppAllocation`, and that triggers state transition from `Running` to `Running`. And in the handler, we can do the task pod cleanup.

pkg/callback/scheduler_callback.go
```diff
@@ -118,15 +118,36 @@ func (callback *AsyncRMCallback) RecvUpdateResponse(response *si.UpdateResponse)
...
-		//handle status update
+		// delete application from context
+		if updated.State == events.States().Application.Completed {
+			err := callback.context.RemoveApplicationInternal(updated.ApplicationID)
+			if err != nil {
+				log.Logger().Error("failed to delete application", zap.Error(err))
+			}
+		}
+		// handle status update
 		dispatcher.Dispatch(cache.NewApplicationStatusChangeEvent(updated.ApplicationID, events.AppStateChange, updated.State))
```
yangwwei, line 151: after we delete the app in line 145, it continues to handle the event in line 151. this seems to be problematic. 
should we add an `else` here?

yangwwei (changes requested):

yangwwei:
hi @HuangTing-Yao  any updates? If you have more questions, pls let me know. Thanks

HuangTing-Yao:
> hi @HuangTing-Yao any updates? If you have more questions, pls let me know. Thanks

almost done, I will upload new version today.

pkg/callback/scheduler_callback.go
HuangTing-Yao: OK, done.

pkg/callback/scheduler_callback.go
HuangTing-Yao: OK, done.

pkg/cache/application_events.go
```diff
@@ -160,3 +162,34 @@ func (fe FailApplicationEvent) GetArgs() []interface{} {
 func (fe FailApplicationEvent) GetApplicationID() string {
 	return fe.applicationID
 }
+
+// ------------------------
+// Release application
```
yangwwei, line 167: NIT: Release application allocations

pkg/cache/context.go
```diff
@@ -552,6 +552,16 @@ func (ctx *Context) RemoveApplication(appID string) error {
 	}
 }
 
+func (ctx *Context) RemoveApplicationInternal(appID string) error {
```
yangwwei, line 555: can you add a UT to cover this function?

pkg/cache/application_events.go
```diff
@@ -160,3 +162,34 @@ func (fe FailApplicationEvent) GetArgs() []interface{} {
...
+// ------------------------
+type ReleaseApplicationEvent struct {
+	applicationID  string
+	allocationUUID string
+	event          events.ApplicationEventType
+}
```
yangwwei, line 173: can you pls add the TERMINATION_TYPE in this message as well?
this gives us the flexibility in case we want to handle different types differently in the future

yangwwei (changes requested):
hi @HuangTing-Yao it looks pretty good now. could you pls check my latest comments? I only have some minor ones. after that, we can get this merged. Thanks!

pkg/cache/application_events.go
HuangTing-Yao: OK, done

pkg/cache/context.go
HuangTing-Yao: OK, done

pkg/cache/application_events.go
HuangTing-Yao: done.

yangwwei (approved):
LGTM, thanks @HuangTing-Yao

PR #223, opened 2021-01-26
[YUNIKORN-512] Remove redundant log from placeholder manager

yangwwei:
- Remove "cleanup orphan pods" log that gets printed every few second
- Update the dependency to the latest core


yangwwei:
@kingamarton could you please help to review? Thanks

kingamarton (approved):

PR #224, opened 2021-01-27
[YUNIKORN-518]  Placeholder manager failed to init during scheduler recovery

yangwwei:


wilfred-s (changes requested):
Logging with a level of FATAL is a really bad idea. That must be removed from the code.

Since `GetPlaceholderManager()` is called in the code without checks we must never return a nil. That means we initialise it always and do error checks on init. The shim should not start if it cannot start without it.
The return value of NewPlaceholderManager should be checked if needed. Similar to the way the app manager is initialised and tracked in the scheduler.

Is there any way we can add tests to cover this because I do not think that this is covered by any tests

yangwwei:
hi @wilfred-s the purpose of adding that FATAL error is to make sure we can crash the scheduler when placeholder manager was not initialized. And the reason we hit this issue was that the `NewPlaceholderManager()` was not called after `GetPlaceholderManager()` during recovery, so it crashed. This should never happen in production. If we remove that and we will get an NPE here, similar but with the FATAL error, we know definitely something goes wrong.

We can check the return value of `NewPlaceholderManager()` but that's really not the root cause. The fix is to adjust the order of services' starting to make sure the `NewPlaceholderManager()` gets called before `GetPlaceholderManager()`.

Does that make sense?

wilfred-s:
No, if we cannot work without a `PlaceholderManager` initialised we should not start up. We need tests for that.
What you are doing now is that we crash the scheduler when, by chance, a task group is requested by an application. That might not come for minutes, hours or even days after starting up. So for no reason you could crash the scheduler after running for days without an issue.
That is not acceptable, and a really bad user experience.

I agree that the fix is to correctly and always initialise the `PlaceholderManager`. However since we have no `nil` checks in the code for `GetPlacholderManager()` we need to make sure that we take down the scheduler immediately if it is not initialised correctly. We do the same when the `AppManagementService` does not start correctly.
Creation of the manager should be part of the `newShimScheduler()` and then in the run we need to check it (not sure if there is anything we need to do beside a simple not nil check) and start it. When it is not correct we call `ss.stop()` which  should also be nil safe to not cause a panic on shutdown.

wilfred-s (approved):
latest changes fix all my comments

PR #225, opened 2021-01-31
[YUNIKORN-531] migrate to github action

yangwwei:


.travis.yml
```diff
@@ -37,36 +37,11 @@ install: true
...
-      - make test
-      - make lint
-      after_success: bash <(curl -s https://codecov.io/bash)
     - stage: publish docker image
       deploy:
         provider: script
         script: make push
         on:
           branch: master
           condition: $TRAVIS_EVENT_TYPE = cron
```
wilfred-s, line 46: This needs to be replaced with a scheduled github action or something like it. We do need to be careful as this is one part of the [distribution guidelines](https://incubator.apache.org/guides/distribution.html#docker) which we are breaking:
```
on:
  schedule:
    - cron:  '0 0 * * *'
```

yangwwei: I leave this as an open item for now, let the Travis job push the docker images until we figure out how to do that with github action. I think I've tried to do this github action before, but I met some token issue unsolved. That was a few months ago, we probably need to fix this in a second step.

wilfred-s: The fact that we push the image means we break the distribution guidelines. The latest tag is pointing to unreleased code which based on the distribution guidelines point 6 we cannot do. We do not rely on the image for testing or other release work should we not turn it off until we have the solution worked out that complies with the distribution guidelines?

yangwwei: I think we can safely remove it, even for the e2e test today, we do the build from scratch.
Let me remove this, we can revisit if we need to have auto-push in the feature.

yangwwei:
The UT `TestValidateConfigMapWrongRequest` failure was because the error contains different messages on the box used to run github action (ubuntu). "Temp failure in name resolution" vs "no such host", Fixing it by simply verify the error is not nil, instead of detail messages.

wilfred-s (approved):
LGTM committing

PR #226, opened 2021-02-02
[YUNIKORN-485] docker image update

wilfred-s:
The current dockerfiles use a go based image for the scheduler and some
tools. There is no need to deploy with go as part of the image.

### DIFF
--- a/deployments/image/configmap/Dockerfile
+++ b/deployments/image/configmap/Dockerfile
@@ -15,7 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.12.13-alpine3.10
+FROM alpine:latest
 
 # admission controller bundles
 RUN apk add curl
--- a/deployments/image/gang/gangclient/dockerfile
+++ b/deployments/image/gang/gangclient/dockerfile
@@ -15,7 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.15.2
+FROM alpine:latest
 WORKDIR /gang/client
 ADD . /gang/client 
 ENTRYPOINT ["./simulation-gang-worker"]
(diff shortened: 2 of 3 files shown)


yangwwei:
hi @wilfred-s the changes look good to me, but can we hold on merging this before releasing v0.10?
It should be fine, but since this changes the base image, I am not sure if we are going to run into some issues in production.

kingamarton (approved):

kingamarton:
> hi @wilfred-s the changes look good to me, but can we hold on merging this before releasing v0.10?
> It should be fine, but since this changes the base image, I am not sure if we are going to run into some issues in production.

@yangwwei, I don't think it might cause any issues. We use the same alpine:latest image for the admission controller as well.

yangwwei:
> > hi @wilfred-s the changes look good to me, but can we hold on merging this before releasing v0.10?
> > It should be fine, but since this changes the base image, I am not sure if we are going to run into some issues in production.
> 
> @yangwwei, I don't think it might cause any issues. We use the same alpine:latest image for the admission controller as well.

Ideally, that's the case. But given we could not stabilize the v0.10 branch as far, my concern is to add this change could lead up to some risk. if not absolutely necessary, I suggest postponing this to post v0.10.

wilfred-s:
We standardise on one image for the scheduler and admission controller. That we should have done earlier.

This change fixes a 1 high and 3 medium impact security issues that are part of the base image we use for the scheduler.
I would say that this is a must before we push the next release. Just that is enough to fix this.

The other images for the gang code have more than 40 high impact security issues open. We should never use them.

yangwwei:
> This change fixes a 1 high and 3 medium impact security issues that are part of the base image we use for the scheduler.
> I would say that this is a must before we push the next release. Just that is enough to make this a must have fix.

Make sense, approving the change. Feel free to merge it. Thanks
BTW, we need to make sure our e2e test is good after this change.

yangwwei (approved):

PR #227, opened 2021-02-03
[YUNIKORN-536] Add resource requests and limits for the admission-controller

yangwwei:


deployments/admission-controllers/scheduler/templates/server.yaml.template
```diff
@@ -37,6 +37,13 @@ spec:
           ports:
           - containerPort: 9089
             name: webhook-api
+          resources:
+            limits:
+              cpu: "100m"
+              memory: 500Mi
+            requests:
+              cpu: "500m"
+              memory: 500Mi
```
kingamarton, line 46: Can you please explain how you decided about this values?

yangwwei: I've done some observations on a live cluster, leveraging kubectl top command, I was able to see the resource utilization of the admission controller pod, it is not using a lot of resources. I tried to create hundreds of pods in a loop to add some loads, it only uses 1~2m CPU, and less than 10m memory, pretty lightweight. So I think this number is sufficient.

wilfred-s (approved):
LGTM

PR #229, opened 2021-02-03
[YUNIKORN-509]Simplify the gangDeploy.sh

HuangTing-Yao:
reuse the gang-coordinator.yaml file and add timeout check.

### DIFF
--- a/deployments/examples/gang/cmd/gangDeploy.sh
+++ b/deployments/examples/gang/cmd/gangDeploy.sh
@@ -88,5 +88,4 @@ spec:
       schedulerName: yunikorn
 EOF
 )
-
 done
--- a/deployments/examples/gang/cmd/gangDeploy.sh
+++ b/deployments/examples/gang/cmd/gangDeploy.sh
@@ -33,7 +33,7 @@ WORKDIR=$(cd "$(dirname "$0")"; pwd)
 kubectl apply -f $WORKDIR/../gang-coordinator.yaml
 
 # wait for web server to be running
-until grep 'Running' <(kubectl get pod gangweb -o=jsonpath='{.status.phase}'); do
+until grep -q 'Running' <(kubectl get pod gangweb -o=jsonpath='{.status.phase}'); do
   sleep 1
   TIMEOUT=$(($TIMEOUT+1))
   if [ $TIMEOUT -ge 20 ]; then
@@ -72,5 +72,6 @@ spec:
           value: "$RUNTIMESEC"
       restartPolicy: Never
       schedulerName: yunikorn
-EOF)
+EOF
+)
 done
--- a/deployments/examples/gang/cmd/gangDeploy.sh
+++ b/deployments/examples/gang/cmd/gangDeploy.sh
@@ -50,7 +50,7 @@ until grep -q 'Running' <(kubectl get pod gangweb -o=jsonpath='{.status.phase}')
   sleep 1
   TIMEOUT=$(($TIMEOUT+1))
   if [ $TIMEOUT -ge 20 ]; then
-    echo "Timeout for waiting web server"
+    echo "ERROR: Timeout for waiting web server"
     kubectl delete -f $WORKDIR/gang-coordinator.yaml
     exit
   fi
--- a/deployments/examples/gang/cmd/gangDeploy.sh
+++ b/deployments/examples/gang/cmd/gangDeploy.sh
@@ -29,8 +29,8 @@ RUNTIMESEC=$3
 TIMEOUT=0
 
 # create service and job counter web server
-WORKDIR=$(cd "$(dirname "$0")"; pwd)
-kubectl apply -f $WORKDIR/../gang-coordinator.yaml
+WORKDIR=$(cd "$(dirname "$0")/../"; pwd)
+kubectl apply -f $WORKDIR/gang-coordinator.yaml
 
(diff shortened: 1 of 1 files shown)


kingamarton (changes requested):
Thank you @HuangTing-Yao for fixing this one. 

Please change export use the yaml file for the gang jobs as well. 
After that please check the file with a shell checker. It points some bugs as well. I am using the following shell checker: https://www.shellcheck.net/

HuangTing-Yao:
Thanks for your review @kingamarton 
Now we run gang jobs with some params, like job number or how many pods in one job.
If we export gang jobs to yaml file, it can't dynamic change params when we run this shell script. 
At that situation, we need to apply jobs one by one.

deployments/examples/gang/cmd/gangDeploy.sh
```diff
@@ -26,46 +26,22 @@ set -o pipefail
 JOBAMOUNT=$1
 GANGMEMBER=$2
 RUNTIMESEC=$3
+TIMEOUT=0
+
+# create service and job counter web server
+WORKDIR=$(cd "$(dirname "$0")"; pwd)
+kubectl apply -f $WORKDIR/../gang-coordinator.yaml
```
yangwwei, line 33: please add a check here in case the `gang-coordinator.yaml` is not in the right place.
we can explicitly fail with an error: "gang-coordinator.yaml is not found in path xxxx"

deployments/examples/gang/cmd/gangDeploy.sh
```diff
@@ -26,46 +26,22 @@ set -o pipefail
...
-      ports:
-        - containerPort: 8863
-EOF)
 # wait for web server to be running
-until grep 'Running' <(kubectl get pod gangweb -o=jsonpath='{.status.phase}'); do
+until grep -q 'Running' <(kubectl get pod gangweb -o=jsonpath='{.status.phase}'); do
   sleep 1
+  TIMEOUT=$(($TIMEOUT+1))
+  if [ $TIMEOUT -ge 20 ]; then
+    echo "Timeout for waiting web server"
```
yangwwei, line 40: before exit, can we delete the resources it created earlier?
```
kubectl delete -f $WORKDIR/../gang-coordinator.yaml
```

yangwwei (changes requested):

deployments/examples/gang/cmd/gangDeploy.sh
HuangTing-Yao: sure, done.

deployments/examples/gang/cmd/gangDeploy.sh
HuangTing-Yao: I had check several times, it can work in my local, do I miss something?

deployments/examples/gang/cmd/gangDeploy.sh
```diff
@@ -26,46 +26,23 @@ set -o pipefail
 JOBAMOUNT=$1
 GANGMEMBER=$2
 RUNTIMESEC=$3
```
wilfred-s, line 28: We really need to check and fail if there are not enough parameters.
We also should check if the parameters are integers with a simple check:
```
if [ "$1" -eq "$1" ] 2>/dev/null
then
    JOBAMOUNT=$1
else    
    echo "ERROR: first parameter must be an integer."
    exit 1
fi
```

wilfred-s (changes requested):
Fix the missing space on the last line of the license:
```
#limitations under the License.
```

deployments/examples/gang/cmd/gangDeploy.sh
```diff
@@ -96,5 +73,6 @@ spec:
           value: "$RUNTIMESEC"
       restartPolicy: Never
       schedulerName: yunikorn
-EOF)
+EOF
+)
 done
```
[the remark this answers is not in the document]
HuangTing-Yao: before the `done`, right?

deployments/examples/gang/cmd/gangDeploy.sh
HuangTing-Yao: Sure, done.

deployments/examples/gang/cmd/gangDeploy.sh
yangwwei: I mean to check the file existence in the code, e.g

```
if [[ -f $WORKDIR/../gang-coordinator.yaml ]]; then
  // do something
else
  // print the error message and return
fi
```

yangwwei:
Looks like the e2e test failure is due to a helm chart upgrade, from 3.5.1 to 3.5.2

deployments/examples/gang/cmd/gangDeploy.sh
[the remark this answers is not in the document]
wilfred-s: After the done. The check for "no newline at the end of file" is always for the last line of the file. It expects a simple empty line at the end of the file.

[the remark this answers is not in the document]
HuangTing-Yao: OK, done, but we can't put the `)` after `EOF` in same line.
It will show some warning, need to put it in next line.

wilfred-s (approved):
looks good now

yangwwei (approved):

PR #230, opened 2021-02-04
[YUNIKORN-538] Scheduler is unable to recovery from a restart.

yangwwei:


wilfred-s (changes requested):
This test uses 3 different ways to test for a nil error value:
1. assert.Assert(t, err == nil)
2. assert.Equal(t, err, nil)
3. assert.NilError(t, err,...)

Should be using `assert.NilError()` consistently in the test.

yangwwei:
> This test uses 3 different ways to test for a nil error value:
> 
> 1. assert.Assert(t, err == nil)
> 2. assert.Equal(t, err, nil)
> 3. assert.NilError(t, err,...)
> 
> Should be using `assert.NilError()` consistently in the test.

Thanks, fixed in the latest commit.

wilfred-s (approved):
LGTM, committing

PR #231, opened 2021-02-15
[YUNIKORN-460] Handle placeholder timeout

kingamarton:


kingamarton:
I still have to cover the changes with unit tests. Before I start to write them, @wilfred-s , @yangwwei can you please check briefly the patch if you agree with this approach?

pkg/callback/scheduler_callback.go
```diff
@@ -139,6 +150,11 @@ func (callback *AsyncRMCallback) RecvUpdateResponse(response *si.UpdateResponse)
 				log.Logger().Error("failed to delete application", zap.Error(err))
 			}
 		} else {
+			if updated.State == events.States().Application.Killed {
+				//TODO: implement the killed event
+				ev := cache.NewFailApplicationEvent(updated.ApplicationID)
+				dispatcher.Dispatch(ev)
+			}
```
yangwwei, line 157: @kingamarton how can we support a "Kill" application event?
I am not fully understanding what will happen behind this.

pkg/common/constants/constants.go
```diff
@@ -50,11 +50,14 @@ const SchedulerName = "yunikorn"
...
 
 // Gang scheduling
 const PlaceholderContainerImage = "k8s.gcr.io/pause"
 const PlaceholderContainerName = "pause"
 const PlaceholderPodRestartPolicy = "Never"
+const AnnotationPlaceholderFlag = "yunikorn.apache.org/placeholder"
+const AnnotationTaskGroupName = "yunikorn.apache.org/task-group-name"
+const AnnotationTaskGroups = "yunikorn.apache.org/task-groups"
+const AnnotationSchedulingPolicyParam = "yunikorn.apache.org/schedulingPolicyParameters"
+const SchedulingPolicyTimeoutParam = "timeout"
```
yangwwei, line 62: We need to do a second think about the name.
"timeout" is not very descriptive to me and may cause confusion.
Maybe we can explicitly call it "reservationTimeout"

kingamarton: I renamed it to placeholderTimeout

pkg/callback/scheduler_callback.go
kingamarton: As we discussed, we will fail the application, also the placeholders will be removed. Since the app will be failed, it will be skipped from the next scheduling cycles.

pkg/callback/scheduler_callback.go
```diff
@@ -139,6 +150,10 @@ func (callback *AsyncRMCallback) RecvUpdateResponse(response *si.UpdateResponse)
 				log.Logger().Error("failed to delete application", zap.Error(err))
 			}
 		} else {
+			if updated.State == events.States().Application.Killed {
+				ev := cache.NewFailApplicationEvent(updated.ApplicationID)
+				dispatcher.Dispatch(ev)
+			}
```
yangwwei, line 156: When we fail an application, we need to expose some pod level events to indicate this issue.
we can do this in a follow up JIRA.

pkg/common/constants/constants.go
```diff
@@ -50,11 +50,14 @@ const SchedulerName = "yunikorn"
...
 
 // Gang scheduling
 const PlaceholderContainerImage = "k8s.gcr.io/pause"
 const PlaceholderContainerName = "pause"
 const PlaceholderPodRestartPolicy = "Never"
+const AnnotationPlaceholderFlag = "yunikorn.apache.org/placeholder"
+const AnnotationTaskGroupName = "yunikorn.apache.org/task-group-name"
+const AnnotationTaskGroups = "yunikorn.apache.org/task-groups"
+const AnnotationSchedulingPolicyParam = "yunikorn.apache.org/schedulingPolicyParameters"
+const SchedulingPolicyTimeoutParam = "placeholderTimeout"
```
yangwwei, line 62: Looks like it implies the unit for `placeholderTimeout` is seconds, I think we should have this declared explicitly in the parameter, otherwise, we will need additional docs to explain the format. Suggest to rename this to `placeholderTimeoutInSeconds`

yangwwei:
Overall the changes looked good, +1.
For the remaining review comments, I have created several follow up JIRAs under https://issues.apache.org/jira/browse/YUNIKORN-553.

yangwwei (approved):

PR #232, opened 2021-02-19
[YUNIKORN-521] Placeholder pods are not cleaned when the job is deleted

kingamarton:
Filled  ownreference field for the placeholder pods


pkg/appmgmt/general/general.go
```diff
@@ -130,6 +130,7 @@ func (os *Manager) getAppMetadata(pod *v1.Pod) (interfaces.ApplicationMetadata,
 		User:          user,
 		Tags:          tags,
 		TaskGroups:    taskGroups,
+		PlaceholderOwnerReferences: pod.OwnerReferences,
```
yangwwei, line 133: I think we can just call this owner reference since this is a general K8s concept.
It doesn't mean only placeholder pod has that.

yangwwei (changes requested):
Can we add some UT in `general_test.go` ? that needs to cover the case where the pod has ownerReference and hasn't.

pkg/cache/application.go
```diff
@@ -208,6 +210,12 @@ func (app *Application) getTaskGroups() []v1alpha1.TaskGroup {
 	return app.taskGroups
 }
 
+func (app *Application) setOwnReference(ref []metav1.OwnerReference) {
```
wilfred-s, line 213: nit: can we name this `setOwnerReference()`

pkg/cache/placeholder.go
```diff
@@ -36,6 +36,11 @@ type Placeholder struct {
 }
 
 func newPlaceholder(placeholderName string, app *Application, taskGroup v1alpha1.TaskGroup) *Placeholder {
+	ownerRef := app.placeholderOwnerReferences
```
wilfred-s, line 39: nit: this is a list of references, should call this `ownerRefs` plural

pkg/cache/placeholder_manager_test.go
```diff
@@ -36,12 +36,69 @@ import (
...
+		Controller: &controller,
+	}
+	app.setOwnReference([]apis.OwnerReference{ownRef})
+	mockedAPIProvider := client.NewMockedAPIProvider()
+	pods := createAndCheckPlaceholderCreate(mockedAPIProvider, app, t)
+	for _, pod := range pods {
+		assert.Assert(t, len(pod.OwnerReferences) == 1, "The pod should have exactly one owner reference set")
+		assert.Assert(t, *pod.OwnerReferences[0].Controller == false, "The owner reference should not be a controller")
+		assert.Assert(t, pod.OwnerReferences[0].Name == ownRef.Name, "The owner reference name does not match")
+		assert.Assert(t, pod.OwnerReferences[0].UID == ownRef.UID, "The owner reference UID does not match")
```
wilfred-s, line 97: nit should we not use `assert.Equals()` all comparisons are simple values?

wilfred-s (commented):
Couple of small points more about style than code content.

pkg/cache/placeholder.go
```diff
@@ -46,6 +46,11 @@ type Placeholder struct {
 }
 
 func newPlaceholder(placeholderName string, app *Application, taskGroup v1alpha1.TaskGroup) *Placeholder {
+	ownerRefs := app.placeholderOwnerReferences
+	controller := false
+	for _, r := range ownerRefs {
+		*r.Controller = controller
+	}
```
yangwwei, line 53: we are setting the controller field to false all the time.
could you pls add some more comments why we are doing this?

pkg/appmgmt/general/general.go
```diff
@@ -123,20 +126,40 @@ func (os *Manager) getAppMetadata(pod *v1.Pod) (interfaces.ApplicationMetadata,
...
+		QueueName:               utils.GetQueueNameFromPod(pod),
+		User:                    user,
+		Tags:                    tags,
+		TaskGroups:              taskGroups,
 		PlaceholderTimeoutInSec: placeholderTimeout,
+		OwnerReferences:         ownerReferences,
 	}, true
 }
 
+func getOwnerReferences(pod *v1.Pod) []metav1.OwnerReference {
```
yangwwei, line 146: pls add comments to this function, to explain how we decide what owner-reference will be set for the placeholders.

yangwwei (approved):
LGTM

PR #233, opened 2021-02-20
[YUNIKORN-511] Support to run e2e tests on local macbook

morpheyesh:
Added support for Mac and ginkgo tests runs fine.


scripts/run-e2e-tests.sh
```diff
@@ -75,18 +75,35 @@ function install_cluster() {
...
+
+  if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+    echo "Installing KIND for Linux.."
+         curl -Lo ./kind "https://kind.sigs.k8s.io/dl/v0.8.0/kind-linux-amd64" \
+            && chmod +x ./kind && mv ./kind $(go env GOPATH)/bin
+         exit_on_error "install KIND failed"
+         check_cmd "kind"
+  elif [[ "$OSTYPE" == "darwin"* ]]; then
+    echo "Installing KIND for Mac.."
+       curl -Lo ./kind "https://kind.sigs.k8s.io/dl/v0.10.0/kind-darwin-amd64" \
```
yangwwei, line 104: can we install same kind version for linux/mac?

scripts/run-e2e-tests.sh
```diff
@@ -75,18 +75,35 @@ function install_cluster() {
...
+          && chmod +x ./kind && mv ./kind $(go env GOPATH)/bin
```
yangwwei, line 105: similar, I think we need to add "exit_on_error" and "check_cmd".
also pls add an "else" in case OSTYPE is something else

yangwwei (commented):
Thanks for working on this. Please see my comments.
And one more thing, what's the reason for updating the go.mod? Is that needed for this patch?
Thanks

scripts/run-e2e-tests.sh
```diff
@@ -75,18 +75,35 @@ function install_cluster() {
...
+    echo "Installing Kubectl for Linux.."
+          stable_release=$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)
+          exit_on_error "unable to retrieve latest stable version of kubectl"
+          curl -LO https://storage.googleapis.com/kubernetes-release/release/${stable_release}/bin/linux/amd64/kubectl \
+          && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
+          exit_on_error "install kubectl failed"
+  elif [[ "$OSTYPE" == "darwin"* ]]; then
+    echo "Installing Kubectl for Mac.."
+              curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/darwin/amd64/kubectl" \
+              && chmod +x kubectl && sudo mv kubectl /usr/local/bin
```
wilfred-s, line 89: We need to pull the version from the same location. We should not mix `dl.k8s.io` and `googleapis.com`
The full procedure is the same for both installs except for a small part of the URL. Instead of having two copied procedures we should have one and just replace `darwin` and `linux` when building the URL.

scripts/run-e2e-tests.sh
wilfred-s, line 104: We need to use the same version of kind on all platforms.
Same comments as above for the URL.

wilfred-s (changes requested):
This should be a shell script change only. No `go.mod` or `go.sum` file changes.

As mentioned in the jira also. On the local machine `helm`, `kubectl` and `kind` might already be installed. We should not pull down and update to the latest version without checking. You could break the install a developer already has on his machine.
We should only perform the install if the tools are not yet available.

Last point: the directory `$HOME/gopath/bin` does not exist on all systems. You cannot expect it to be there.  `go get` will use a valid GOPATH, If $GOPATH is not specified, $HOME/go will be used. Which means that you also need to fix the `go get` for ginkgo and gomega

wilfred-s:
@morpheyesh I looked for a user with a similar name in the apache jira to assign [YUNIKORN-511](https://issues.apache.org/jira/browse/YUNIKORN-511) to you. I did not find you. We track our issues in jira when you create a ID in jira I can add you to the contributors list and assign the jira to you.

scripts/run-e2e-tests.sh
```diff
@@ -75,18 +75,35 @@ function install_cluster() {
 
   # install kubectl
   echo "step 3/6: installing kubectl"
-  stable_release=$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)
-  exit_on_error "unable to retrieve latest stable version of kubectl"
-  curl -LO https://storage.googleapis.com/kubernetes-release/release/${stable_release}/bin/linux/amd64/kubectl \
-    && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
-  exit_on_error "install kubectl failed"
+
+  if [[ "$OSTYPE" == "linux-gnu"* ]]; then
```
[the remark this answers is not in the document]
morpheyesh: It is fixed now.

scripts/run-e2e-tests.sh
```diff
@@ -75,18 +75,35 @@ function install_cluster() {
...
+          stable_release=$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)
+          exit_on_error "unable to retrieve latest stable version of kubectl"
+          curl -LO https://storage.googleapis.com/kubernetes-release/release/${stable_release}/bin/linux/amd64/kubectl \
+          && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
+          exit_on_error "install kubectl failed"
+  elif [[ "$OSTYPE" == "darwin"* ]]; then
+    echo "Installing Kubectl for Mac.."
+              curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/darwin/amd64/kubectl" \
+              && chmod +x kubectl && sudo mv kubectl /usr/local/bin
+  fi
```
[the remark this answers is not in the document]
morpheyesh: Let me know if this looks good.

scripts/run-e2e-tests.sh
morpheyesh: Yep. 0.10.0 for both.

morpheyesh:
@wilfred-s Please have a look now. I made some changes to the script but I am not entirely sure how you want `ginkgo` and `gomega` managed. I believe if $GOPATH exists, `go get` should install it now.

scripts/run-e2e-tests.sh
```diff
@@ -69,24 +124,32 @@ function install_cluster() {
...
   check_cmd "curl"
-  curl -L https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash
-  exit_on_error "install helm-v3 failed"
-  check_cmd "helm"
+  if ! command -v helm &> /dev/null
+  then
+      install_helm
+  else
+    echo "Helm already installed"
+  fi
```
yangwwei, line 132: just a small suggestion, the check can be added in the `install_helm` function, which makes the main flow simpler.
same as the rest install functions.

yangwwei (commented):
Just added a minor comment, other than that looks good to me

scripts/run-e2e-tests.sh
morpheyesh: Good point. Looks neater now :)

yangwwei (approved):
LGTM. @wilfred-s if all your comments have been addressed so far? pls take a look

yangwwei:
Re-running the jobs as the UT failure seems to be an intermittent issue.

scripts/run-e2e-tests.sh
```diff
@@ -55,11 +124,12 @@ function install_cluster() {
...
 
-  # install ginkgo and gomega for e2e tests
+  # Check if go is installed.
   check_cmd "go"
+
+  # install ginkgo and gomega for e2e tests
   go get -v github.com/onsi/ginkgo/ginkgo
   go get -v github.com/onsi/gomega
-  export PATH=$PATH:$HOME/gopath/bin
   check_cmd "ginkgo"
```
wilfred-s, line 133: I have one minor thing left: should we add a comment that the install for ginkgo can be found in the directory: ``echo `go env GOPATH`/bin``

morpheyesh: Should be good now. :)

wilfred-s (approved):
LGTM, committing

PR #234, opened 2021-02-23
YUNIKORN-520: Gang scheduling waits indefinitely for placeholder pod …

manirajv06:
…allocation even where is no quota left in the queue


wilfred-s (approved):
Just a small one: there is a typo in the function name, missing the **r** in getPlaceholderAsk()
Beside that change looks good. I'll fix the typo on commit

PR #235, opened 2021-02-25
[YUNIKORN-549] Scheduler recovery failure occasionally while recovering a large number of applications.

yangwwei:
The fix removes the flag `AddApplicationRequest#ApplicationMetadata#Recovery`, the recovery process no longer depends on which pod is seen by the shim to determine if the app needs recovery or not. Instead, it calls the K8s lister to list all apps that needs to be recovered, and then trigger recovery for them.


wilfred-s (approved):
change LGTM, discussed this offline.
testing is a problem as the order of pods is not something that we can easily influence.

PR #236, opened 2021-03-01
[YUNIKORN-550] placeholderAsk scaling

wilfred-s:
Resources inside the shim are scaled before sending to the core. The
placeholderAsk was not scaled before sending to the core causing limit
checks to fail.
Moving the conversion into the common code to line up with other types.


wilfred-s:
The `TestSetTaskGroupsAndSchedulingPolicy()` covers the new code. Not sure why codecov is not picking that up.

yangwwei (approved):
hi @wilfred-s thanks for the quick fix. +1

PR #238, opened 2021-03-10
[YUNIKORN-559] Placeholder pods must be running as a non-root user.

yangwwei:


pkg/cache/placeholder.go
```diff
@@ -29,6 +29,10 @@ import (
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/utils"
 )
 
+// MUST: run the placeholder pod as non-root user
+var runAsUser int64 = 1000
+var runAsGroup int64 = 3000
```
kingamarton, line 34: What these values mens? I think we should add some comments to explain why this values are used

pkg/cache/placeholder.go
```diff
@@ -50,6 +54,10 @@ func newPlaceholder(placeholderName string, app *Application, taskGroup v1alpha1
 			},
 		},
 		Spec: v1.PodSpec{
+			SecurityContext: &v1.PodSecurityContext{
+				RunAsUser:  &runAsUser,
+				RunAsGroup: &runAsGroup,
```
kingamarton, line 59: What is the setting for the firs pod? Wouldn't be better to use the same setting instead of defining some default user and group?

pkg/cache/placeholder.go
wilfred-s: The IDs are just random user and group IDs. The IDs do not have to map to real users in the password file etc. Probably better if they do not. IDs 1000 and above are used by OS tools like `useradd` when adding end users. Anything below that is considered a system account. The linux spec defines a smaller range from 0-499 for the OS, however 1000 is a safer base.
The only thing I would do is keep user and group ID the same: both 1000. The comment seems clear enough.

yangwwei: I thought I have replied but looks like I forgot to hit the send button.
I have added the comments, hopefully that explains things.
The reason to just put 1000/3000 is to follow the recommendation from the K8s security context doc, I think that should be fine.

kingamarton (approved):

PR #239, opened 2021-03-11
[YUNIKORN-562] Support add labels/annotations to taskGroup

yangwwei:
sometimes we need to pass in the labels/annotations to the placeholder pods in order to make them more "like" the real pods. This helps certain cases, for example, if some namespace has sidecar container auto-injection enabled, we can leverage the annotation field to enable or disable the injection, just to comply with the real pod definition.

This PR adds labels/annotations to the taskGroup definition and makes it configurable, so the client can pass in the required values if needed.


kingamarton (commented):
Overall the change looks good, but I have one question: why we are not simply using the labels and annotations from the first real pod instead of asking the user to pass them in the taskGroup?

go.sum
```diff
@@ -829,6 +829,7 @@ golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7/go.mod h1:I/5z698sn9Ka8T
 golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
 golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543 h1:E7g+9GITq07hpfrRu66IVDexMakfv52eLZ2CXBWiKr4=
 golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
+golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 h1:go1bK/D/BFZV2I8cIQd1NKEZ+0owSTG1fDTci4IqFcE=
```
wilfred-s, line 832: We should not update the go.sum in this change.

pkg/cache/placeholder.go
```diff
@@ -73,6 +73,22 @@ func newPlaceholder(placeholderName string, app *Application, taskGroup v1alpha1
 	}
 }
 
+// merge two string maps
+// if the same key defined in the first and second maps
+// the value will be set by the second map
+func mergeMaps(first, second map[string]string) map[string]string {
+	// add default labels first
```
wilfred-s, line 80: This comment does not make sense: we're walking over a list of keys that could be anything, nothing to do with a label. Same for the comment below referencing taskgroup

pkg/cache/placeholder.go
```diff
@@ -73,6 +73,22 @@ func newPlaceholder(placeholderName string, app *Application, taskGroup v1alpha1
 	}
 }
 
+// merge two string maps
+// if the same key defined in the first and second maps
+// the value will be set by the second map
```
wilfred-s, line 78: This function should be defined as part of the `common.utils`. Easier for re-use.

pkg/cache/placeholder.go
```diff
@@ -40,14 +40,14 @@ func newPlaceholder(placeholderName string, app *Application, taskGroup v1alpha1
...
 				constants.LabelApplicationID: app.GetApplicationID(),
 				constants.LabelQueueName:     app.GetQueue(),
-			},
-			Annotations: map[string]string{
+			}, taskGroup.Labels),
+			Annotations: mergeMaps(map[string]string{
 				constants.AnnotationPlaceholderFlag: "true",
 				constants.AnnotationTaskGroupName:   taskGroup.Name,
-			},
+			}, taskGroup.Annotations),
```
wilfred-s, line 50: I am not sure if we should override these values with the values from the annotations. We really need to make sure that these are set as we need and not allow overriding.

pkg/cache/placeholder.go
```diff
@@ -40,14 +40,14 @@ func newPlaceholder(placeholderName string, app *Application, taskGroup v1alpha1
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      placeholderName,
 			Namespace: app.tags[constants.AppTagNamespace],
-			Labels: map[string]string{
+			Labels: mergeMaps(map[string]string{
 				constants.LabelApplicationID: app.GetApplicationID(),
 				constants.LabelQueueName:     app.GetQueue(),
-			},
-			Annotations: map[string]string{
+			}, taskGroup.Labels),
```
wilfred-s, line 46: The labels we set should not be overwritten by what the end user provides. We need it to be set to the correct app etc without allowing an override

pkg/cache/placeholder_test.go
```diff
@@ -68,6 +68,82 @@ func TestNewPlaceholder(t *testing.T) {
...
+	assert.Equal(t, holder.pod.Labels["labelKey1"], "labelKeyValue1")
+	assert.Equal(t, holder.pod.Annotations["annotationKey0"], "annotationValue0")
+	assert.Equal(t, holder.pod.Annotations["annotationKey1"], "annotationValue1")
+	assert.Equal(t, holder.pod.Annotations["annotationKey2"], "annotationValue2")
+}
+
+func TestMergeMaps(t *testing.T) {
+	result := mergeMaps(nil, nil)
+	assert.Assert(t, result != nil)
+	assert.Equal(t, len(result), 0)
```
wilfred-s, line 113: two nil maps as input should return a nil map as output.

wilfred-s (changes requested):
The merge as it is now could cause issues as the end user could override the flags and values we need to be set.
We also have a lint failure in the build

wilfred-s:
> Overall the change looks good, but I have one question: why we are not simply using the labels and annotations from the first real pod instead of asking the user to pass them in the taskGroup?

There could be a difference between the annotations for different taskgroups. We saw a sidecar injected on placeholders for one task group's placeholders which were blocked by an annotation on the real allocation. It only happened for 1 of the 2 task groups defined for the application.

go.sum
yangwwei: Reverted

pkg/cache/placeholder.go
yangwwei: That was leftover when I renaming the function, removed them.

pkg/cache/placeholder.go
yangwwei: Moved to util package for re-use

pkg/cache/placeholder.go
yangwwei: good point, these values are needed, we should not allow overwrite. 
I've switched around the argument to make sure they are always set.

pkg/cache/placeholder.go
yangwwei: Fixed

pkg/cache/placeholder_test.go
yangwwei: Fixed

yangwwei:
Thanks for the quick review, @wilfred-s , @kingamarton pls check the latest changes.

pkg/common/utils/utils_test.go
```diff
@@ -331,3 +331,39 @@ func TestGetApplicationIDFromPod(t *testing.T) {
...
+		"b": "b2",
+		"d": "d2",
+	})
+	assert.Assert(t, result != nil)
+	assert.Equal(t, len(result), 4)
+	assert.Equal(t, result["a"], "a2")
+	assert.Equal(t, result["b"], "b2")
+	assert.Equal(t, result["c"], "c1")
+	assert.Equal(t, result["d"], "d2")
+}
```
kingamarton, line 369: nit: if it is possible in go it is suggested to write table driven tests. It is more clear and elegant.
```
testCases := []struct {
		name string
		map1       map[string]string
		map2 map[string]string
		expectedResult map[string]string
	}{
		{"Nil maps", nil, nil, nil},
                  ...... <create here the other cases>
	}
	for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			assert.DeepEqual(t, MergeMaps(tc.map1, tc.map2), tc.expectedResult)
		})
	}
```

pkg/cache/placeholder_test.go
```diff
@@ -68,6 +68,45 @@ func TestNewPlaceholder(t *testing.T) {
...
+	})
+
+	holder := newPlaceholder("ph-name", app, app.taskGroups[0])
+	assert.Equal(t, len(holder.pod.Labels), 4)
+	assert.Equal(t, len(holder.pod.Annotations), 5)
+	assert.Equal(t, holder.pod.Labels["labelKey0"], "labelKeyValue0")
+	assert.Equal(t, holder.pod.Labels["labelKey1"], "labelKeyValue1")
+	assert.Equal(t, holder.pod.Annotations["annotationKey0"], "annotationValue0")
+	assert.Equal(t, holder.pod.Annotations["annotationKey1"], "annotationValue1")
+	assert.Equal(t, holder.pod.Annotations["annotationKey2"], "annotationValue2")
```
kingamarton, line 107: nit: we can compare two maps in one line using assert.DeepEqual. With that we can make the comparison in 2 lines instead of 7

PR #240, opened 2021-03-16
[YUNIKORN-333]Reduce the number events published to K8s event system when predicate fails

HuangTing-Yao:
Add rate limit to the area which publish k8s event.
Now it will publish predicate fails message once per sec.

### DIFF
--- a/pkg/plugin/predicates/predictor.go
+++ b/pkg/plugin/predicates/predictor.go
@@ -22,8 +22,10 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"time"
 
 	"go.uber.org/zap"
+	"golang.org/x/time/rate"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/plugin/predicates/predictor.go
```diff
@@ -278,8 +283,10 @@ func (p *Predictor) predicatesAllocate(pod *v1.Pod, meta predicates.PredicateMet
 			}
 
 			if !fit {
-				events.GetRecorder().Eventf(pod, v1.EventTypeWarning,
-					"FailedScheduling", "%v", reasons)
+				if isAllowed := limiter.AllowN(time.Now(), 1); isAllowed {
```
wilfred-s, line 286: This can be simplified to just become `limiter.Allow()`

HuangTing-Yao: OK, Sure.

wilfred-s (approved):
LGTM

PR #243, opened 2021-03-17
[YUNIKORN-571]Add a label to the placeholder pods the selector

HuangTing-Yao:
Rename `AnnotationPlaceholderFlag` to `PlaceholderFlag`, and add same label when placeholder created.


pkg/common/constants/constants.go
```diff
@@ -55,7 +55,7 @@ const AppManagerHandlerName = "yunikorn-app"
 const PlaceholderContainerImage = "k8s.gcr.io/pause"
 const PlaceholderContainerName = "pause"
 const PlaceholderPodRestartPolicy = "Never"
-const AnnotationPlaceholderFlag = "yunikorn.apache.org/placeholder"
+const PlaceholderFlag = "yunikorn.apache.org/placeholder"
```
wilfred-s, line 58: For the `ApplicationID` we use an unqualified `applicationId` in labels and a qualified version `yunikorn.apache.org/app-id` in the annotation. Shouldn't we follow the same rule here? I can live with either solution, I just think we need to set a standard for what we do.

pkg/common/utils/gang_utils.go
```diff
@@ -75,7 +75,7 @@ func GetPlaceholderResourceRequest(resources map[string]resource.Quantity) v1.Re
 }
 
 func GetPlaceholderFlagFromPodSpec(pod *v1.Pod) bool {
-	if value, ok := pod.Annotations[constants.AnnotationPlaceholderFlag]; ok {
+	if value, ok := pod.Annotations[constants.PlaceholderFlag]; ok {
```
wilfred-s, line 78: Do we add the label just as a convenience or is there more behind this?
The appID retrieval uses a logic to look at annotations and then labels (normal followed by spark).

wilfred-s (changes requested):
Two points on consistency more than on the code itself.
We need to be clear and consistent in what we do.

pkg/common/constants/constants.go
HuangTing-Yao: Sure, follow the some rule will more clear to read.

pkg/common/utils/gang_utils.go
HuangTing-Yao: I think we can follow the same logic in `GetApplicationIDFromPod()`, add the label part in `GetPlaceholderFlagFromPodSpec()`

wilfred-s (approved):
LGTM, please feel free to commit

PR #244, opened 2021-03-17
[YUNIKORN-579] Handle release allocation/ask events under Failing state

kingamarton:


pkg/callback/scheduler_callback.go
```diff
@@ -88,6 +88,27 @@ func (callback *AsyncRMCallback) RecvUpdateResponse(response *si.UpdateResponse)
...
+			zap.String("appId", updated.ApplicationID),
+			zap.String("new status", updated.State))
+		// delete application from context
+		if updated.State == events.States().Application.Completed {
+			err := callback.context.RemoveApplicationInternal(updated.ApplicationID)
+			if err != nil {
+				log.Logger().Error("failed to delete application", zap.Error(err))
+			}
+		} else {
+			if updated.State == "Failing" || updated.State == events.States().Application.Failed {
```
yangwwei, line 103: do we need to handle the Failing state in the shim side? If I understand correctly, when the core side puts an app into Failing, it will no longer schedule any of its asks. Why we need this change?

pkg/callback/scheduler_callback.go
```diff
@@ -88,6 +88,27 @@ func (callback *AsyncRMCallback) RecvUpdateResponse(response *si.UpdateResponse)
...
+				ev := cache.NewFailApplicationEvent(updated.ApplicationID)
+				dispatcher.Dispatch(ev)
+			}
+			// handle status update
+			dispatcher.Dispatch(cache.NewApplicationStatusChangeEvent(updated.ApplicationID, events.AppStateChange, updated.State))
+		}
+	}
+
```
yangwwei, line 111: I do not think this order matters, can we move it back?

pkg/callback/scheduler_callback.go
kingamarton: Yes, we need this change. Without this check the shim will continue to send the real pods for scheduling. Even if we would handle it correctly in the core side, it is wrong. We should not send any new requests when the app is failing or failed. Also the app in the shim and core side should be in sync. We should not keep the app running in the shim side while it is already Failed in the core side. 
For now I think we can keep this string check for the "Failing" state, but I opened a Jira for introducing a new state in the shim side as well: YUNIKORN-585

pkg/callback/scheduler_callback.go
kingamarton: moved it back

pkg/callback/scheduler_callback.go
yangwwei: Got it, but I think we need to introduce the Failing state in the shim side as well to keep it in sync with the core side state. We need to follow that up in YUNIKORN-585. It doesn't seem good to handle ReleaseAppAllocation/ask event in Failed state, which is quite weird. Lets get this fixed in YUNIKORN-585. thanks

yangwwei (approved):
+1

PR #245, opened 2021-03-18
[YUNIKORN-556]Expose pod level events when an app is failed in scheduling

HuangTing-Yao:


pkg/cache/application.go
```diff
@@ -507,6 +514,19 @@ func (app *Application) handleCompleteApplicationEvent(event *fsm.Event) {
 	// TODO app lifecycle updates
 }
 
+func (app *Application) handleFailApplicationEvent(event *fsm.Event) {
+	// unallocated task states include New, Pending and Scheduling
+	unalloc := app.GetNewTasks()
+	unalloc = append(unalloc, app.GetPendingTasks()...)
+	unalloc = append(unalloc, app.GetSchedulingTask()...)
```
kingamarton, line 521: You have a deadlock here, because the application is already locked: https://github.com/apache/incubator-yunikorn-k8shim/blob/master/pkg/cache/application.go#L153-L155

kingamarton: You can simply use `unalloc := app.getTasks(events.States().Task.New)` ...

kingamarton (changes requested):

pkg/cache/application.go
HuangTing-Yao: Thanks. Done.

pkg/cache/application.go
```diff
@@ -507,6 +508,19 @@ func (app *Application) handleCompleteApplicationEvent(event *fsm.Event) {
...
 }
 
+func (app *Application) handleFailApplicationEvent(event *fsm.Event) {
+	// unallocated task states include New, Pending and Scheduling
+	unalloc := app.getTasks(events.States().Task.New)
+	unalloc = append(unalloc, app.getTasks(events.States().Task.Pending)...)
+	unalloc = append(unalloc, app.getTasks(events.States().Task.Scheduling)...)
+	// publish pod level event to unallocated pods
+	for _, task := range unalloc {
+		events.GetRecorder().Eventf(task.GetTaskPod(), v1.EventTypeWarning, "ApplicationFailed",
```
yangwwei, line 518: We need to print the message about this failed app, otherwise, users will be confused about what happened.
This can get via `si.UpdatedApplication.Message`, so in the `scheduler_callback.go`, we need to add that in the FailApplication event as well.

yangwwei (commented):
To test this changes, pls use gang scheduling app. 
YK is not failing an app in many cases, one simple case I can image is to create 2 taskGroups, 1  taskGroup gives some invalid node selector, set the placeholderTimeoutInSeconds with a small value, e.g 1 minute. Then YK should transit the app to Failed after 1 minute.

pkg/cache/application.go
HuangTing-Yao: OK, done

pkg/cache/application.go
```diff
@@ -506,13 +507,33 @@ func (app *Application) onReservationStateChange(event *fsm.Event) {
...
+			"Application %s scheduling failed, reason: %s", app.applicationID, errMess)
+	}
+	log.Logger().Warn("failed to schedule app", zap.String("appID", app.applicationID))
```
yangwwei, line 534: I think we have already logged this, pls remove this line.

pkg/cache/application_test.go
```diff
@@ -177,6 +263,32 @@ func (ms *mockSchedulerAPI) ReloadConfiguration(rmID string) error {
 	return nil
 }
 
+type mockRecorder struct {
```
yangwwei, line 266: Pls rename this struct to `MockedRecorder` and move it to a separate file and put it into https://github.com/apache/incubator-yunikorn-k8shim/tree/master/pkg/common/events package. So in the future we can easily reuse it.

pkg/cache/application_test.go
```diff
@@ -177,6 +263,32 @@ func (ms *mockSchedulerAPI) ReloadConfiguration(rmID string) error {
...
+	updateFn func()
```
yangwwei, line 267: pls rename updateFn to `OnEventf()`, which means this is to be called when `Eventf()` is called

pkg/common/events/recorder.go
```diff
@@ -31,7 +31,7 @@ import (
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/conf"
 )
 
-var eventRecorder record.EventRecorder = record.NewFakeRecorder(1024)
+var EventRecorder record.EventRecorder = record.NewFakeRecorder(1024)
```
yangwwei, line 34: we do not need to expose this if we put the mocked class to the same package, we just need to expose a function to the callers so we can modify the OnXXX functions.

yangwwei (changes requested):

pkg/common/events/recorder.go
HuangTing-Yao: got it.

pkg/cache/application_test.go
HuangTing-Yao: sure

pkg/cache/application_test.go
HuangTing-Yao: OK, done.

pkg/cache/application.go
HuangTing-Yao: OK.

yangwwei (approved):
LGTM

PR #246, opened 2021-03-19
[YUNIKORN-584] App recovery is skipped when applicationID is not set in pods' label.

yangwwei:
### The problem

Credits to Chaoran who finds this bug and figured out the root cause.
The symptom of this issue is described in this issue: https://issues.apache.org/jira/browse/YUNIKORN-584.
The problem is when we list applications, we have a filter that checks label `applicationId` exists in pods, pods without this label will be excluded in this scan because [these lines.](https://github.com/apache/incubator-yunikorn-k8shim/blob/fd69f44d0028d09918715cd222775901400f936d/pkg/appmgmt/general/general.go#L297-L302). When we recover Spark jobs, spark pods do not necessarily have this label, we can recognize them using spark-app-selector. This caused the job to get excluded from the recovery app list, and then subsequent node recovery also failed because it finds the existing allocation for the job but the job was not recovered in the scheduler. 

### The Fix

The fix is to simply remove that few lines. We do not need to filter pods by applicationID label. Instead, we should just scan all the pods, because the next step is to filter apps by app metadata. Lots of code added was for the UT coverage.


pkg/appmgmt/general/general_test.go
```diff
@@ -526,3 +527,269 @@ func TestGetOwnerReferences(t *testing.T) {
...
+			Labels: map[string]string{
+				constants.LabelApplicationID: app01,
+				constants.LabelQueueName:     queue01,
+			},
+		},
+		Spec: v1.PodSpec{
+			SchedulerName: constants.SchedulerName,
+			NodeName:      "allocated-node",
+		},
+	})
```
wilfred-s, line 583: This is a duplicate from app01pod00001 and does not test anything new, should be removed.

pkg/appmgmt/general/general_test.go
```diff
@@ -526,3 +527,269 @@ func TestGetOwnerReferences(t *testing.T) {
...
+			Namespace: ns02,
+			Labels: map[string]string{
+				constants.LabelApplicationID: app02,
+				constants.LabelQueueName:     queue02,
+			},
+		},
+		Spec: v1.PodSpec{
+			SchedulerName: constants.SchedulerName,
+		},
+	})
```
wilfred-s, line 622: This is a duplicate from app01pod00003 and does not test anything new, should be removed.

pkg/appmgmt/general/general_test.go
```diff
@@ -526,3 +527,269 @@ func TestGetOwnerReferences(t *testing.T) {
...
```
wilfred-s, line 640: This is a duplicate from app01pod00003 and app02pod0001 and does not test anything new, should be removed.

pkg/appmgmt/general/general_test.go
```diff
@@ -526,3 +527,269 @@ func TestGetOwnerReferences(t *testing.T) {
...
+			Namespace: ns02,
+			Labels: map[string]string{
+				constants.SparkLabelAppID: app03,
+			},
+		},
+		Spec: v1.PodSpec{
+			SchedulerName: constants.SchedulerName,
+			NodeName:      "some-node",
+		},
+	})
```
wilfred-s, line 679: This is a duplicate from app03pod0001 and does not test anything new, should be removed.

pkg/appmgmt/general/general_test.go
```diff
@@ -526,3 +527,269 @@ func TestGetOwnerReferences(t *testing.T) {
...
+			Namespace: ns01,
+			Annotations: map[string]string{
+				constants.AnnotationApplicationID: app04,
+			},
+		},
+		Spec: v1.PodSpec{
+			SchedulerName: constants.SchedulerName,
+			NodeName:      "some-node",
+		},
+	})
```
wilfred-s, line 718: This is a duplicate from app04pod0001 and does not test anything new, should be removed.

pkg/appmgmt/general/general_test.go
```diff
@@ -526,3 +527,269 @@ func TestGetOwnerReferences(t *testing.T) {
 	assert.Equal(t, returnedOwnerRefs[0].Kind, "Pod", "Unexpected owner reference Kind")
 	assert.Equal(t, returnedOwnerRefs[0].APIVersion, v1.SchemeGroupVersion.String(), "Unexpected owner reference Kind")
 }
+
+// nolint: funlen
+func TestListApplication(t *testing.T) {
```
wilfred-s, line 532: This should be a table driven test.
We have 4 things that can change: Labels, Annotations, SchedulerName and NodeName. The outcome is known: app exists or not. Setting up a table with those 4 inputs and the expected output. Create a new pod lister or add a reset function on the mocked pod lister and run the checks in a loop.

wilfred-s (changes requested):
Main issue is in the test. A table driven test makes it more readable and makes sure we have covered all the cases that need to be tested.

pkg/appmgmt/general/general_test.go
kingamarton: I agree with @wilfred-s. When possible we should use table driven tests. Is much easier to read, and it will be shorter as well, so we might not need to skip the lint funlen check, if we do, then maybe it worth to split the test instead of just ignoring it.

pkg/appmgmt/general/general.go
```diff
@@ -293,15 +292,8 @@ func (os *Manager) deletePod(obj interface{}) {
 }
 
 func (os *Manager) ListApplications() (map[string]interfaces.ApplicationMetadata, error) {
-	// selector: applicationID exist
-	slt := labels.NewSelector()
-	req, err := labels.NewRequirement(constants.LabelApplicationID, selection.Exists, nil)
-	if err != nil {
-		return nil, err
-	}
-	slt = slt.Add(*req)
```
kingamarton, line 302: Instead of removing the label selector I think we should modify the selector to check for either appID, or spark-app-selector.

kingamarton (changes requested):

yangwwei:
The review comments about the UT could not coverage, it needs some more time to get fully addressed.
To unblock the 0.10 build, I suggest getting this fix in and I have filed a follow-up JIRA: https://issues.apache.org/jira/browse/YUNIKORN-593.

kingamarton (approved):

PR #248, opened 2021-03-25
[YUNIKORN-600] PlaceholderManager structure init

wilfred-s:
Fix the initialisation of the PlaceholderManager structure to init the
channel and orphanPods map (renamed).
Fix the tests to call NewPlaceholderManager to not init the structure in
the test code and thus hiding init issues.

The stop channel is never closed and should be reused. We should not
create a new channel in the Start but in the New call. Fix up the Stop
checks to not use one long 5 second timeout but check every 100ms. This
change speeds up the test (down to 0.2s from 8s) and the shutdown of the
whole shim.

### DIFF
--- a/pkg/cache/placeholder_manager_test.go
+++ b/pkg/cache/placeholder_manager_test.go
@@ -20,8 +20,6 @@ package cache
 
 import (
 	"fmt"
-	"sync"
-	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


yangwwei:
@HuangTing-Yao  can you help to review this change? THX

pkg/cache/placeholder_manager_test.go
```diff
@@ -220,27 +220,44 @@ func TestCleanOrphanPlaceholders(t *testing.T) {
...
+	assert.Equal(t, mgr.isRunning(), true, "manager should be running after start")
+
+	// starting a second time should do nothing
+	mgr.Start()
+	assert.Equal(t, mgr.isRunning(), true, "sending start 2nd time should not do anything")
+
+	// this is a blocking call
+	mgr.Stop()
+	// allow time for processing as the call can return faster than the flag is set
+	time.Sleep(5 * time.Millisecond)
```
HuangTing-Yao, line 244: In `Statt()` it might sleep `100 * time.Millisecond` then check the stop chan, but here only sleep `5 * time.Millisecond`.
I am not sure it is enough for it.

wilfred-s: This sleep time has nothing to do with the sleep in the PlaceholderManager. The sleep that is here is purely to allow the go routine to process updating the stop flag so we can see that it has been done.
The `stopChan` is an unbuffered channel. The `mgr.Stop()` call is thus blocking until the channel has been read. The channel read happens before the flag gets set that is why we need to sleep in the test. The atomic update of the flag must be done before we check. Without the small sleep in the test the go routine might not have done that on a multi core machine when the test progresses.

You can see that the call is blocking when you run the old code and the test. It takes 8s to complete. When I replaced the `time.Sleep(3 * time.Second)` in the old test with `time.Sleep(5 * time.Millisecond)` the test finished in 5s. Similar when I changed the clean interval the test ran as long as the clean interval was set to.

HuangTing-Yao: OK, got it. Thanks.

HuangTing-Yao (approved):
LGTM

PR #252, opened 2021-04-10
[YUNIKORN-628]Publish a pod event to indicate the task is being gang scheduling

HuangTing-Yao:
### What is this PR for?
Publish the pod level event message to pods, which will be gang scheduled.
Indicate user those pods will be gang scheduled.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-628

Unit test had been added to test the the `GetRecorder()` and `Eventf()` called time. 
### Questions:

### DIFF
--- a/pkg/cache/task.go
+++ b/pkg/cache/task.go
@@ -265,11 +265,12 @@ func (task *Task) handleSubmitTaskEvent(event *fsm.Event) {
 
 	events.GetRecorder().Eventf(task.pod, v1.EventTypeNormal, "Scheduling",
 		"%s is queued and waiting for allocation", task.alias)
-	// if this task is belong to the gang member, then post a message to indicate the pod will being scheduled as gang member
+	// if this task belongs to a task group, that means the app has gang scheduling enabled
+	// in this case, post an event to indicate the task is being gang scheduled
 	if !task.placeholder && task.taskGroupName != "" {
 		events.GetRecorder().Eventf(task.pod,
 			v1.EventTypeNormal, "GangScheduling",
-			"Pod belongs to the taskGroup %s, it will being scheduled as a gang member", task.taskGroupName)
+			"Pod belongs to the taskGroup %s, it will be scheduled as a gang member", task.taskGroupName)
 	}
 }
 
--- a/pkg/cache/task.go
+++ b/pkg/cache/task.go
@@ -265,6 +265,12 @@ func (task *Task) handleSubmitTaskEvent(event *fsm.Event) {
 
 	events.GetRecorder().Eventf(task.pod, v1.EventTypeNormal, "Scheduling",
 		"%s is queued and waiting for allocation", task.alias)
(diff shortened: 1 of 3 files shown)


pkg/cache/task.go
```diff
@@ -265,6 +265,12 @@ func (task *Task) handleSubmitTaskEvent(event *fsm.Event) {
 
 	events.GetRecorder().Eventf(task.pod, v1.EventTypeNormal, "Scheduling",
 		"%s is queued and waiting for allocation", task.alias)
+	// if this task is belong to the gang member, then post a message to indicate the pod will being scheduled as gang member
```
yangwwei, line 268: Suggest changing the text to the following:

```
// if this task belongs to a task group, that means the app has gang scheduling enabled
// in this case, post an event to indicate the task is being gang scheduled
```

pkg/cache/task_test.go
```diff
@@ -28,13 +29,21 @@ import (
...
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/events"
+	"github.com/apache/incubator-yunikorn-k8shim/pkg/conf"
 	"github.com/apache/incubator-yunikorn-scheduler-interface/lib/go/si"
 )
 
+// type recorderTime struct {
+// 	time int64
+// 	lock *sync.RWMutex
+// }
+
```
yangwwei, line 46: if this is not used, pls remove them

pkg/cache/task.go
```diff
@@ -265,6 +265,12 @@ func (task *Task) handleSubmitTaskEvent(event *fsm.Event) {
...
+	if !task.placeholder && task.taskGroupName != "" {
+		events.GetRecorder().Eventf(task.pod,
+			v1.EventTypeNormal, "GangScheduling",
+			"Pod belongs to the taskGroup %s, it will being scheduled as a gang member", task.taskGroupName)
```
yangwwei, line 272: Suggest changing the message to:

"Pod belongs to the taskGroup %s, it will be scheduled as a gang member"

pkg/cache/task.go
HuangTing-Yao: OK,Sure

pkg/cache/task_test.go
HuangTing-Yao: OK.

pkg/cache/task.go
HuangTing-Yao: Sure.

HuangTing-Yao:
When testing I found that data race happened, it caused by `GetRecorder()` and `SetRecorderForTest()`, so add lock for them.

yangwwei (approved):
Looks good, thanks @HuangTing-Yao

PR #253, opened 2021-04-14
[YUNIKORN-643] Improve the spark-k8s-operator app mgmt plugin to better handling Spark CRDs

yuchaoran2011:
### What is this PR for?
Improved the integration with Spark operator plugin. The areas of improvement as as follows:
* Make sure that individual pods fall under the responsibility of the general plugin. Make the Spark operator plugin manage the SparkApplication CRD only
* Make sure that the general and operator plugins see the same application ID for the same SparkApplication. The ID is the string generated by the Spark K8s backend that starts with "spark-"
* Make sure that when a SparkApplication completes or fails, cleanup is performed for all placeholders if there's any that remain at that time
* When cleaning up placeholders, there are often cases where the placeholder doesn't exist anymore and that will cause the deletion to fail. Do not add the placeholder to the orphan pod list in this case because there's no need to retry deletion.
* Added and updated some logging to aid monitoring and debugging

### What type of PR is it?
* [x] - Bug Fix
* [x] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* This is only part of the work for the referenced JIRA. A separate PR will be made on https://github.com/GoogleCloudPlatform/spark-on-k8s-operator to complete the integration.

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-558 (Integrate YuniKorn with Spark K8S Operator)

All the existing unit tests passed. Not quite sure how to unit test this feature specifically. But probably we can enhance the integration tests to cover it. We have been running this patch for relatively large Spark operator workloads for a few weeks and it has been stable for us.


yuchaoran2011:
Looks like there's a potential data race that happened in one of the unit tests. I'll look into it

pkg/cache/application.go
```diff
@@ -511,11 +517,24 @@ func (app *Application) handleRejectApplicationEvent(event *fsm.Event) {
 		fmt.Sprintf("application %s is rejected by scheduler", app.applicationID)))
 }
 
+func placeholderCleanup(app *Application) {
+	placeholderManager := getPlaceholderManager()
+	if placeholderManager != nil {
+		placeholderManager.cleanUp(app)
+	}
+}
+
```
wilfred-s, line 526: `getPlaceholderManager()` cannot return a nil we can simplify this by directly calling this inside the go function while handling the events:
```
go func() {
    getPlaceholderManager().cleanup(app)
}
```

pkg/cache/placeholder_manager.go
```diff
@@ -46,10 +47,13 @@ type PlaceholderManager struct {
 }
 
 var placeholderMgr *PlaceholderManager
+var placeholderManagerLock sync.Mutex
```
wilfred-s, line 50: The init and usage of the manager has not changed. Having a lock like this seems weird because we should only ever the one that was created by the current test, or when the context is created in production.
This probably also explains why you needed the nil check in the app code before.
This is a test problem. By adding the manager usage to the fail application test we are re-using the manager from a previous test. We should not be doing that. Please check the way `TestTryReserve()` and `TestTryReservePostRestart()` initiate the manager.
It looks like the test context does not create a manager and we might not clean up correctly either

wilfred-s (changes requested):

pkg/cache/placeholder_manager.go
yuchaoran2011: You are exactly correct. I have fixed the `TestFailApplication` following the example of `TestTryReserve`

go.mod
```diff
@@ -18,20 +18,20 @@
 
 module github.com/apache/incubator-yunikorn-k8shim
 
-go 1.12
+go 1.14
```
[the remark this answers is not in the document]
yuchaoran2011: Agreed. Rolled back

pkg/appmgmt/general/general.go
```diff
@@ -34,6 +33,8 @@ import (
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/utils"
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/log"
 	"github.com/apache/incubator-yunikorn-scheduler-interface/lib/go/si"
+
+	"go.uber.org/zap"
```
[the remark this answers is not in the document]
yuchaoran2011: This is due to the usage of logging in this file. If I removed this import, `make lint` would complain

go.mod
```diff
@@ -18,20 +18,20 @@
...
 	github.com/looplab/fsm v0.1.0
-	github.com/onsi/ginkgo v1.11.0
-	github.com/onsi/gomega v1.7.0
-	go.uber.org/zap v1.13.0
+	github.com/onsi/ginkgo v1.15.1
+	github.com/onsi/gomega v1.11.0
+	go.uber.org/zap v1.16.0
 	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
-	gopkg.in/yaml.v2 v2.2.8
+	gopkg.in/yaml.v2 v2.4.0
```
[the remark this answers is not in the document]
yuchaoran2011: Agreed. Rolled back

.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea
 _output/
+vendor
```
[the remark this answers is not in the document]
yuchaoran2011: Yeah this is mostly for my own convenience when looking at the code of the dependencies. Removed

deployments/image/configmap/Dockerfile
```diff
@@ -21,7 +21,7 @@ FROM alpine:latest
 RUN apk add curl
 RUN apk add jq
 RUN apk add --update openssl
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.15.12/bin/linux/amd64/kubectl
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.18.16/bin/linux/amd64/kubectl
```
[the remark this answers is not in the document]
yuchaoran2011: Sure. Rolled back

pkg/appmgmt/sparkoperator/spark.go
```diff
@@ -51,6 +40,7 @@ type Manager struct {
 }
 
 func NewManager(amProtocol interfaces.ApplicationManagementProtocol, apiProvider client.APIProvider) *Manager {
+	log.Logger().Info("New AppMgr initialized")
```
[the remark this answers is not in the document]
yuchaoran2011: Yeah this is redundant. Removed

go.mod
```diff
@@ -21,7 +21,7 @@ module github.com/apache/incubator-yunikorn-k8shim
 go 1.12
 
 require (
-	github.com/GoogleCloudPlatform/spark-on-k8s-operator v0.0.0-20200817155620-c19d2b8660d8
+	github.com/GoogleCloudPlatform/spark-on-k8s-operator v0.0.0-20201215015655-2e8b733f5ad0
```
yangwwei, line 24: which version of spark-k8s-operator this line point to? 
I think we need to point to the most stable released version

deployments/image/configmap/Dockerfile
```diff
@@ -46,8 +46,8 @@ ENV DISPATCHER_TIMEOUT "300s"
 ENV KUBE_CLIENT_QPS "1000"
 ENV KUBE_CLIENT_BURST "1000"
 ENV PREDICATES ""
-ENV OPERATOR_PLUGINS "general"
 ENV ENABLE_CONFIG_HOT_REFRESH "true"
+ENV OPERATOR_PLUGINS "general,spark-k8s-operator"
```
yangwwei, line 50: this change will build the docker image by default with spark-k8s-operator plugin enabled. 
We might want to provide an option instead of doing this by default. I think we can add the env var  `OPERATOR_PLUGINS` in the helm chart template: https://github.com/apache/incubator-yunikorn-release/blob/2ade62ac43f9f62f804739f418fcdcb256b9703b/helm-charts/yunikorn/templates/deployment.yaml#L55.  And expose this in https://github.com/apache/incubator-yunikorn-release/blob/master/helm-charts/yunikorn/values.yaml.

pkg/appmgmt/sparkoperator/spark.go
```diff
@@ -23,25 +23,14 @@ import (
...
-	"github.com/apache/incubator-yunikorn-scheduler-interface/lib/go/si"
 )
 
-const (
-	LabelAnnotationPrefix = "sparkoperator.k8s.io/"
-	SparkAppNameLabel     = LabelAnnotationPrefix + "app-name"
-)
-
-// implements interfaces#Recoverable, interfaces#AppManager
+// Manager implements interfaces#Recoverable, interfaces#AppManager
```
yangwwei, line 33: minor: suggest to add some more code comment to explain what has been done by this plugin

pkg/cache/application.go
```diff
@@ -460,9 +461,14 @@ func (app *Application) postAppAccepted() {
 	// app could have allocated tasks upon a recovery, and in that case,
 	// the reserving phase has already passed, no need to trigger that again.
 	var ev events.SchedulingEvent
+	log.Logger().Info("postAppAccepted on cached app",
+		zap.String("appID", app.applicationID),
+		zap.Int("numTaskGroups", len(app.taskGroups)),
+		zap.Int("numAllocatedTasks", len(app.getTasks(events.States().Task.Allocated))))
```
yangwwei, line 467: this seems more like  a DEBUG level logging
any reason why we want to expose this in INFO level?

pkg/cache/application.go
```diff
@@ -460,9 +461,14 @@ func (app *Application) postAppAccepted() {
...
 	if len(app.taskGroups) != 0 &&
 		len(app.getTasks(events.States().Task.Allocated)) == 0 {
 		ev = NewSimpleApplicationEvent(app.applicationID, events.TryReserve)
+		log.Logger().Info("tryReserve")
```
yangwwei, line 471: minor: pls improve this line of logging
something like: "app <appID> has taskGroups defined, trying to reserve resources for the gang members"

pkg/common/events/events.go
```diff
@@ -33,20 +33,20 @@ type SchedulingEvent interface {
...
+	RunApplication          ApplicationEventType = "RunApplication"
+	RejectApplication       ApplicationEventType = "RejectApplication"
+	CompleteApplication     ApplicationEventType = "CompleteApplication"
+	FailApplication         ApplicationEventType = "FailApplication"
+	KillApplication         ApplicationEventType = "KillApplication"
+	KilledApplication       ApplicationEventType = "KilledApplication"
+	ReleaseAppAllocation    ApplicationEventType = "ReleaseAppAllocation"
 	ReleaseAppAllocationAsk ApplicationEventType = "ReleaseAppAllocationAsk"
-	AppStateChange       ApplicationEventType = "ApplicationStateChange"
+	AppStateChange          ApplicationEventType = "ApplicationStateChange"
```
yangwwei, line 49: do we have any change here other than formating?
if not, prefer not to change this in this PR

yangwwei (changes requested):

go.mod
yuchaoran2011: It's this version: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/releases/tag/v1beta2-1.2.2-3.0.0. When I manually entered `v1beta2-1.2.2-3.0.0`, the go module system automatically changed it to the string in the PR

deployments/image/configmap/Dockerfile
yuchaoran2011: Sure, I'll roll back this line to include only the general plugin by default. I'll create another JIRA to track the work on exposing it in the Helm chart

pkg/cache/application.go
yuchaoran2011: Yeah makes sense. I'll update to DEBUG level

pkg/common/events/events.go
yuchaoran2011: Ok. I'll roll back this change. It was introduced when I ran `go fmt`

pkg/appmgmt/sparkoperator/spark.go
yuchaoran2011: I added more comments in this file to explain what the plugin does

deployments/image/configmap/Dockerfile
yuchaoran2011: Created https://issues.apache.org/jira/browse/YUNIKORN-642 to track the follow-up work

yuchaoran2011:
It's strange why the latest commit caused e2e test failures. It only contained changes to the comments and formatting, as well as roll back the plugins that are enabled by default to the "general" only

yangwwei:
> It's strange why the latest commit caused e2e test failures. It only contained changes to the comments and formatting, as well as roll back the plugins that are enabled by default to the "general" only

Looks like some intermittent issue, I have rerun the tests and it looks good now.

yangwwei (approved):
LGTM

PR #258, opened 2021-05-05
[YUNIKORN-657] Expose application failure reasons

yuchaoran-apple:
### What is this PR for?
When an application fails or is rejected by the scheduler, the pods currently stay in pending forever without visibility on what happened to them. This PR propagates the scheduling error or failure to the pods so that the user will be able to see the reason of failure. If third-party operators (e.g. Spark K8s Operator) are used, they can also react according to the pod status change and update the status of the corresponding CRD and possibly retry. Core side changes are in https://github.com/apache/incubator-yunikorn-core/pull/271

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-657

Added unit tests to cover this change. When running in a cluster, tested that scheduling failures will be reflected on the pods:

For example, if an application is rejected, its pods will be marked as failed and its `status` field will be populated with the following information:
```
status:
  message: ' application ''spark-<ID>'' rejected, cannot
    create queue ''root.spark'' without placement rules'
  phase: Failed
  reason: ApplicationRejected
```
Another example is that if a gang-scheduled application cannot successfully run all placeholders before they expire, the app pods' `status` field will show:
```
status:
  message: Scheduling has timed out due to insufficient resources
  phase: Failed
  reason: InsufficientResources
```

### Questions:

### DIFF
--- a/pkg/client/kubeclient.go
+++ b/pkg/client/kubeclient.go
@@ -118,3 +118,28 @@ func (nc SchedulerKubeClient) Delete(pod *v1.Pod) error {
 	}
 	return nil
 }
+
+func (nc SchedulerKubeClient) Get(podNamespace string, podName string) (*v1.Pod, error) {
+	pod, err := nc.clientSet.CoreV1().Pods(podNamespace).Get(podName, apis.GetOptions{})
+	if err != nil {
+		log.Logger().Warn("failed to get pod",
+			zap.String("namespace", pod.Namespace),
+			zap.String("podName", pod.Name),
+			zap.Error(err))
+		return nil, err
+	}
+	return pod, nil
+}
+
+func (nc SchedulerKubeClient) Update(pod *v1.Pod) (*v1.Pod, error) {
+	var updatedPod *v1.Pod
+	var err error
+	if updatedPod, err = nc.clientSet.CoreV1().Pods(pod.Namespace).UpdateStatus(pod); err != nil {
+		log.Logger().Warn("failed to update pod",
+			zap.String("namespace", pod.Namespace),
+			zap.String("podName", pod.Name),
+			zap.Error(err))
+		return pod, err
+	}
+	return updatedPod, nil
+}
--- a/pkg/client/kubeclient.go
+++ b/pkg/client/kubeclient.go
@@ -131,11 +131,11 @@ func (nc SchedulerKubeClient) Get(podNamespace string, podName string) (*v1.Pod,
 	return pod, nil
 }
 
-func (nc SchedulerKubeClient) Update(pod *v1.Pod) (*v1.Pod, error) {
+func (nc SchedulerKubeClient) UpdateStatus(pod *v1.Pod) (*v1.Pod, error) {
 	var updatedPod *v1.Pod
 	var err error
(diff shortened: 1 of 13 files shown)


pkg/common/utils/utils.go
```diff
@@ -58,6 +58,10 @@ func NeedRecovery(pod *v1.Pod) (bool, error) {
 		return true, nil
 	}
 
+	if pod.Status.Phase == v1.PodFailed && pod.Status.Reason == constants.ApplicationRejectedFailure || pod.Status.Reason == constants.ApplicationInsufficientResourcesFailure {
+		return false, nil
+	}
```
yangwwei, line 63: Can we simply return false here when pod phase is Failed? 
Do we have to check the reason?

pkg/shim/scheduler.go
```diff
@@ -234,7 +234,17 @@ func (ss *KubernetesShim) GetSchedulerState() string {
...
+		eventArgs := make([]string, 2)
+		if err = events.GetEventArgsAsStrings(eventArgs, se.GetArgs()); err != nil {
+			log.Logger().Error("fail to parse event arg", zap.Error(err))
+		}
+		message := eventArgs[1]
+		err = ss.stateMachine.Event(string(se.GetEvent()), message)
+	} else {
+		err = ss.stateMachine.Event(string(se.GetEvent()))
+	}
 	if err != nil && err.Error() == "no transition" {
```
yangwwei, line 248: Can we use a more general form? The interface looks like

```
type SchedulerEvent interface {
  GetEvent() SchedulerEventType
  GetArgs() []interface{}
}
```

Can we leverage the args here:

```
err = ss.stateMachine.Event(string(se.GetEvent()), se.GetArgs())
```

yangwwei (changes requested):

yangwwei:
Overall looks good, thanks for improving this! Please see some minor comments, thank you @yuchaoran-apple .

pkg/client/kubeclient.go
```diff
@@ -118,3 +118,28 @@ func (nc SchedulerKubeClient) Delete(pod *v1.Pod) error {
...
+	var err error
+	if updatedPod, err = nc.clientSet.CoreV1().Pods(pod.Namespace).UpdateStatus(pod); err != nil {
+		log.Logger().Warn("failed to update pod",
+			zap.String("namespace", pod.Namespace),
+			zap.String("podName", pod.Name),
+			zap.Error(err))
+		return pod, err
+	}
+	return updatedPod, nil
+}
```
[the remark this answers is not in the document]
yuchaoran-apple: Good catch. Changed to `UpdateStatus`

pkg/common/utils/utils.go
yuchaoran-apple: You are right. Removed reason check

pkg/shim/scheduler.go
yuchaoran-apple: Good point. Updated accordingly. Now it's more generic and cleaner 👍

yangwwei (approved):
LGTM

PR #260, opened 2021-05-13
YUNIKORN-585: Create a Failing app state in shim side

manirajv06:
### What is this PR for?

Create a "Failing state" for the application in the shim side and take care of corresponding events transition.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-585

### Questions:


pkg/cache/application_test.go
```diff
@@ -880,3 +880,144 @@ func TestTriggerAppRecovery(t *testing.T) {
...
+	task1.sm.SetState(events.States().Task.New)
+	task2.sm.SetState(events.States().Task.Pending)
+	task3.sm.SetState(events.States().Task.Scheduling)
+	task4.sm.SetState(events.States().Task.Allocated)
+	app.addTask(task1)
+	app.addTask(task2)
+	app.addTask(task3)
+	app.addTask(task4)
+	task1.allocationUUID = UUID
+	app.SetState(events.States().Application.Accepted)
```
kingamarton, line 1004: Why are you setting the app state to Accepted first? Couldn't we just set it directly to Running?

kingamarton:
@manirajv06  I restarted the tests, but it failed again. Can you please investigate on the cause of the failure?

manirajv06:
> @manirajv06 I restarted the tests, but it failed again. Can you please investigate on the cause of the failure?

It is complaining about "key", "operator" etc as unknown fields. I don't think it is related to this change.

pkg/cache/application.go
```diff
@@ -107,13 +107,13 @@ func NewApplication(appID, queueName, user string, tags map[string]string, sched
 				Src: []string{states.Running},
 				Dst: states.Running},
 			{Name: string(events.ReleaseAppAllocation),
-				Src: []string{states.Failed},
+				Src: []string{states.Failing},
```
kingamarton, line 110: I think this transition is not correct. Let's say we have an application with 3 allocations. The app will be moved into the Failed state right after removing the first allocation. We want to have the app Failed only when everything is cleared and no allocations and asks are left.

pkg/cache/application.go
```diff
@@ -107,13 +107,13 @@ func NewApplication(appID, queueName, user string, tags map[string]string, sched
...
 				Dst: states.Failed},
 			{Name: string(events.ReleaseAppAllocationAsk),
 				Src: []string{states.Running, states.Accepted, states.Reserving},
 				Dst: states.Running},
 			{Name: string(events.ReleaseAppAllocationAsk),
-				Src: []string{states.Failed},
+				Src: []string{states.Failing},
```
kingamarton, line 116: The same here as for the ReleaseAppAllocation

kingamarton (changes requested):
I think the state machine is not complete. I cannot see when we will go into Failing state.

manirajv06:
@kingamarton as we discussed earlier, travis errors not related to this change. Can we take a final look and commit?

kingamarton:
> @kingamarton as we discussed earlier, travis errors not related to this change. Can we take a final look and commit?

I restarted the precommit tests and now there are some unit test failures, can you please check them?

pkg/shim/scheduler_test.go
```diff
@@ -132,7 +132,7 @@ partitions:
 
 	// wait for scheduling app and tasks
 	// verify app state
-	cluster.waitAndAssertApplicationState(t, appID, events.States().Application.Failed)
+	cluster.waitAndAssertApplicationState(t, appID, events.States().Application.Failing)
```
kingamarton, line 135: I don't think this fix is OK here. When the application is rejected by the scheduler, the app should fail. Otherwise in this scenario, the app will never Fail, it will stuck in the Failing state.

kingamarton (changes requested):

pkg/shim/scheduler_test.go
manirajv06: Yes, correct. In case of app rejection, we should directly move the app to 'Failed' state. Handled it using appropriate state transition, rejected -> failed during fail application event. Please check.

kingamarton (approved):
The changes now looks good. Thanks @manirajv06

PR #264, opened 2021-05-18
[YUNIKORN-591] move TaskGroupInstanceCountMap to its own file

0yukali0:
### What is this PR for?
move TaskGroupInstanceCountMap to its own file.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-591

Use go test or debug in IDE to view message. 

### Questions:


pkg/common/utils/gang_TGInstanceCountMap.go
```diff
@@ -0,0 +1,93 @@
+/*
```
yangwwei, line 1: this file name is too long. Since we have a short term "tg" for task group, I suggest renaming this file to:  `tg_map.go`

yangwwei (changes requested):
The UT failure was caused by a data race, I don't think that is introduced by this PR.  I have created https://issues.apache.org/jira/browse/YUNIKORN-673 to track that.

yangwwei (approved):
LGTM

PR #265, opened 2021-05-20
[YUNIKORN-677] Potential resource leak when complete and allocate pod happens simultaneously

yangwwei:
### What is this PR for?
When pod deletion and allocation happen simultaneously, there is a race that can possibly lead to resource leaks. This causes the pods could not be released properly even the job has finished. This can be observed from the web UI if a job already finished (no pods running anymore on K8s) but on YK UI it is still showing as Running with a few allocations, which means we are hitting this issue.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-677

Tested locally

### Questions:


wilfred-s (commented):
The fix is inline with the analysis that I did on the issue we saw on our deployment. One question to make sure that my understanding is complete:
The check we do for the allocation is in the before hook for task allocated. In that hook we check if the state is task completed. That looks good, however does that cover the fact that a task could already be deleted or is that not a possible case to cover?

The lint check fails on a implicit indirection that should be fixed before we go further.

yangwwei:
> The check we do for the allocation is in the before hook for task allocated. In that hook we check if the state is task completed. That looks good, however does that cover the fact that a task could already be deleted or is that not a possible case to cover?

Yes, it should be covered. Today, when we delete a pod, that also triggers a CompleteTask event, so the task will be always in the Completed state. This reminds me that we need to do the proper cleanup for the apps in the shim side cache. We can do that when the app is in the terminated state on the core side, so we can safely remove it.

wilfred-s (approved):
LGTM

PR #267, opened 2021-05-23
[YUNIKORN-686] [Regression] Placeholders for completed applications are recreated during recovery

yangwwei:
### What is this PR for?
After https://issues.apache.org/jira/browse/YUNIKORN-677, @yuchaoran2011 found a regression. See more info in https://issues.apache.org/jira/browse/YUNIKORN-686. The fix in #265 introduces an issue that while we do recovery, we added the terminated pod into the app with New state, this will cause the scheduler to try to schedule the pod again. When GS is enabled, the placeholders are created again.

While fixing this issue, I decided to do some code refactoring to simplify the logic. That's why it seems I have changed more lines of code than expected. The refactoring removes the "Recovery" flag in `TaskMetadata`, and moves the logic to `task. initialization()`, this ensures when we create the task in the cache, we always ensure the task state is initialized correctly. To be more specific:

1.  if the task pod was already allocated and running, the task state post-recovery will be `Allocated`
2. if the task pod was already allocated and terminated (failed or succeed), the task state post-recovery will be "Completed"
3. if the task pod was in Pending state, the task state post-recovery will be `New`

Also, this PR includes some new UT to cover these scenarios.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-686

Tested recovery locally:
1.  submit a job and let the job finish, restart the scheduler, make sure the app was recovered with 0 allocation
2. submit a job and keep the job running, restart the scheduler, make sure the app was recovered with all existing allocations recovered
3. submit a GS job, wait for the job to complete. restart the scheduler, make sure no placeholders are created

### Questions:


pkg/cache/application.go
```diff
@@ -456,6 +456,29 @@ func (app *Application) handleRecoverApplicationEvent(event *fsm.Event) {
 	}
 }
 
+func (app *Application) skipReservationStage() bool {
+	// no task groups defined, skip reservation
+	if len(app.taskGroups) == 0 {
+		log.Logger().Debug("Skip reservation stage: no task groups defined")
```
wilfred-s, line 462: Add the appID to the log to allow tracking the flow correctly. There might be multiple apps doing the same thing.

pkg/cache/application.go
```diff
@@ -456,6 +456,29 @@ func (app *Application) handleRecoverApplicationEvent(event *fsm.Event) {
...
+	}
+
+	// if there is any task already passed New state,
+	// that means the scheduler has already tried to schedule it
+	// in this case, we should skip the reservation stage
+	if len(app.taskMap) > 0 {
+		for _, task := range app.taskMap {
+			if task.GetTaskState() != events.States().Task.New {
+				log.Logger().Debug("Skip reservation stage: found task already has been scheduled before.",
+					zap.String("taskID", task.GetTaskID()),
```
wilfred-s, line 473: We should have the appID also in this log entry.

pkg/cache/application.go
```diff
@@ -456,6 +456,29 @@ func (app *Application) handleRecoverApplicationEvent(event *fsm.Event) {
...
+		return true
+	}
+
+	// if there is any task already passed New state,
+	// that means the scheduler has already tried to schedule it
+	// in this case, we should skip the reservation stage
+	if len(app.taskMap) > 0 {
+		for _, task := range app.taskMap {
+			if task.GetTaskState() != events.States().Task.New {
+				log.Logger().Debug("Skip reservation stage: found task already has been scheduled before.",
```
wilfred-s, line 472: The application has _at least_ one task that has been scheduled before. The exact task does not matter and since it is a range over a map it will log a random taskID. The same input will generate different log entries.

wilfred-s (changes requested):
logging changes.
I assumed the logic is covering all cases based on @yuchaoran2011 comment in the jira

pkg/cache/application.go
yangwwei: This is correct, but I think this should be fine.

pkg/cache/application.go
yangwwei: Make sense, fixed.

pkg/cache/application.go
yangwwei: Make sense, fixed.

pkg/cache/task.go
```diff
@@ -236,6 +238,36 @@ func (task *Task) isTerminated() bool {
...
+// we should mark the task as completed according.
+func (task *Task) initialize() {
+	task.lock.Lock()
+	defer task.lock.Unlock()
+
+	// task needs recovery means the task has already been
+	// scheduled by us with an allocation, instead of starting
+	// from New, directly set the task to Allocated.
+	if utils.NeedRecovery(task.pod) {
+		log.Logger().Info("init allocated task")
```
kingamarton, line 254: Here in the log message we should include the at least the allocation UID,for better troubleshooting

kingamarton (changes requested):
Overall looks good. I have a small comment related a log message, where some identification details were missing.

pkg/cache/task.go
yangwwei: Added some more logs

kingamarton (approved):

PR #269, opened 2021-05-30
YUNIKORN-679: Add gangSchedulingStyle scheduling parameter

manirajv06:
### What is this PR for?

Introduce gangSchedulingStyle attribute, attach with pod's annotation: yunikorn.apache.org/schedulingPolicyParameters and send to core as part of Application Metadata to help core to decide the next steps (whether to continue to return the error or continue as simple application) after gang request timeout has been exceeded. Possible values are 1. Soft 2. Hard (Default).
Also, introduce "resuming" app state and takes care of events transition.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-679

### Questions:


pkg/appmgmt/general/general.go
```diff
@@ -134,12 +134,18 @@ func (os *Manager) getAppMetadata(pod *v1.Pod) (interfaces.ApplicationMetadata,
 	}
 	ownerReferences := getOwnerReferences(pod)
 
-	placeholderTimeout, err := utils.GetPlaceholderTimeoutParam(pod)
-	if err != nil {
+	placeholderTimeout, placeholderTimeoutErr, schedulingStyle, schedulingStyleErr := utils.GetSchedulingPolicyParam(pod)
```
kingamarton, line 137: this is a bit confusing. Please create 2 different methods (GetPlaceholderTimeoutParam and GetSchedulingStyle) and eventually export the repeated parts into a separate one.

pkg/appmgmt/general/general_test.go
```diff
@@ -96,6 +98,9 @@ func TestGetAppMetadata(t *testing.T) {
 				"queue":                    "root.b",
 				"yunikorn.apache.org/user": "testuser",
 			},
+			Annotations: map[string]string{
+				constants.SchedulingPolicyStyleParam: "gangSchedulingStyle=Hard",
```
kingamarton, line 102: Please export the parameter name in a constant and use that one. Also use constants for the possible values

pkg/callback/scheduler_callback.go
```diff
@@ -149,6 +149,16 @@ func (callback *AsyncRMCallback) RecvUpdateResponse(response *si.UpdateResponse)
...
 			}
+		} else if updated.State == events.States().Application.Resuming {
+			app := callback.context.GetApplication(updated.ApplicationID)
+			if app != nil && app.GetApplicationState() == events.States().Application.Reserving {
+				ev := cache.NewRunApplicationEvent(updated.ApplicationID)
+				dispatcher.Dispatch(ev)
+
+				// handle status update
+				dispatcher.Dispatch(cache.NewApplicationStatusChangeEvent(updated.ApplicationID, events.AppStateChange, updated.State))
+			}
```
kingamarton, line 160: The core side will move the application into the Resuming state before we have all the placeholders removed. See the changes in the PR: https://github.com/apache/incubator-yunikorn-core/pull/275/files

I think when we need to move this part to when we remove the allocations and asks, to make sure that first we delete all the placeholders.

pkg/common/utils/gang_utils.go
```diff
@@ -123,24 +125,42 @@ func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]v1alpha1.TaskGroup, error) {
 	return taskGroups, nil
 }
 
-func GetPlaceholderTimeoutParam(pod *v1.Pod) (int64, error) {
+func GetSchedulingPolicyParam(pod *v1.Pod) (int64, error, string, error) {
```
kingamarton, line 128: Please use two methods instead of mixing the 2 calls. You can export the common parts into one method. See my comment in the general.go file

kingamarton (changes requested):
We need a new `Resuming` state for the shim app state as well, and we should keep the application into this state until all the placeholders are deleted. Please add this state info the state machine as well.
Also you need to pass this value to the core in the `application.go` file `handleSubmitApplicationEvent` method. 
You can check (and review if you want :) ) the interface and core side changes in the following PR's:
interface: https://github.com/apache/incubator-yunikorn-scheduler-interface/pull/38
core: https://github.com/apache/incubator-yunikorn-core/pull/275/files

Also please make some end to end manual testing to be sure that changes are working together properly.

manirajv06:
@kingamarton @yangwwei 

Have taken care of state transitions by introducing new event and state. Please take a look. Will need to cover test cases for sure.

pkg/common/utils/gang_utils.go
```diff
@@ -123,24 +125,53 @@ func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]v1alpha1.TaskGroup, error) {
...
+	var timeoutErr = fmt.Errorf("")
+	var styleErr = fmt.Errorf("")
+	style := constants.SchedulingPolicyStyleParamDefault
 	param, ok := pod.Annotations[constants.AnnotationSchedulingPolicyParam]
 	if !ok {
-		return 0, nil
+		schedulingPolicyParams["placeholderTimeout"] = timeout
+		schedulingPolicyParams["placeholderTimeoutErr"] = timeoutErr
+		schedulingPolicyParams["schedulingStyle"] = style
+		schedulingPolicyParams["schedulingStyleErr"] = styleErr
```
yangwwei, line 139: please do not add `placeholderTimeoutErr` and `schedulingStyleErr` into the return map, which makes the return value pretty odd. for easier access, we can return a struct such as:

```
type schedulingParameters struct {
  placeholderTimeout int64
  gangSchedulingStyle string
}
```

pkg/appmgmt/interfaces/amprotocol.go
```diff
@@ -81,6 +81,7 @@ type ApplicationMetadata struct {
 	TaskGroups              []v1alpha1.TaskGroup
 	PlaceholderTimeoutInSec int64
 	OwnerReferences         []metav1.OwnerReference
+	SchedulingStyle         string
```
yangwwei, line 84: change this to the struct `SchedulingParameters` and add `placeholderTimeout` and `gangSchedulingStyle` into the single struct. So it will be easier to extend as well.

pkg/cache/application.go
```diff
@@ -241,6 +252,12 @@ func (app *Application) setOwnReferences(ref []metav1.OwnerReference) {
 	app.placeholderOwnerReferences = ref
 }
 
+func (app *Application) setSchedulingStyle(schedulingStyle string) {
+	app.lock.RLock()
+	defer app.lock.RUnlock()
```
yangwwei, line 257: this should take the write lock

pkg/cache/application.go
```diff
@@ -599,6 +618,19 @@ func (app *Application) handleReleaseAppAllocationEvent(event *fsm.Event) {
...
 	}
+	app.postAppReleased()
+}
+
+func (app *Application) postAppReleased() {
+	canRunApplication := true
+	for _, task := range app.taskMap {
+		if !utils.IsPodTerminated(task.GetTaskPod()) {
+			canRunApplication = false
+		}
```
yangwwei, line 629: when the shim deletes the placeholder pod, the pod will be gone, I don’t think this is a safe way to check.
instead of doing this check in the shim side, can we leverage the core side app state?
IIUC, the core side has the check when all the placeholders are removed, then moves the app to the Running state. In that case, can we simply rely on that, if the app state from the core transits to the Running state, and the shim is not in Running state yet, we trigger a RunApplication event.

yangwwei (changes requested):

pkg/appmgmt/interfaces/amprotocol.go
manirajv06: Taken care

pkg/cache/application.go
manirajv06: > when the shim deletes the placeholder pod, the pod will be gone, I don’t think this is a safe way to check.
> instead of doing this check in the shim side, can we leverage the core side app state?
> IIUC, the core side has the check when all the placeholders are removed, then moves the app to the Running state. In that case, can we simply rely on that, if the app state from the core transits to the Running state, and the shim is not in Running state yet, we trigger a RunApplication event.

IIU @kingamarton's earlier comments and offline discussions, shim is the right place to ensure all placeholders has been cleaned up correctly and app state would be "reserving" at this point in shim side. So, tried to make changes in that direction as depending blindly on core state doesn't ensure all placeholders has been cleaned up completely. core simply make the state as "resuming" as soon as it sees scheduling style as "soft" when placeholders being timed out. @kingamarton Can you also share your thoughts?

pkg/appmgmt/general/general.go
```diff
@@ -133,22 +133,16 @@ func (os *Manager) getAppMetadata(pod *v1.Pod) (interfaces.ApplicationMetadata,
...
-	}
+	schedulingPolicyParams := utils.GetSchedulingPolicyParam(pod)
 	return interfaces.ApplicationMetadata{
 		ApplicationID:           appID,
 		QueueName:               utils.GetQueueNameFromPod(pod),
 		User:                    user,
 		Tags:                    tags,
 		TaskGroups:              taskGroups,
-		PlaceholderTimeoutInSec: placeholderTimeout,
+		PlaceholderTimeoutInSec: schedulingPolicyParams.GetPlaceholderTimeout(),
```
yangwwei, line 143: instead of having separate parameters `PlaceholderTimeoutInSec` and `SchedulingStyle`, let's use `SchedulingPolicyParameters` field in the `ApplicationMetadata`, this will be more flexible for future changes.

pkg/common/utils/gang_utils.go
```diff
@@ -123,24 +143,47 @@ func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]v1alpha1.TaskGroup, error) {
...
-		timeoutParam := strings.Split(p, "=")
-		if timeoutParam[0] == constants.SchedulingPolicyTimeoutParam {
-			if len(timeoutParam) != 2 {
-				return 0, fmt.Errorf("unable to parse timeout value from annotation")
+		param := strings.Split(p, "=")
+		if param[0] == constants.SchedulingPolicyTimeoutParam {
+			if len(param) != 2 {
+				timeoutErr = true
+				log.Logger().Info("Unable to parse timeout value from annotation", zap.String("namespace", pod.Namespace), zap.String("name", pod.Name), zap.Int64("Using Placeholder timeout: ", timeout), zap.String("Placeholder timeout passed in annotation: ", p))
+			}
```
yangwwei, line 165: this is dup code, we could move the check before the parameter key check.
If len(param) != 2, we can simply skip it with a logging message:  "skipping malformed scheduling policy parameter: xxxx". We can still ensure the default values are used. With this, we can remove `timeoutErr`, `styleErr`, `styleExists` bool values and further simplify the code.

pkg/common/utils/gang_utils.go
```diff
@@ -123,24 +143,47 @@ func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]v1alpha1.TaskGroup, error) {
...
+					zap.String("namespace", pod.Namespace), zap.String("name", pod.Name), zap.String("Gang scheduling style passed in annotation: ", p))
 			}
-			timeout, err := strconv.ParseInt(timeoutParam[1], 10, 64)
-			if err != nil {
-				return 0, fmt.Errorf("failed to parse timeout value: %s", timeoutParam[1])
+			if ! styleErr {
+				style, styleExists = constants.SchedulingPolicyStyleParamValues[param[1]]
+				if !styleExists {
+					log.Logger().Info("Unknown gang scheduling style, using "+constants.SchedulingPolicyStyleParamDefault+" style as default",
+						zap.String("namespace", pod.Namespace), zap.String("name", pod.Name), zap.String("Gang scheduling style passed in annotation: ", p))
```
yangwwei, line 182: log as a WARN

pkg/common/utils/gang_utils.go
```diff
@@ -123,24 +143,47 @@ func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]v1alpha1.TaskGroup, error) {
...
+		if param[0] == constants.SchedulingPolicyTimeoutParam {
+			if len(param) != 2 {
+				timeoutErr = true
+				log.Logger().Info("Unable to parse timeout value from annotation", zap.String("namespace", pod.Namespace), zap.String("name", pod.Name), zap.Int64("Using Placeholder timeout: ", timeout), zap.String("Placeholder timeout passed in annotation: ", p))
+			}
+			if ! timeoutErr {
+				timeout, err = strconv.ParseInt(param[1], 10, 64)
+				if err != nil {
+					log.Logger().Info("Failed to parse timeout value from annotation", zap.String("namespace", pod.Namespace), zap.String("name", pod.Name), zap.Int64("Using Placeholder timeout: ", timeout), zap.String("Placeholder timeout passed in annotation: ", p))
+				}
```
yangwwei, line 170: log as a WARN

pkg/common/utils/gang_utils.go
```diff
@@ -24,13 +24,33 @@ import (
...
+}
+
+func (spp *SchedulingPolicyParameters) GetPlaceholderTimeout() int64 {
+	return spp.placeholderTimeout
+}
+
+func (spp *SchedulingPolicyParameters) GetGangSchedulingStyle() string{
+	return spp.gangSchedulingStyle
+}
+
```
yangwwei, line 53: move this to `amprotocol.go`

pkg/common/utils/gang_utils_test.go
```diff
@@ -290,17 +290,19 @@ func TestGetTaskGroupFromAnnotation(t *testing.T) {
...
-		isErr             bool
 		want              int64
+		expectedStyle     string
 	}{
-		{constants.AnnotationSchedulingPolicyParam, "unknownPara=unkown  placeholderTimeoutInSeconds=50=25", true, int64(0)},
-		{constants.AnnotationSchedulingPolicyParam, "unknownPara=unkown  placeholderTimeoutInSeconds=oneSecond", true, int64(0)},
-		{constants.AnnotationSchedulingPolicyParam, "unknownPara=unkown", false, int64(0)},
-		{"policyParamUndefined", "unknownPara=unkown placeholderTimeoutInSeconds=50", false, int64(0)},
-		{constants.AnnotationSchedulingPolicyParam, "unknownPara=unkown placeholderTimeoutInSeconds=50", false, int64(50)},
+		{constants.AnnotationSchedulingPolicyParam, "unknownPara=unkown placeholderTimeoutInSeconds=50=25 gangSchedulingStyle=Soft=Hard", int64(0), "Hard"},
```
yangwwei, line 299: can we add some code to cover the following cases?

```
gangSchedulingStyle=abc
```

and (without values)

```
placeholderTimeoutInSeconds gangSchedulingStyle
```

yangwwei (changes requested):

kingamarton (changes requested):
Besides the comments left by @yangwwei, please update the core and interface dependency as well. Those changes are already merged, so we can have the e2e test results as well.

manirajv06:
> Besides the comments left by @yangwwei, please update the core and interface dependency as well. Those changes are already merged, so we can have the e2e test results as well.

Can you help me to get the build details of dependencies?

yuchaoran2011:
@manirajv06 Under your local shim repo, run the following two commands will update your dependencies in `go.mod`:
```
go get github.com/apache/incubator-yunikorn-core@23af830
go get github.com/apache/incubator-yunikorn-scheduler-interface@da24a8e
```
The hashes above correspond to the latest commits in the core and SI repos respectively and they contain the gang scheduling style related changes

kingamarton:
> > Besides the comments left by @yangwwei, please update the core and interface dependency as well. Those changes are already merged, so we can have the e2e test results as well.
> 
> Can you help me to get the build details of dependencies?

You can easily update the dependency with the following command: `go get github.com/apache/incubator-yunikorn-scheduler-interface@master` and `go get github.com/apache/incubator-yunikorn-core@master`. This will update the go.mod and go.sum files.

pkg/cache/application.go
```diff
@@ -628,6 +647,22 @@ func (app *Application) handleReleaseAppAllocationEvent(event *fsm.Event) {
...
+	}
+	canRunApplication := true
+	for _, task := range app.taskMap {
+		if task.GetTaskState() != events.States().Task.Completed {
+			canRunApplication = false
+		}
+	}
+	if canRunApplication {
+		dispatcher.Dispatch(NewRunApplicationEvent(app.applicationID))
+	}
```
yangwwei, line 665: I think we miss one important check here (apologies did not find this out last time). 
When we check task states, we should only check the task that are placeholders. The second minor thing, I think we can remove bool var `canRunApplication` to simplify this function a bit. Something like this:

```
if app.GetApplicationState() != events.States().Application.Resuming {
  return
}	

for _, task := range app.taskMap {	
  if  task.placeholder && task.GetTaskState() != events.States().Task.Completed {
     return
  }
}

log.Logger().Info("Resuming completed, start to run the app",
  zap.String("appID", app.applicationID),
  zap.String("reason", eventArgs[0]))
dispatcher.Dispatch(NewRunApplicationEvent(app.applicationID))
```

pkg/common/constants/constants.go
```diff
@@ -64,6 +64,10 @@ const AnnotationTaskGroups = "yunikorn.apache.org/task-groups"
 const AnnotationSchedulingPolicyParam = "yunikorn.apache.org/schedulingPolicyParameters"
 const SchedulingPolicyTimeoutParam = "placeholderTimeoutInSeconds"
 const SchedulingPolicyParamDelimiter = " "
+const SchedulingPolicyStyleParam = "gangSchedulingStyle"
+const SchedulingPolicyStyleParamDefault = "Hard"
```
yangwwei, line 68: this should be `Soft`
we need to keep this consistent with the core side: https://github.com/apache/incubator-yunikorn-core/blob/ef77f43d5c2829bedb12ed18578b8bf155793920/pkg/scheduler/objects/application.go#L107-L111, please remember to update the UT accordingly after this change.

yangwwei (changes requested):

pkg/cache/application.go
```diff
@@ -628,6 +647,22 @@ func (app *Application) handleReleaseAppAllocationEvent(event *fsm.Event) {
 			}
 		}
 	}
+	app.postAppReleased(eventArgs)
```
yangwwei, line 650: hi @manirajv06 sorry, I did not mean to add `eventArgs` to this function, this is not needed. I forgot to delete that from my copy-paste code, sorry about that. Can you remove this arg and keep the function as simple as it was? thanks!

yangwwei (changes requested):

yangwwei (approved):
Look good from my side.
@kingamarton could u pls take another look?

pkg/cache/application.go
```diff
@@ -628,6 +647,21 @@ func (app *Application) handleReleaseAppAllocationEvent(event *fsm.Event) {
...
 	}
+	app.postAppReleased()
+}
+
+func (app *Application) postAppReleased() {
+	if app.GetApplicationState() != events.States().Application.Resuming {
+		return
+	}
+	for _, task := range app.taskMap {
+		if task.placeholder && task.GetTaskState() != events.States().Task.Completed {
```
kingamarton, line 658: This check is not working, even after correcting the missing parts in the `states.go` file. At this point the tasks are not in Completed state but in Scheduling or Bound. I spent some time to understand why the transition is not made, but I have no idea so far.
Even if the transition would work, the state transition is handled in an async way, so we cannot guarantee, that at this point the task will have the updated state. 

Checking the pos's status is also not a good idea, because since the pod is gone, thee task has an old reference.

Maybe we can check if the the task has the TIMEOUT set as terminationType, since that is always set by the core. @yangwwei @manirajv06  what do you think?

kingamarton (changes requested):

pkg/callback/scheduler_callback.go
```diff
@@ -143,13 +143,22 @@ func (callback *AsyncRMCallback) RecvUpdateResponse(response *si.UpdateResponse)
...
+		switch updated.State {
+		case events.States().Application.Completed:
 			err := callback.context.RemoveApplicationInternal(updated.ApplicationID)
 			if err != nil {
 				log.Logger().Error("failed to delete application", zap.Error(err))
 			}
-		} else {
+		case events.States().Application.Resuming:
+			app := callback.context.GetApplication(updated.ApplicationID)
+			if app != nil && app.GetApplicationState() == events.States().Application.Reserving {
```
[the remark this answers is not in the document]
yangwwei: This is a good point, need to get this addressed.

pkg/cache/application.go
yangwwei: Hmm, need to understand why the tasks are not in Completed state. Since the `deletePod` will trigger a `CompleteTask` event and that should move the task state to `Completed`. Can you pls print all the task states in this case?

kingamarton: The transition is processed, but later, since it is an async message, so in this place we cannot rely on the task state. In my test scenarios I had a job with 4 placeholders, and the task state transition was processed after all the asks and allocations were released, so the app will stuck in Resuming state, because at this point all the tasks has the old state. 

What about checking here the terminationType?

pkg/common/events/states.go
```diff
@@ -53,6 +53,7 @@ type ApplicationStates struct {
 	Failing    string
 	Failed     string
 	Resuming   string
+	Any        []string // Any refers to all possible states
```
kingamarton, line 56: Sorry for this one, as I can se we are not using it. I mixed it up with the Task state, where we are using it, and during my debugging I thought that the transition is broken because of that one. If you agree that we don't need it you can remove it.

kingamarton (changes requested):

pkg/cache/application.go
manirajv06: I've made the change as it seem to be the only option for our testing purpose. But still not convinced completely. How does this check guarantees that task has been completed (may be already completed or going to complete) from shim side?

yangwwei: hi @kingamarton @manirajv06 IMO, we should leverage the state machine transitions to handle this. The problem here is where to put the check for transiting the app out of `Resuming`. I think we can do the following:

When the pod got actually deleted, it will call `context.NotifyTaskComplete()`:  https://github.com/apache/incubator-yunikorn-k8shim/blob/18b722990b2bee5cdfaedc441d4c76da604bbd40/pkg/cache/context.go#L423-L431, here currently we only send an event to the task. We can also send an event `AppTaskCompleted` to the app

```
appEv := NewSimpleApplicationEvent(appID, events.AppTaskCompleted) 
dispatcher.Dispatch(appEv) 
```

and we handle 

1. `Resuming` -> `Resuming` on `AppTaskCompleted`, if not all tasks are Completed
2. `Resuming` -> `Running` on `AppTaskCompleted`, if all tasks are in Completed state

I think this will work, could you please see if this makes sense?

manirajv06: Overall approach makes sense to me from place (where event gets triggered) perspective, but I am not really sure the necessity of new app event and its transitions.

Based on the recent changes in this PR, app would be moved from "resuming" to "running" through run application event in newly introduced method postAppReleased(). Now, we can completely remove this flow and its corresponding state transitions (release* events etc). Instead, we can fire run application event only when all tasks are placeholder and completed in NotfiyTaskComplete() method. Above all, this check will happen only app state is resuming. Also, I don't see a need to keep the app state in resuming itself as it won't change by any other events at this point. BTW, NotifyTaskComplete() is being called from update and delete pod. But, I don't think it matters.

Thoughts?

kingamarton: I think both approaches can work, but I think the cleaner solution is what @yangwwei mentioned, to handle the check in the application code, its state machine. 
My problem with adding it to `NotifyTaskComplete` we will change the whole method's purpose and scope.

pkg/cache/application.go
```diff
@@ -693,7 +676,25 @@ func (app *Application) handleReleaseAppAllocationAskEvent(event *fsm.Event) {
...
+
+func (app *Application) handleAppTaskCompletedEvent(event *fsm.Event) {
+	eventArgs := make([]string, 2)
+	if err := events.GetEventArgsAsStrings(eventArgs, event.Args); err != nil {
+		log.Logger().Error("fail to parse event arg", zap.Error(err))
+		return
+	}
+	if app.GetApplicationState() != events.States().Application.Resuming {
+		return
+	}
```
yangwwei, line 689: The state machine should have ensured this is only handled under `Resuming` state:

```
{
  Name: string(events.AppTaskCompleted),
  Src: []string{states.Resuming,
  Dst: states.Resuming
},
```
I think we can safely remove this.

yangwwei (changes requested):
Pls see my comments, also pls fix the lint error. thanks

pkg/cache/application.go
```diff
@@ -100,8 +102,14 @@ func NewApplication(appID, queueName, user string, tags map[string]string, sched
 			{Name: string(events.UpdateReservation),
 				Src: []string{states.Reserving},
 				Dst: states.Reserving},
+			{Name: string(events.ResumingApplication),
+				Src: []string{states.Reserving},
+				Dst: states.Resuming},
+			{Name: string(events.AppTaskCompleted),
+				Src: []string{states.Resuming},
+				Dst: states.Resuming},
```
kingamarton, line 110: We need to define the following transitions as well: 
```
{Name: string(events.ReleaseAppAllocation),
				Src: []string{states.Resuming},
				Dst: states.Resuming},
			{Name: string(events.ReleaseAppAllocationAsk),
				Src: []string{states.Resuming},
				Dst: states.Resuming},
```
Otherwise if the app is in Resuming state, this events are not processed and the placeholders are not deleted.

pkg/common/events/states.go
```diff
@@ -93,6 +95,7 @@ func States() *AllStates {
 				Killing:    "Killing",
 				Killed:     "Killed",
 				Failed:     "Failed",
```
kingamarton, line 97: Is not related to this change but can you please add the string representation for Failing state? `Failing:    "Failing",`

kingamarton (changes requested):
I found the root cause for why the placeholder were not cleaned up: we are missing a transition in the app's state machine. With the suggested changes it is working. 
@manirajv06 if you want I can update thee pr with the changes, since I have them locally and then I can merge it. What do you think?

manirajv06:
> I found the root cause for why the placeholder were not cleaned up: we are missing a transition in the app's state machine. With the suggested changes it is working.
> @manirajv06 if you want I can update thee pr with the changes, since I have them locally and then I can merge it. What do you think?

Thank you for thorough testing to ensure this works end to end.

kingamarton (approved):

PR #270, opened 2021-06-02
YUNIKORN-657: Get the latest copy of a pod before attempting status update. Retry in case of failure

ycr-oss:
### What is this PR for?
Pod status update may occasionally fail because the pod could have been modified since the last Get operation. This PR changed the status update logic to get the latest version before update. Also retry a few times if the update fails.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-657

Ran a deployment with this patch for two weeks and haven't found another instance where the update didn't work

### Questions:

### DIFF
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -557,7 +557,7 @@ func (app *Application) handleFailApplicationEvent(event *fsm.Event) {
 		return
 	}
 	errMsg := eventArgs[0]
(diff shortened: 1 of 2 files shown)


yangwwei (approved):
Looks good, waiting for Travis job results.

PR #273, opened 2021-06-02
[YUNIKORN-674] Fixed admission controller for K8s v. 1.19

kingamarton:
### What is this PR for?
The admission controller failed to start because we failed to generate the certificate.

The error related the cert geneeration is the following one:

`no kind "CertificateSigningRequest" is registered for version "certificates.k8s.io/v1" in scheme "k8s.io/kubectl/pkg/scheme/scheme.go:28"`
Upgrading kubectl to 1.19 will solve the issue for 1.19 version.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [X] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-674

- running the e2e tests for both K8s 1.19 and for older versions as well
- manual tests

### Questions:

### DIFF
--- a/deployments/image/configmap/Dockerfile
+++ b/deployments/image/configmap/Dockerfile
@@ -21,7 +21,7 @@ FROM alpine:latest
 RUN apk add curl
 RUN apk add jq
 RUN apk add --update openssl
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.15.12/bin/linux/amd64/kubectl
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.19.0/bin/linux/amd64/kubectl
 RUN chmod +x ./kubectl
 RUN mv ./kubectl /usr/local/bin/kubectl
 COPY admission-controller-init-scripts/admission_util.sh /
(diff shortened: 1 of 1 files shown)


deployments/image/configmap/Dockerfile
```diff
@@ -21,7 +21,7 @@ FROM alpine:latest
 RUN apk add curl
 RUN apk add jq
 RUN apk add --update openssl
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.15.12/bin/linux/amd64/kubectl
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.19.0/bin/linux/amd64/kubectl
```
yuchaoran2011, line 24: Why don't we use the latest 1.19, which is 1.19.11?

yangwwei: Agree I think we should use latest stable version

yangwwei (changes requested):

yangwwei (approved):

PR #276, opened 2021-06-14
Abort deleting resources when the scheduler is still running

ycr-oss:
### What is this PR for?
When upgrading YuniKorn with `helm upgrade`, the admission controller is deleted, an example of what happens:
```
$ k get po -w
NAME                                             READY   STATUS    RESTARTS   AGE
yunikorn-admission-controller-6f6b9b8dff-4pqv6   1/1     Running   0          2d5h
yunikorn-scheduler-64f8746476-4mqjl              2/2     Running   0          2d5h
yunikorn-scheduler-dcdd789b4-rs4km               0/2     Pending   0          0s
yunikorn-scheduler-dcdd789b4-rs4km               0/2     Pending   0          0s
yunikorn-scheduler-dcdd789b4-rs4km               0/2     ContainerCreating   0          0s
yunikorn-scheduler-dcdd789b4-rs4km               2/2     Running             0          6s
yunikorn-scheduler-64f8746476-4mqjl              2/2     Terminating         0          2d5h
yunikorn-admission-controller-6f6b9b8dff-4pqv6   1/1     Terminating         0          2d5h
yunikorn-admission-controller-6f6b9b8dff-4pqv6   0/1     Terminating         0          2d5h
yunikorn-admission-controller-6f6b9b8dff-4pqv6   0/1     Terminating         0          2d5h
yunikorn-admission-controller-6f6b9b8dff-4pqv6   0/1     Terminating         0          2d5h
yunikorn-scheduler-64f8746476-4mqjl              0/2     Terminating         0          2d5h
yunikorn-scheduler-64f8746476-4mqjl              0/2     Terminating         0          2d5h
yunikorn-scheduler-64f8746476-4mqjl              0/2     Terminating         0          2d5h
```
As shown above, when the upgrade is complete, only the scheduler pod remains but the admission controller pod is gone. This is due to the postStop hook deleting the admission controller. This PR added a simple check in this deletion step: if the scheduler is still healthy and running, don't delete.

Note that after this change, commands like
```
kubectl scale deployment yunikorn-scheduler --replicas=[0|1]
```
still work exactly as before

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-696

Install YuniKorn, then upgrade it with `helm upgrade`, check the right pods are running.

### Questions:

### DIFF
--- a/deployments/admission-controllers/scheduler/admission_util.sh
+++ b/deployments/admission-controllers/scheduler/admission_util.sh
@@ -59,6 +59,11 @@ if [ -z "$ENABLE_CONFIG_HOT_REFRESH" ]; then
   ENABLE_CONFIG_HOT_REFRESH=`cat ${CONF_FILE} | grep ^enableConfigHotRefresh | cut -d "=" -f 2`
 fi
 delete_resources() {
+  readyReplicas=$(kubectl -n ${NAMESPACE} get deployment yunikorn-scheduler -o=jsonpath='{.status.readyReplicas}')
+  if [ "$readyReplicas" -eq "1" ]; then
+    echo "scheduler deployment has 1 replica ready. Aborting deleting resources";
+    return 0
+  fi
   kubectl delete -f server.yaml
   # cleanup admissions
   for admission in $REGISTERED_ADMISSIONS


yangwwei (changes requested):
hi @yuchaoran2011 , sorry for the late response, this issue slips off my radar. Thanks for working on this issue! I think your fix works, but after looking more into this, I think there might be a better way to fix this. The reason is we already have this cleanup job: https://github.com/apache/incubator-yunikorn-release/blob/master/helm-charts/yunikorn/templates/cleanup.yaml. This job runs as a chart hook to clean up the YK resources. So we now have 2 parts of cleanup (a bit weird - this is a old issue):

1.  A `preStop` hook: https://github.com/apache/incubator-yunikorn-release/blob/188b17c6728b40f94a3ded06d093230e2f8cd834/helm-charts/yunikorn/templates/deployment.yaml#L89-L91
2. A `post-delete` hook: https://github.com/apache/incubator-yunikorn-release/blob/188b17c6728b40f94a3ded06d093230e2f8cd834/helm-charts/yunikorn/templates/cleanup.yaml#L23-L24

what we can do is to move the `preStop` cleanup code to the cleanup job, that handles every resource created for the admission-controller, and only gets called when we delete the helm chart.  So we can remove the `preStop` hook, and then add few lines like in [cleanup.yaml](https://github.com/apache/incubator-yunikorn-release/blob/master/helm-charts/yunikorn/templates/cleanup.yaml) , such as

```
+              curl --cacert ${CACERT} --header "Authorization: Bearer ${TOKEN}" -X DELETE ${APISERVER}/apis/apps/v1/namespaces/${NAMESPACE}/deployments/yunikorn-admission-controller;
+              curl --cacert ${CACERT} --header "Authorization: Bearer ${TOKEN}" -X DELETE ${APISERVER}/apis/apps/v1/namespaces/${NAMESPACE}/services/yunikorn-admission-controller-service;
```
And we also need to change the hook from `post-delete` to `pre-delete`, because if we do all the cleanup tasks in the cleanup job pod, we'd better do it BEFORE the scheduler pod gets deleted. Otherwise, the cleanup job pod will be assigned with `schedulerName=yunikorn` and then left Pending because the YK scheduler pod already terminated.

Do you think this will be better? 
 
One more thing. After this change, we need to add some docs:
1) document how to do `helm upgrade`
2) add a note to the `helm upgrade` doc, say that the upgrade only updates the scheduler not the admission-controller.

can we have a follow up JIRA to track the doc changes?
hope this makes sense, thanks!

ycr-oss:
@yangwwei Thanks for the review! Your suggestion sounds good. I'll make the changes next week and also create a JIRA to track the work on documentation

ycr-oss:
Closing this PR as I'll make another one at the release repo. https://issues.apache.org/jira/browse/YUNIKORN-710 created to track the doc work

ycr-oss:
@yangwwei I just realized one potential issue with the suggested approach. If we remove the `preStop` lifecycle hook, then when running the command:
```
kubectl scale deployment yunikorn-scheduler --replicas=0
```
will disable the scheduler pod but leave the admission controller running, causing all incoming pods to be patched with `schedulerName=yunikorn` but they can't be scheduled because the scheduler pod is down

yangwwei:
That's correct. I also noticed the same. I think after this change, we should document this and let the users use `helm install/delete/upgrade` to manage YK. And command `scale deployment` will only be used if the user wants to restart YK.

ycr-oss:
Sounds good. I'll take care of it in the documentation

PR #280, opened 2021-07-12
[YUNIKONR-706] Fix the resource calculation when the pod has init-containers

0yukali0:
### What is this PR for?
Original function will miss initcontainers requirement in podspec when pod has initcontainers and containers. 
The reason is that resources calculation function only check resources requirement of containers in pod.
I modify resource calculation in task which use function in common package to confirm it.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-706

I add test case to check calculation correctly.
Use make test to recheck.

### Questions:

### DIFF
--- a/pkg/common/resource.go
+++ b/pkg/common/resource.go
@@ -72,7 +72,33 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
 		containerResource := getResource(resourceList)
 		podResource = Add(podResource, containerResource)
 	}
+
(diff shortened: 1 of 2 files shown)


0yukali0:
Sorry , I just found out there is logic problem here.
I will fix it and ask a PR later.

PR #281, opened 2021-07-12
[YUNIKONR-706] Fix the resource calculation when the pod has init-containers

0yukali0:
### What is this PR for?
Original function will miss initcontainers requirement in podspec when pod has initcontainers and containers.
The reason is that resources calculation function only check resources requirement of containers in pod.
I modify resource calculation in task which use function in common package to confirm it.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-706

I add test case to check calculation correctly.
Use make test to check.
### Questions:

### DIFF
--- a/pkg/common/resource.go
+++ b/pkg/common/resource.go
@@ -73,21 +73,22 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
 		podResource = Add(podResource, containerResource)
 	}
 
-	// vcore, mem compare between initcontainer and containers and replace
-	// For each resource, podResourceRequest = max(sum(Containers requirement), InitContainers requirement)
+	// each resource compare between initcontainer and sum of containers
+	// max(sum(Containers requirement), InitContainers requirement)
 	if pod.Spec.InitContainers != nil {
-		isNeedMoreResourceAndSet(pod, podResource)
+		checkInitContainerRequest(pod, podResource)
 	}
 
 	return podResource
 }
 
-func isNeedMoreResourceAndSet(pod *v1.Pod, containersResources *si.Resource) {
+func checkInitContainerRequest(pod *v1.Pod, containersResources *si.Resource) {
 	for _, c := range pod.Spec.InitContainers {
 		resourceList := c.Resources.Requests
 		ICResource := getResource(resourceList)
 		for resourceName, ICRequest := range ICResource.Resources {
 			containersRequests, exist := containersResources.Resources[resourceName]
+			// addtional resource request from init cont, add it to request.
 			if !exist {
 				containersResources.Resources[resourceName] = ICRequest
 				continue
--- a/pkg/common/resource.go
+++ b/pkg/common/resource.go
@@ -57,7 +57,7 @@ func (w *ResourceBuilder) Build() *si.Resource {
 // values, limits are ignored in the current setup.
 // BestEffort pods are scheduled using a minimum resource of 1MB only.
 func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
-	//var memory, vcore = int64(0), int64(0)
+	// var memory, vcore = int64(0), int64(0)
 	var podResource *si.Resource
 	// A QosBestEffort pod does not request any resources and thus cannot be
 	// scheduled. Handle a QosBestEffort pod by setting a tiny memory value.
@@ -85,14 +85,14 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
 func isNeedMoreResourceAndSet(pod *v1.Pod, containersResources *si.Resource) {
 	for _, c := range pod.Spec.InitContainers {
 		resourceList := c.Resources.Requests
-		initCResource := getResource(resourceList)
-		for resourceName, initCRequest := range initCResource.Resources {
+		ICResource := getResource(resourceList)
+		for resourceName, ICRequest := range ICResource.Resources {
 			containersRequests, exist := containersResources.Resources[resourceName]
 			if !exist {
 				continue
 			}
-			if initCRequest.GetValue() > containersRequests.GetValue() {
-				containersResources.Resources[resourceName] = initCRequest
+			if ICRequest.GetValue() > containersRequests.GetValue() {
+				containersResources.Resources[resourceName] = ICRequest
 			}
 		}
 	}
--- a/pkg/common/resource.go
+++ b/pkg/common/resource.go
@@ -82,21 +82,17 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
 }
 
 func IsNeedMoreResourceAndSet(pod *v1.Pod, containersResources *si.Resource) {
-	var initContainersResource *si.Resource
-
 	for _, c := range pod.Spec.InitContainers {
 		resourceList := c.Resources.Requests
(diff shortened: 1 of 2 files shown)


pkg/common/resource.go
```diff
@@ -72,9 +72,31 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
...
+
+	// vcore, mem compare between initcontainer and containers and replace(choose bigger one)
+	if pod.Spec.InitContainers != nil {
+		IsNeedMoreResourceAndSet(pod, podResource)
+	}
+
 	return podResource
 }
 
+func IsNeedMoreResourceAndSet(pod *v1.Pod, containersResources *si.Resource) {
```
kingamarton, line 84: We can keep this method private (`isNeedMoreResourceAndSet`) since it is used only from this place

pkg/common/resource.go
```diff
@@ -72,9 +72,31 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
...
+	for _, c := range pod.Spec.InitContainers {
+		resourceList := c.Resources.Requests
+		initCResource := getResource(resourceList)
+		for resouceName, v1 := range initCResource.Resources {
+			v2, exist := containersResources.Resources[resouceName]
```
kingamarton, line 89: Please use more self explanatory parameter naming than v1 and v2.

pkg/common/resource.go
```diff
@@ -72,9 +72,31 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
...
+			if !exist {
+				continue
+			}
+			if v1.GetValue() > v2.GetValue() {
+				containersResources.Resources[resouceName] = v1
```
kingamarton, line 94: v1 is the value of thee init resource. If the pod has init resource is it ok to just replace it, or we should add its value to the pods resources. How is the resource calculated in this case? For me it seems more logical to add them up instead of just replacing it. Please double check this.

kingamarton (changes requested):

pkg/common/resource.go
0yukali0: Ok

pkg/common/resource.go
0yukali0: Got it

pkg/common/resource.go
0yukali0: Thank you for advise, i will check object information in cache of core-scheduler to make sure it right.
This code idea is from k8s default scheduler.
k8s default scheduler make assigned node has enough allocated resource to handle max resource requirement in pod.
Calculation function: podResourceRequest = max(sum(podSpec.Containers), podSpec.InitContainers) + overHead

k8s doc of initcontainer describe requirement of resource usage.
related doc:
https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#resources

It show up in default scheduler predicate and score.
prefilter and filter use the calculation function to check if node has enough allocated resource for maximum of pod requirement.
related code:
https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/noderesources/fit.go#L135-L179
https://github.com/kubernetes/kubernetes/blob/eae87bfe7e9636694a5cb45a9863715b7b25435c/pkg/scheduler/framework/plugins/noderesources/resource_allocation.go#L108-L135

wilfred-s:
Looking at the code and the tests I think there is a problem with the calculation.
This is the base you gave:
```
	// pod
	// initcontainers
	// IC1{500mi, 1000m, 1}
	// IC2{1024mi, 2000m, 4}
	// containers
	// C1{4096mi, 2000m, 2}
	// C2{1024mi, 5000m, 2}
	// result is {4096mi, 5000m, 5}
```
There is a three step process:
1) Get the maximum for the init containers. They real serially which means that I just need to make sure that I can accomodate the largest request for each resource type.
That gives me in this case cpu: 1024mi, memory 2000m and gpu 4. In this specific case it all maps to IC2. IC1 has all types smaller than IC2.
2) Get the maximum usage for the real containers. They run in parallel which means I need to sum up all the resources used for all containers.
The total usage of all regular containers is:  cpu: 5120mi, memory 7000m and gpu 4.
3) Calculate the maximum number of resources needed in the startup phase for the init containers and the normal process:
```
init cpu: 1024mi, normal cpu: 5120mi --> cpu: 5120mi
init memory 2000m, normal memory 7000m --> memory 7000m
init gpu 4, normal gpu 4 --> gpu 4
```

The request for this specific pod passed on by the shim to the core is cpu: 5120mi, memory 7000m, gpu 4

If for example the init container IC2 would request 10000m for memory and 0 (zero) gpu the request would become: 5120mi, memory 10000m, gpu 4. CPU does not change: sum of C1 and C2, memory is now based on IC2 max, gpu remains at 4 as init and regular containers used the same.

0yukali0:
Thanks, i will fix it.

Wilfred Spiegelenburg ***@***.***> 於 2021年8月2日 週一 下午7:24寫道：

pkg/common/resource.go
```diff
@@ -72,9 +72,32 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
...
+	// vcore, mem compare between initcontainer and containers and replace
+	// For each resource, podResourceRequest = max(sum(Containers requirement), InitContainers requirement)
+	if pod.Spec.InitContainers != nil {
+		isNeedMoreResourceAndSet(pod, podResource)
+	}
+
 	return podResource
 }
 
+func isNeedMoreResourceAndSet(pod *v1.Pod, containersResources *si.Resource) {
```
wilfred-s, line 85: Can we rename this function? something like `checkInitContainerRequest()`

pkg/common/resource.go
```diff
@@ -72,9 +72,32 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
...
+	for _, c := range pod.Spec.InitContainers {
+		resourceList := c.Resources.Requests
+		ICResource := getResource(resourceList)
+		for resourceName, ICRequest := range ICResource.Resources {
+			containersRequests, exist := containersResources.Resources[resourceName]
```
wilfred-s, line 90: Does this mean we skip a request that is not defined on the container resources? That does not look right. 
cont = 1GB, 1cpu
init cont = 2GB, 1GPU
the result should be 2GB, 1cpu, 1GPU a merge of all requests.

wilfred-s (changes requested):

pkg/common/resource.go
0yukali0: ok, i wiil rename it in next commit

pkg/common/resource.go
0yukali0: I will add test for this situation in next commit.
In my thought, it will be like following
loop1 men:2GB > 1GB replace.
loop2 GPU is not found in cont request, additional request add.

0yukali0: Ok

0yukali0: Thanks @wilfred-s.
Additional resource from init cont  also add into request in next commit.

PR #285, opened 2021-07-22
[YUNIKORN-561]Some container tags are missing upon scheduler restart

HuangTing-Yao:
### What is this PR for?
Fix the AllocationTags missing when scheduler restart.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-561

### Questions:


pkg/common/si_helper.go
```diff
@@ -43,6 +43,10 @@ func createTagsForTask(pod *v1.Pod) map[string]string {
 	return tags
 }
 
+func CreateTagsForTask(pod *v1.Pod) map[string]string {
+	return createTagsForTask(pod)
+}
+
```
yangwwei, line 49: Looks good, for simplicity, I think we can just rename this function to `CreateTagsForTask` to expose it.

yangwwei (changes requested):

pkg/common/si_helper.go
HuangTing-Yao: Sure, done.

yangwwei (approved):
+1

PR #287, opened 2021-07-26
[YUNIKORN-765] the RBAC role created by yunikorn-rbac.yaml is not used by scheduler.yaml

chia7712:
### What is this PR for?
yunikorn-rbac.yaml creates a ClusterRoleBinding for account: yunikorn-admin. However, scheduler.yaml uses `default` account to query cluster so it encounters permission issue even if we have applied yunikorn-rbac.yaml according to docs.  

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-765

test it manually

### Questions:

### DIFF
--- a/deployments/scheduler/yunikorn-rbac.yaml
+++ b/deployments/scheduler/yunikorn-rbac.yaml
@@ -14,7 +14,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: yunikorn-admin
+  namespace: default
+---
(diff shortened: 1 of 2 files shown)


deployments/scheduler/yunikorn-rbac.yaml
```diff
@@ -14,7 +14,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+apiVersion: v1
```
chia7712, line 17: create account automatically to simply our life :)

HuangTing-Yao (approved):
LGTM

PR #288, opened 2021-07-29
[YUNIKORN-766] driver/executor get hang occasionally if the applicati…

chia7712:
…onId is defined

### What is this PR for?
The code used to get applicationId is shown below.
```go
// application ID can be defined in labels
for name, value := range pod.Labels {
// application ID can be defined as a label
if name == constants.LabelApplicationID
{ return value, nil }

// if a pod for spark already provided appID, reuse it
if name == constants.SparkLabelAppID { return value, nil }
}
```
The iteration order is random so it could get application id from 'spark-app-selector' rather than `applicationId` (even though `applicationId` is defined by users)

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-766

`TestGetApplicationIDFromPod`

### Questions:

### DIFF
--- a/pkg/common/utils/utils.go
+++ b/pkg/common/utils/utils.go
@@ -83,7 +83,6 @@ func GetQueueNameFromPod(pod *v1.Pod) string {
 }
 
 func GetApplicationIDFromPod(pod *v1.Pod) (string, error) {
-
 	// application ID can be defined in annotations
 	if value, found := pod.Annotations[constants.AnnotationApplicationID]; found {
 		return value, nil
--- a/pkg/common/utils/utils.go
+++ b/pkg/common/utils/utils.go
@@ -83,24 +83,22 @@ func GetQueueNameFromPod(pod *v1.Pod) string {
 }
 
 func GetApplicationIDFromPod(pod *v1.Pod) (string, error) {
+
 	// application ID can be defined in annotations
-	for name, value := range pod.Annotations {
-		if name == constants.AnnotationApplicationID {
-			return value, nil
-		}
+	if value, found := pod.Annotations[constants.AnnotationApplicationID]; found {
+		return value, nil
 	}
 
 	// application ID can be defined in labels
-	for name, value := range pod.Labels {
-		// application ID can be defined as a label
-		if name == constants.LabelApplicationID {
-			return value, nil
-		}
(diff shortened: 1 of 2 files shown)


pkg/common/utils/utils.go
```diff
@@ -83,24 +83,22 @@ func GetQueueNameFromPod(pod *v1.Pod) string {
...
 	}
 
 	// application ID can be defined in labels
-	for name, value := range pod.Labels {
-		// application ID can be defined as a label
-		if name == constants.LabelApplicationID {
-			return value, nil
-		}
+	if value, found := pod.Labels[constants.LabelApplicationID]; found {
+		fmt.Printf("[CHIA] name: %+v value: %+v\n", constants.LabelApplicationID, value)
```
yangwwei, line 94: is this the log message for debug? can we remove it?

pkg/common/utils/utils.go
```diff
@@ -83,24 +83,22 @@ func GetQueueNameFromPod(pod *v1.Pod) string {
...
+		return value, nil
+	}
 
-		// if a pod for spark already provided appID, reuse it
-		if name == constants.SparkLabelAppID {
-			return value, nil
-		}
+	// if a pod for spark already provided appID, reuse it
+	if value, found := pod.Labels[constants.SparkLabelAppID]; found {
+		fmt.Printf("[CHIA] name: %+v value: %+v\n", constants.SparkLabelAppID, value)
```
yangwwei, line 100: redundant log message

yangwwei (changes requested):
hi @chia7712  good catch, and thanks for the fix. I have some minor comments about logs, pls check, thank you!

chia7712:
>  I have some minor comments about logs, pls check.
pls fix the lint error as well, thank you!

@yangwwei thanks for quick review. will copy that.

chia7712:
`FAIL	github.com/apache/incubator-yunikorn-k8shim/pkg/cache	24.389s`

it passes on my local. will rerun this PR

yangwwei (approved):

PR #290, opened 2021-07-30
[YUNIKORN-770] Remove unnecessary DEBUG log while removing a pod from the cache

chia7712:
### What is this PR for?
```go
	n, ok := cache.nodesMap[pod.Spec.NodeName]
	if !ok {
		return fmt.Errorf("node %v is not found", pod.Spec.NodeName)
	}
	if err := n.RemovePod(pod); err != nil {
		return err
	}
```
It can causes DEBUG log("node  is not found") if node name is an empty string. For example: the pod status is updated (according to outstanding requests) before it is assumed to be run on a node.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-770

`TestRemovePodWithoutNodeName`

### Questions:

### DIFF
--- a/pkg/cache/external/scheduler_cache.go
+++ b/pkg/cache/external/scheduler_cache.go
@@ -245,12 +245,14 @@ func (cache *SchedulerCache) RemovePod(pod *v1.Pod) error {
 }
 
 func (cache *SchedulerCache) removePod(pod *v1.Pod) error {
-	n, ok := cache.nodesMap[pod.Spec.NodeName]
-	if !ok {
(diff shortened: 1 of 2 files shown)


yangwwei (approved):
+1 to remove this, this is being a bit confusing in the log sometimes.
thanks @chia7712  for fixing it.

PR #292, opened 2021-08-02
[YUNIKORN-744]Call healthcheck REST API in e2e tests

HuangTing-Yao:
### What is this PR for?
Call and check the yunikorn HealthCheck API in all kind of e2e test.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-744

### Questions:

### DIFF
--- a/test/e2e/app/app_test.go
+++ b/test/e2e/app/app_test.go
@@ -116,7 +116,9 @@ var _ = ginkgo.Describe("App", func() {
 		ginkgo.AfterEach(func() {
 			// call the healthCheck api to check scheduler health
 			ginkgo.By("Check Yunikorn's health")
-			yunikorn.HealthCheck()
+			checks, err := yunikorn.GetFailedHealthChecks()
+			gomega.Ω(err).NotTo(gomega.HaveOccurred())
+			gomega.Ω(checks).To(gomega.Equal(""), checks)
 		})
 
 		ginkgo.AfterSuite(func() {
--- a/test/e2e/app/app_test.go
+++ b/test/e2e/app/app_test.go
@@ -113,14 +113,10 @@ var _ = ginkgo.Describe("App", func() {
 			gomega.Ω(appCRD.Spec.TaskGroups[0].Tolerations[0].Effect).To(gomega.Equal(v1.TaintEffectNoSchedule))
 		})
 
-		ginkgo.It("Verify the Yunikorn Scheduler healthy", func() {
-			ginkgo.By("Call the HealthCheck API")
-			healthCheck, err := yunikorn.GetHealthCheck()
-			gomega.Ω(err).NotTo(gomega.HaveOccurred())
-			gomega.Ω(healthCheck.Healthy).To(gomega.BeTrue())
-			for _, check := range healthCheck.HealthChecks {
-				gomega.Ω(check.Succeeded).To(gomega.BeTrue())
-			}
+		ginkgo.AfterEach(func() {
+			// call the healthCheck api to check scheduler health
+			ginkgo.By("Check Yunikorn's health")
+			yunikorn.HealthCheck()
 		})
 
 		ginkgo.AfterSuite(func() {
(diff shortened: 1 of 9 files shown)


test/e2e/app/app_test.go
```diff
@@ -113,6 +113,16 @@ var _ = ginkgo.Describe("App", func() {
...
 
+		ginkgo.It("Verify the Yunikorn Scheduler healthy", func() {
+			ginkgo.By("Call the HealthCheck API")
+			healthCheck, err := yunikorn.GetHealthCheck()
+			gomega.Ω(err).NotTo(gomega.HaveOccurred())
+			gomega.Ω(healthCheck.Healthy).To(gomega.BeTrue())
+			for _, check := range healthCheck.HealthChecks {
+				gomega.Ω(check.Succeeded).To(gomega.BeTrue())
+			}
+		})
```
yangwwei, line 124: I am not very familiar with ginkgo framework, but since this is the same function call for every test, can we move this to a utility function and reuse the code everywhere? If that is possible.

test/e2e/app/app_test.go
```diff
@@ -113,6 +113,16 @@ var _ = ginkgo.Describe("App", func() {
...
 		})
 
+		ginkgo.It("Verify the Yunikorn Scheduler healthy", func() {
+			ginkgo.By("Call the HealthCheck API")
+			healthCheck, err := yunikorn.GetHealthCheck()
+			gomega.Ω(err).NotTo(gomega.HaveOccurred())
+			gomega.Ω(healthCheck.Healthy).To(gomega.BeTrue())
+			for _, check := range healthCheck.HealthChecks {
+				gomega.Ω(check.Succeeded).To(gomega.BeTrue())
+			}
```
yangwwei, line 123: if line 120 returns true, I do not think we need to double-check every check.
If any check fails, the overall result should be false.
however, one thing not very clear to me is, when this fails, can we print the detailed messages in the log to help us debug?

yangwwei (changes requested):

test/e2e/framework/helpers/yunikorn/rest_api_utils.go
```diff
@@ -224,6 +224,17 @@ func isRootSched(policy string) wait.ConditionFunc {
 	}
 }
 
+func GetHealthCheck() (dao.SchedulerHealthDAOInfo, error) {
+	restClient := RClient{}
```
ronazhan, line 228: The rest calls are generally designed as methods on the type RClient. See [GetApps()](https://github.com/apache/incubator-yunikorn-k8shim/blob/4d28705e4723c27178c7f8d31751ecea2150398d/test/e2e/framework/helpers/yunikorn/rest_api_utils.go#L126), or most other methods in the rest_api_utils.go for an example

test/e2e/app/app_test.go
ronazhan: Instead of making this a standalone test for each suite, include the health check as part of the AfterEach(). Should not be much overhead for just a rest call, and this will make it easier to pinpoint after which test Yunikorn scheduler entered unhealthy state. 

Each suite can have an AfterEach, and the AfterEach code will run after each test in the suite.

test/e2e/app/app_test.go
ronazhan: On failed assertions, gomega can also print a formatted log: https://onsi.github.io/gomega/#annotating-assertions

yangwwei:
Looks much better. @ronazhan can you pls give another look at this PR?

HuangTing-Yao:
Hi @ronazhan I have updated this PR, Could you take a look at this PR?

ronazhan:
lgtm. @yangwwei want to take another pass?

yangwwei:
looks good as well, thank you @HuangTing-Yao for getting this done, thank you @ronazhan  for the review!

yangwwei (approved):

PR #296, opened 2021-08-19
[YUNIKORN-805] Fix race condition on the TestMode flag

0yukali0:
### What is this PR for?
Avoid data race which R/W in same time.
Add getting TestMode in config object with R lock

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-805

make test with Makefile

### Questions:

### DIFF
--- a/pkg/shim/scheduler.go
+++ b/pkg/shim/scheduler.go
@@ -19,6 +19,7 @@
 package main
 
 import (
+	"os"
 	"sync"
 	"time"
 
@@ -145,6 +146,9 @@ func (ss *KubernetesShim) register(e *fsm.Event) {
 
 func (ss *KubernetesShim) handleSchedulerFailure(e *fsm.Event) {
 	ss.stop()
+	if !conf.GetSchedulerConf().IsTestMode() {
+		os.Exit(1)
+	}
 }
 
 func (ss *KubernetesShim) triggerSchedulerStateRecovery(e *fsm.Event) {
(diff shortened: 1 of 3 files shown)


pkg/shim/scheduler.go
```diff
@@ -145,6 +146,10 @@ func (ss *KubernetesShim) register(e *fsm.Event) {
 
 func (ss *KubernetesShim) handleSchedulerFailure(e *fsm.Event) {
 	ss.stop()
+	// testmode will be true when mock scheduler intailize
+	if configured := conf.GetSchedulerConf(); !configured.IsTestMode() {
```
yangwwei, line 150: this should be 
```
if !conf.GetSchedulerConf().IsTestMode() {
  ...
}
```
conf.GetSchedulerConf() should never be nil.

yangwwei (changes requested):
the change is incompleted, @0yukali0 could u pls find all the direct references of `conf.TestMode` and make sure they all use `conf.IsTestMode()`?

pkg/shim/scheduler.go
0yukali0: ok

yangwwei (approved):
+1

PR #302, opened 2021-09-13
[YUNIKORN-841] Data race in TestResumingStateTransitions()

HuangTing-Yao:
### What is this PR for?
Solve the data race from UT

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-841

### Questions:

### DIFF
--- a/pkg/cache/placeholder_manager.go
+++ b/pkg/cache/placeholder_manager.go
@@ -43,7 +43,7 @@ type PlaceholderManager struct {
 	stopChan   chan struct{}
 	running    atomic.Value
 	// a simple mutex will do we do not have separate read and write paths
-	sync.Mutex
+	sync.RWMutex
 }
 
 var placeholderMgr *PlaceholderManager
@@ -61,8 +61,11 @@ func NewPlaceholderManager(clients *client.Clients) *PlaceholderManager {
 }
 
 func getPlaceholderManager() *PlaceholderManager {
-	placeholderMgr.Lock()
-	defer placeholderMgr.Unlock()
+	if placeholderMgr == nil {
+		return nil
+	}
+	placeholderMgr.RLock()
+	defer placeholderMgr.RUnlock()
 	return placeholderMgr
 }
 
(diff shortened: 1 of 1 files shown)


pkg/cache/placeholder_manager.go
```diff
@@ -61,6 +61,8 @@ func NewPlaceholderManager(clients *client.Clients) *PlaceholderManager {
 }
 
 func getPlaceholderManager() *PlaceholderManager {
+	placeholderMgr.Lock()
+	defer placeholderMgr.Unlock()
```
yangwwei, line 65: add nil check in case `placeholderMgr` is not initiated while calling `getPlaceholderManager()`?
and RLock should be sufficient

yangwwei (changes requested):

pkg/cache/placeholder_manager.go
```diff
@@ -61,6 +61,11 @@ func NewPlaceholderManager(clients *client.Clients) *PlaceholderManager {
 }
 
 func getPlaceholderManager() *PlaceholderManager {
+	placeholderMgr.RLock()
+	defer placeholderMgr.RUnlock()
+	if placeholderMgr == nil {
+		return nil
```
yangwwei, line 67: need to put the nil check upfront

```
if placeholderMgr == nil {
		return nil
}
placeholderMgr.RLock()
defer placeholderMgr.RUnlock()
...

```

yangwwei (changes requested):

pkg/cache/placeholder_manager.go
HuangTing-Yao: but it will cause other data race, when we check the nil, need to read the placeholderMgr.

yangwwei (approved):
+1. thanks for fixing this @HuangTing-Yao

PR #305, opened 2021-09-22
[YUNIKORN-858] Pass node labels as node attributes for node sorting

chenya-zhang:
### What is this PR for?

Node labels need to be passed into "YuniKorn Core" via "YuniKorn Scheduler Interface" (YK SI) as node attributes. It then can be considered during node sorting.

In particular, to keep the protobuf value type of node attributes still string, we plan to marshal node labels (map[string]string) into a JSON string before passing to YK SI.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [X] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-858

### Questions:

### DIFF
--- a/pkg/cache/nodes.go
+++ b/pkg/cache/nodes.go
@@ -19,6 +19,7 @@
 package cache
 
 import (
+	"encoding/json"
 	"fmt"
 	"sync"
 
@@ -100,10 +101,18 @@ func (nc *schedulerNodes) addAndReportNode(node *v1.Node, reportNode bool) {
 
 	// add node to nodes map
 	if _, ok := nc.nodesMap[node.Name]; !ok {
+		nodeLabels, err := json.Marshal(node.Labels)
+		if err != nil {
+			log.Logger().Error("failed to marshall node labels to json", zap.Error(err))
+			return
+		}
+
 		log.Logger().Info("adding node to context",
 			zap.String("nodeName", node.Name),
+			zap.String("nodeLabels", string(nodeLabels)),
 			zap.Bool("schedulable", !node.Spec.Unschedulable))
-		newNode := newSchedulerNode(node.Name, string(node.UID),
+
+		newNode := newSchedulerNode(node.Name, string(node.UID), string(nodeLabels),
 			common.GetNodeResource(&node.Status), nc.proxy, !node.Spec.Unschedulable)
 		nc.nodesMap[node.Name] = newNode
 	}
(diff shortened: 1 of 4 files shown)


pkg/cache/nodes.go
```diff
@@ -100,10 +101,18 @@ func (nc *schedulerNodes) addAndReportNode(node *v1.Node, reportNode bool) {
 
 	// add node to nodes map
 	if _, ok := nc.nodesMap[node.Name]; !ok {
+		nodeLabels, err := json.Marshal(node.Labels)
```
yangwwei, line 104: what if `node.Labels` is nil?
can we  also add a UT coverage for the case where node has no labels defined?

yangwwei (changes requested):

pkg/cache/nodes.go
```diff
@@ -100,10 +101,18 @@ func (nc *schedulerNodes) addAndReportNode(node *v1.Node, reportNode bool) {
...
+		if err != nil {
+			log.Logger().Error("failed to marshall node labels to json", zap.Error(err))
+			return
```
yangwwei, line 107: I don't think we should return here, even we can't marshall node labels, we should still report the node
otherwise the node will be skipped adding to the scheduler, that might not be something we want

yangwwei (changes requested):

pkg/cache/nodes.go
chenya-zhang: Yeah, good catch. 

Since we are only using node labels to support node affinity with a **_preferred_** scheduling policy, it makes sense to not skip this node even if labels can't be marshaled.

pkg/cache/nodes.go
yangwwei: hi @chenya-zhang can you pls also take a look at this comment pls? thx

chenya-zhang: Thanks for the reminder @yangwwei! Sorry I failed to see this comment last time.

If node.Labels is nil, JSON marshal will encode it as the "null" JSON value. We will check string "null" when decoding it.
Our existing UT in https://github.com/apache/incubator-yunikorn-k8shim/blob/84f52b7d72cdac1bc4904cf5a631617070225d29/pkg/cache/nodes_test.go#L76 should cover the case if a node has no labels. 

On the other hand, your comment reminds me to double check what if we fail to marshal. Unfortunately, JSON marshal will return a nil along with an err. This nil will be problematic if passing to string() directly. In this case, I assign a 0 byte to the JSON format of node labels, which is an empty string for us to check in YK core.

yangwwei (approved):
LGTM

PR #321, opened 2021-10-23
YUNIKORN-906: Shim side changes of YUNIKORN-337

manirajv06:
### What is this PR for?
https://issues.apache.org/jira/browse/YUNIKORN-906

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-906

### Questions:

### DIFF
--- a/pkg/cache/node.go
+++ b/pkg/cache/node.go
@@ -150,7 +150,7 @@ func (n *SchedulerNode) handleNodeRecovery(event *fsm.Event) {
 		zap.Bool("schedulable", n.schedulable))
 
 	allocRequest := &si.AllocationRequest{
-		RmID:     conf.GetSchedulerConf().ClusterID,
+		RmID: conf.GetSchedulerConf().ClusterID,
 	}
 	nodeRequest := &si.NodeRequest{
 		Nodes: []*si.NodeInfo{
@@ -187,7 +187,7 @@ func (n *SchedulerNode) handleDrainNode(event *fsm.Event) {
 		zap.String("nodeID", n.name))
 
 	allocRequest := &si.AllocationRequest{
-		RmID:     conf.GetSchedulerConf().ClusterID,
+		RmID: conf.GetSchedulerConf().ClusterID,
 	}
 	nodeRequest := &si.NodeRequest{
 		Nodes: []*si.NodeInfo{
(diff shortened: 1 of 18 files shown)


pkg/cache/node.go
```diff
@@ -149,10 +149,13 @@ func (n *SchedulerNode) handleNodeRecovery(event *fsm.Event) {
 		zap.String("nodeID", n.name),
 		zap.Bool("schedulable", n.schedulable))
 
-	request := &si.UpdateRequest{
+	allocRequest := &si.AllocationRequest{
 		Asks:     nil,
 		Releases: nil,
-		NewSchedulableNodes: []*si.NewNodeInfo{
+		RmID:     conf.GetSchedulerConf().ClusterID,
+	}
```
wilfred-s, line 156: Sending this would be a nop in the core. The core checks for the length of Asks and Releases and if they are zero (0) we just return without doing anything. See `handleRMUpdateAllocationEvent()`

pkg/cache/node.go
```diff
@@ -163,29 +166,38 @@ func (n *SchedulerNode) handleNodeRecovery(event *fsm.Event) {
...
 	log.Logger().Info("node enters draining mode",
 		zap.String("nodeID", n.name))
 
-	request := &si.UpdateRequest{
+	allocRequest := &si.AllocationRequest{
 		Asks:     nil,
 		Releases: nil,
-		UpdatedNodes: []*si.UpdateNodeInfo{
+		RmID:     conf.GetSchedulerConf().ClusterID,
+	}
```
wilfred-s, line 195: Same here: this is a nop on the core side

pkg/cache/node.go
```diff
@@ -197,23 +209,32 @@ func (n *SchedulerNode) handleDrainNode(event *fsm.Event) {
...
 	log.Logger().Info("restore node from draining mode",
 		zap.String("nodeID", n.name))
 
-	request := &si.UpdateRequest{
+	allocRequest := &si.AllocationRequest{
 		Asks:     nil,
 		Releases: nil,
-		UpdatedNodes: []*si.UpdateNodeInfo{
+		RmID:     conf.GetSchedulerConf().ClusterID,
+	}
```
wilfred-s, line 232: Same here: this is a nop on the core side

wilfred-s (changes requested):
Beside the 3 calls that are nop's in the core this looks good.
Please check those calls and if they are needed to pass tests etc we should fix the tests but we should not leave the nop calls in the production code.

pkg/cache/node.go
manirajv06: Just retained those request params as it was there before. Removed the same now.

wilfred-s (approved):
LGTM, waiting for a build to pass before we commit the change
depends on #327

PR #326, opened 2021-11-08
[YUNIKORN-932] Added some extra log message

kingamarton:
### What is this PR for?
Added a few new log messages for the case when there is no outstanding requests. 
Also added a log message for the recovery part, to check if we got an answer from the api server or not

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-932

### Questions:

### DIFF
--- a/pkg/appmgmt/general/general.go
+++ b/pkg/appmgmt/general/general.go
@@ -338,13 +338,13 @@ func (os *Manager) ListApplications() (map[string]interfaces.ApplicationMetadata
 		// this means the pod is already scheduled by scheduler for an existing app
 		if utils.GeneralPodFilter(pod) && utils.IsAssignedPod(pod) {
 			if meta, ok := os.getAppMetadata(pod); ok {
-				podsRecovered += 1
+				podsRecovered++
 				log.Logger().Debug("Adding appID as recovery candidate", zap.String("appID", meta.ApplicationID))
 				if _, exist := existingApps[meta.ApplicationID]; !exist {
 					existingApps[meta.ApplicationID] = meta
 				}
 			} else {
-				podsWithoutMetaData += 1
+				podsWithoutMetaData++
 			}
 		}
 	}
--- a/pkg/appmgmt/general/general.go
+++ b/pkg/appmgmt/general/general.go
@@ -320,13 +320,14 @@ func (os *Manager) deletePod(obj interface{}) {
 }
 
 func (os *Manager) ListApplications() (map[string]interfaces.ApplicationMetadata, error) {
+	log.Logger().Info("Listing recoverable apps")
 	// list all pods on this cluster
 	slt := labels.NewSelector()
 	appPods, err := os.apiProvider.GetAPIs().PodInformer.Lister().List(slt)
 	if err != nil {
 		return nil, err
 	}
-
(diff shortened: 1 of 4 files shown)


pkg/appmgmt/general/general.go
```diff
@@ -342,6 +343,7 @@ func (os *Manager) ListApplications() (map[string]interfaces.ApplicationMetadata
 			}
 		}
 	}
+	log.Logger().Info("Listing recoverable apps finished", zap.Int("nr of recoverable apps", len(existingApps)))
```
wilfred-s, line 346: "Listing recoverable apps finished" --> "application recovery statistics"
adding the two values mentioned above to the number of apps

pkg/shim/scheduler.go
```diff
@@ -256,7 +265,9 @@ func (ss *KubernetesShim) canHandle(se events.SchedulerEvent) bool {
 func (ss *KubernetesShim) schedule() {
 	apps := ss.context.SelectApplications(nil)
 	for _, app := range apps {
-		app.Schedule()
+		if app.Schedule() {
+			ss.outstandingAppsFound = true
+		}
```
wilfred-s, line 270: We should be able to simplify this to:
```
ss.outstandingAppsFound = app.Schedule()
```

wilfred-s (changes requested):
minor changes mainly for text in the log messages

pkg/shim/scheduler.go
kingamarton: We cannot simplify it because if in a cycle we found some outstanding apps, but in the next one not, we don't want to owerride it.

pkg/appmgmt/general/general.go
```diff
@@ -320,13 +320,14 @@ func (os *Manager) deletePod(obj interface{}) {
...
 	slt := labels.NewSelector()
 	appPods, err := os.apiProvider.GetAPIs().PodInformer.Lister().List(slt)
 	if err != nil {
 		return nil, err
 	}
-
+	log.Logger().Info("Got the list of pods from the api server", zap.Int("nr of pods", len(appPods)))
 	// get existing apps
 	existingApps := make(map[string]interfaces.ApplicationMetadata)
 	for _, pod := range appPods {
```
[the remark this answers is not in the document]
kingamarton: Added 3 extra fields: total pod nr, nr of pods without app metadata, nr of pods linked to recoverable apps

pkg/shim/scheduler.go
kingamarton: We cannot simplify it because if in a cycle we found some outstanding apps, but in the next one not, we don't want to owerride it.

wilfred-s: ok

wilfred-s (approved):
LGTM, waiting for a build to pass before we commit the change
depends on #327

kingamarton:
Thank you @wilfred-s for the review! I am going to commit the changes, since we have the successful build

PR #329, opened 2021-11-16
[YUNIKORN-508] fix log collection path

wilfred-s:
### What is this PR for?
The e2e tests generate a log for analysis. The log directory is created
as part of the test run. The default path set in the code must be a
creatable or pre-existing path.
If the directory exists it is re-used. Files inside the directory will
be truncated and re-written when needed.
Changing default to: /tmp/e2e-test-reports

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
YUNIKORN-508 Fix/remove log collection error from upstream runs

NA, removes messages about non existing directories in e2e test

### DIFF
--- a/test/e2e/framework/configmanager/config.go
+++ b/test/e2e/framework/configmanager/config.go
@@ -23,7 +23,7 @@ import (
 	"time"
 )
 
-// YuniKornTestConfigType holds all of the configurable elements of the testsuite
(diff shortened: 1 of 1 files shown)


craigcondit (approved):
LGTM. FYI, unit tests passed on re-run and E2e test failures don't seem related to this change. Will merge shortly.

PR #336, opened 2021-12-08
[YUNIKORN-971] Implement YuniKorn as a Kubernetes scheduler plugin.

craigcondit:
### What is this PR for?
This PR implements a new build mode for YuniKorn which allows it to run as a set of plugins on top of the default Kubernetes scheduler. This should allow for closer behavior to the default scheduler as well as the ability to bypass YuniKorn-specific functionality for pods which are not annotated with an `applicationId` label.

This PR makes no functional changes to the default YuniKorn build, but adds new Makefile targets to support build / running YuniKorn as a scheduler plugin:

- `build_plugin` Build the plugin code.
- `run_plugin` Run the plugin code locally.
- `plugin_image` Create a Docker image from the plugin code. this should mostly be a drop-in replacement for the default YuniKorn image.

This feature should be considered `experimental` at this stage. Hopefully we can provide it as a tech-preview for 1.0.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-971

Some unit tests added for configuration, remainder is covered via e2e tests (all passing).

### Questions:

### DIFF
--- a/pkg/cache/external/scheduler_cache.go
+++ b/pkg/cache/external/scheduler_cache.go
@@ -36,23 +36,23 @@ import (
 // nodes are cached in the form of de-scheduler nodeInfo, instead of re-creating all nodes info from scratch,
 // we replicate nodes info from de-scheduler, in order to re-use predicates functions.
 type SchedulerCache struct {
-	// node name to NodeInfo map
-	nodesMap map[string]*framework.NodeInfo
-	podsMap  map[string]*v1.Pod
-	// this is a map of assumed pods,
-	// the value indicates if a pod volumes are all bound
-	assumedPods map[string]bool
-	lock        sync.RWMutex
-	// client APIs
-	clients *client.Clients
+	nodesMap              map[string]*framework.NodeInfo // node name to NodeInfo map
+	podsMap               map[string]*v1.Pod
+	assumedPods           map[string]bool   // map of assumed pods, value indicates if pod volumes are all bound
+	pendingAllocations    map[string]string // map of pod to node ID, presence indicates a pending allocation for scheduler
+	inProgressAllocations map[string]string // map of pod to node ID, presence indicates an in-process allocation for scheduler
+	lock                  sync.RWMutex
+	clients               *client.Clients // client APIs
 }
 
 func NewSchedulerCache(clients *client.Clients) *SchedulerCache {
 	cache := &SchedulerCache{
-		nodesMap:    make(map[string]*framework.NodeInfo),
-		podsMap:     make(map[string]*v1.Pod),
-		assumedPods: make(map[string]bool),
-		clients:     clients,
+		nodesMap:              make(map[string]*framework.NodeInfo),
+		podsMap:               make(map[string]*v1.Pod),
+		assumedPods:           make(map[string]bool),
+		pendingAllocations:    make(map[string]string),
+		inProgressAllocations: make(map[string]string),
+		clients:               clients,
 	}
 	return cache
 }
@@ -130,6 +130,45 @@ func (cache *SchedulerCache) removeNode(node *v1.Node) error {
 
 	delete(cache.nodesMap, node.Name)
 	return nil
+}
+
+func (cache *SchedulerCache) AddPendingPodAllocation(podKey string, nodeID string) {
+	cache.lock.Lock()
+	defer cache.lock.Unlock()
+	delete(cache.inProgressAllocations, podKey)
+	cache.pendingAllocations[podKey] = nodeID
+}
+
+func (cache *SchedulerCache) RemovePendingPodAllocation(podKey string) {
+	cache.lock.Lock()
+	defer cache.lock.Unlock()
+	delete(cache.pendingAllocations, podKey)
+	delete(cache.inProgressAllocations, podKey)
+}
+
+func (cache *SchedulerCache) GetPendingPodAllocation(podKey string) (nodeID string, ok bool) {
+	cache.lock.RLock()
+	defer cache.lock.RUnlock()
+	res, ok := cache.pendingAllocations[podKey]
+	return res, ok
+}
+
+func (cache *SchedulerCache) GetInProgressPodAllocation(podKey string) (nodeID string, ok bool) {
+	cache.lock.RLock()
+	defer cache.lock.RUnlock()
+	res, ok := cache.inProgressAllocations[podKey]
+	return res, ok
+}
+
+func (cache *SchedulerCache) StartPodAllocation(podKey string, nodeID string) bool {
+	cache.lock.Lock()
+	defer cache.lock.Unlock()
+	expectedNodeID, ok := cache.pendingAllocations[podKey]
+	if ok && expectedNodeID == nodeID {
+		delete(cache.pendingAllocations, podKey)
+		cache.inProgressAllocations[podKey] = nodeID
+	}
+	return ok
 }
 
 // return if pod is assumed in cache, avoid nil
@@ -249,6 +288,8 @@ func (cache *SchedulerCache) removePod(pod *v1.Pod) error {
 			return err
 		}
 	}
+	delete(cache.pendingAllocations, string(pod.UID))
+	delete(cache.inProgressAllocations, string(pod.UID))
 	return nil
 }
 
--- a/pkg/cache/external/scheduler_cache.go
+++ b/pkg/cache/external/scheduler_cache.go
@@ -32,9 +32,17 @@ import (
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/log"
 )
 
-// scheduler cache maintains some critical information about nodes and pods used for scheduling
-// nodes are cached in the form of de-scheduler nodeInfo, instead of re-creating all nodes info from scratch,
+// SchedulerCache maintains some critical information about nodes and pods used for scheduling.
+// Nodes are cached in the form of de-scheduler nodeInfo. Instead of re-creating all nodes info from scratch,
 // we replicate nodes info from de-scheduler, in order to re-use predicates functions.
+//
+// When running YuniKorn as a scheduler plugin, we also track pod allocations that YuniKorn has decided upon, but which
+// have not yet been fulfilled by the default scheduler. This tracking is needed to ensure that we pass along
+// allocations to the default scheduler once (and only) once. Allocations can be in one of two states, either pending or
+// in-progress. A pending allocation is one which has been decided upon by YuniKorn but has not yet been communicated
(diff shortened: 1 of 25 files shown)


craigcondit:
Rebased on latest master (post 0.12.1).

craigcondit:
@wilfred-s, @yangwwei, can you review?

pkg/cache/external/scheduler_cache.go
```diff
@@ -36,23 +36,23 @@ import (
...
-	// the value indicates if a pod volumes are all bound
-	assumedPods map[string]bool
-	lock        sync.RWMutex
-	// client APIs
-	clients *client.Clients
+	nodesMap              map[string]*framework.NodeInfo // node name to NodeInfo map
+	podsMap               map[string]*v1.Pod
+	assumedPods           map[string]bool   // map of assumed pods, value indicates if pod volumes are all bound
+	pendingAllocations    map[string]string // map of pod to node ID, presence indicates a pending allocation for scheduler
+	inProgressAllocations map[string]string // map of pod to node ID, presence indicates an in-process allocation for scheduler
```
pbacsko, line 43: We need some extra comment which explains what is the difference between a "pending" and an "in progress" allocation. Just by looking at the code, it's not straightforward which comes first and what is the lifecycle of the pods.

pkg/cache/task.go
```diff
@@ -440,6 +460,23 @@ func (task *Task) beforeTaskAllocated(event *fsm.Event) {
 }
 
 func (task *Task) postTaskBound(event *fsm.Event) {
+	if task.pluginMode {
+		// when the pod is scheduling by yunikorn, it is moved to the default-scheduler's
+		// unschedulable queue, if nothing changes, the pod will be staying in the unschedulable
+		// queue for unschedulableQTimeInterval long (default 1 minute). hence, we are updating
+		// the pod status explicitly, when there is a status change, the default scheduler will
+		// move the pod back to the active queue immediately.
+		podCopy := task.pod.DeepCopy()
```
pbacsko, line 469: Can we reason about the performance effects of this copy? If we do this all the time (and that's what seems to be happening), we end up creating extra objects which affects GC time. Maybe not a big deal, just something to think about.

pkg/cache/external/scheduler_cache.go
```diff
@@ -132,6 +132,45 @@ func (cache *SchedulerCache) removeNode(node *v1.Node) error {
...
+	defer cache.lock.RUnlock()
+	res, ok := cache.inProgressAllocations[podKey]
+	return res, ok
+}
+
+func (cache *SchedulerCache) StartPodAllocation(podKey string, nodeID string) bool {
+	cache.lock.Lock()
+	defer cache.lock.Unlock()
+	expectedNodeID, ok := cache.pendingAllocations[podKey]
+	if ok && expectedNodeID == nodeID {
```
pbacsko, line 167: Ok, from what I understand `expectedNodeID` comes from YK itself, and `nodeID` from the default scheduler. Questions:
* How often do these two mismatch?
* What happens if we return `false` here?
* Can YK and the def. scheduler end up repeatedly allocating to different nodes? 

Another thing is that do we even need to consider YK's allocation? Because in `task.postTaskAllocated()`, the original code path calls `KubeClient.Bind()` but in plugin mode, we delegate this operation as I can see.

pkg/schedulerplugin/scheduler_plugin.go
```diff
@@ -0,0 +1,192 @@
...
+					zap.String("taskID", task.GetTaskID()))
+				sp.context.RemovePendingPodAllocation(string(pod.UID))
+				dispatcher.Dispatch(cache.NewRejectTaskEvent(app.GetApplicationID(), task.GetTaskID(),
+					fmt.Sprintf("task %s rejected by scheduler", task.GetTaskID())))
+				return framework.NewStatus(framework.Unschedulable, "Pod is not ready for scheduling")
+			}
+
+			nodeID, ok := sp.context.GetPendingPodAllocation(string(pod.UID))
+			if !ok {
+				nodeID = "<none>"
```
pbacsko, line 92: Is this value used for anything?

pkg/cache/external/scheduler_cache.go
```diff
@@ -132,6 +132,45 @@ func (cache *SchedulerCache) removeNode(node *v1.Node) error {
 	return nil
 }
 
+func (cache *SchedulerCache) AddPendingPodAllocation(podKey string, nodeID string) {
+	cache.lock.Lock()
+	defer cache.lock.Unlock()
+	delete(cache.inProgressAllocations, podKey)
```
pbacsko, line 138: Is this delete necessary? Under what circumstances can we have an entry?

As I can see it, `inProgressAllocations` is populated from `StartPodAllocation()` but that already requires that pending has an entry which we just add in line 139. Am I missing something here?

Is it a bug if we already have something in `inProgressAllocations` at this point for the same `podKey`?

pkg/cache/external/scheduler_cache.go
craigcondit: A pending allocation is one which YuniKorn has decided on, but which has not yet been communicated to the default scheduler via the scheduling framework (the PreFilter()/Filter() calls in the plugin). An in-progress allocation is one which HAS been communicated to the default scheduler via PreFilter() / Filter() but which we have not yet got confirmation of the Pod being bound (PostBind() in the plugin has not yet been called). This is used to track state to ensure that we don't reprocess the same Pod or leak in the case where the default scheduler is unable to bind the pod for some reason.

pkg/cache/task.go
craigcondit: This cost is negligible, because the copy happens very infrequently (typically once per pod), when YuniKorn has decided to allocate the pod. We update the pod status once at this point to force the default scheduler logic to re-consider the pod for scheduling. It will very shortly call PreFilter() / Filter() and push the Pod through the scheduling cycle.

pkg/cache/external/scheduler_cache.go
craigcondit: This code is used to verify that the requested node (by the default scheduler) is the one that YuniKorn decided to allocate the pod to. Returning `false` here will result in the Filter() call in the plugin returning `Unschedulable` to the default scheduler for this pod / node combination, and a subsequent Filter() call with a matching node will return `Success` instead. Effectively, we use this code to veto node allocations other than the node that YK has decided on.

The bind is deferred to the default scheduler logic once Filter() succeeds, which is why we don't handle the bind logic in postTaskAllocated().

pkg/schedulerplugin/scheduler_plugin.go
craigcondit: in the logging message below to make things clearer.

pkg/cache/external/scheduler_cache.go
craigcondit: If a pod fails to bind in the default scheduler for some reason, the only notice we may get (as there is not a callback in the scheduler framework for failed binds) is another call to PreFilter() / Filter() for this pod. So it is possible in that case for us to have an in-progress allocation. So, we remove it and reset the state back to pending to be safe.

craigcondit:
Changes in latest version:

- Rebased PR on master
- Fixed event recorder changes after YUNIKORN-937
- Implemented `--plugin` flag for `scripts/run-e2e-tests.sh`. E2E tests can be run using the plugin implementation with something like:

`scripts/run-e2e-tests.sh -a test -n "yk8s" -v "kindest/node:v1.21.2" --plugin`

craigcondit:
Rebased on master.

pkg/cache/external/scheduler_cache.go
craigcondit: Added additional documentation in scheduler_cache.go.

pkg/schedulerplugin/scheduler_plugin.go
craigcondit: Per offline discussion, removed this as the log message will never print it anyway.

craigcondit:
Added additional documentation to the scheduler_plugin.go and scheduler_cache.go files. Hopefully this is now clear.

craigcondit:
@pbacsko can you review again?

pkg/cache/external/scheduler_cache.go
pbacsko: Thanks, this definitely helps to understand the code.

pkg/schedulerplugin/scheduler_plugin.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return framework.NewStatus(framework.Unschedulable, "Pod is not ready for scheduling")
+}
+
+// PreFilterExtensions is unused
+func (sp *YuniKornSchedulerPlugin) PreFilterExtensions() framework.PreFilterExtensions {
+	return nil
+}
+
+// Filter is used to release specific pod/node combinations to scheduler
```
pbacsko, line 132: The thing that's probably worth mentioning here is what happens when `expectedNode != node`, pointing out the minimal performance impact and the immediate invocation of `Filter()` again. You can probably also add that currently this is the only way to force the default scheduler to match YK's decision.

pkg/schedulerplugin/scheduler_plugin.go
```diff
@@ -0,0 +1,209 @@
...
+// allocated the pod, but the default scheduler (via the plugin interface) has not yet been notified.
+//
+// Once PreFilter() has been called for a particular pod, that allocation is marked as "in progress" meaning it has been
+// communicated to the default scheduler, but has not yet been fulfilled.
+//
+// Finally, in PostBind(), the allocation is removed as we now know that the pod has been allocated successfully.
+// If a pending or in-progress allocation is detected for a pod in PreFilter(), we remove the allocation and force the
+// pod to be rescheduled, as this means the prior allocation could not be completed successfully by the default
+// scheduler for some reason.
+type YuniKornSchedulerPlugin struct {
```
pbacsko, line 70: Solid comment, it's great to have this here.

pbacsko (changes requested):
Pending LGTM +1 - I just asked to add one more minor thing.

pkg/schedulerplugin/scheduler_plugin.go
craigcondit: Added comments.

craigcondit:
@pbacsko can you re-review? I've addressed the comments you made on the last revision.

pbacsko (approved):
+1

PR #344, opened 2021-12-15
YUNIKORN-462: Streamline core to shim update on allocation change

manirajv06:
### What is this PR for?
Cache update and corresponding pod bind/delete happens together for new/release allocations respectively in a single callback method which gets triggered from core synchronously.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-462

### Questions:


wilfred-s:
Changed this to draft for now, lets finish up the SI and core first so we can update in one go.

manirajv06:
as discussed, cache would be cleared in all 4 remove allocations cases through ForgetPod()

wilfred-s:
@craigcondit I think we are OK with this one now. Please check with me so we have the correct behaviour.

craigcondit (approved):
Looks reasonable to me.

wilfred-s (approved):
LGTM

PR #345, opened 2021-12-16
[YUNIKORN-928] refactor admission controller code

chungen0126:
### What is this PR for?

This pr refactor the admission control code, and add a nil check for admission request.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [x] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-928
There is a existing unit test.
### Questions:

### DIFF
--- a/pkg/plugin/admissioncontrollers/webhook/admission_controller.go
+++ b/pkg/plugin/admissioncontrollers/webhook/admission_controller.go
@@ -294,16 +294,18 @@ func (c *admissionController) serve(w http.ResponseWriter, r *http.Request) {
 	req := ar.Request
 	urlPath := r.URL.Path
 	if urlPath == mutateURL || urlPath == validateConfURL {
-		if _, _, err := deserializer.Decode(body, nil, &ar); err != nil {
+		_, _, err := deserializer.Decode(body, nil, &ar)
+		switch {
+		case err != nil:
 			log.Logger().Error("Can't decode the body", zap.Error(err))
 			admissionResponse = admissionResponseBuilder("4qjvp2775w", false, err.Error(), nil)
-		} else if req != nil {
+		case req != nil:
 			if urlPath == mutateURL {
 				admissionResponse = c.mutate(req)
 			} else if urlPath == validateConfURL {
 				admissionResponse = c.validateConf(req)
 			}
-		} else {
+		default:
 			log.Logger().Warn("request is not exist", zap.String("urlPath", urlPath))
 		}
 	} else {
--- a/pkg/plugin/admissioncontrollers/webhook/admission_controller.go
+++ b/pkg/plugin/admissioncontrollers/webhook/admission_controller.go
@@ -216,7 +216,8 @@ func isConfigMapUpdateAllowed(userInfo string) bool {
 func (c *admissionController) validateConf(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionResponse {
 	uid := string(req.UID)
 	if !isConfigMapUpdateAllowed(req.UserInfo.Username) {
-		return admissionResponseBuilder(uid, false, fmt.Sprintf(configHotFreshResponse), nil)
+		errMessage := fmt.Sprintf(configHotFreshResponse)
+		return admissionResponseBuilder(uid, false, errMessage, nil)
 	}
 
 	var requestKind = req.Kind.Kind
--- a/pkg/plugin/admissioncontrollers/webhook/admission_controller.go
+++ b/pkg/plugin/admissioncontrollers/webhook/admission_controller.go
@@ -47,7 +47,7 @@ const (
 	yunikornPod                  = "yunikorn"
 	defaultQueue                 = "root.default"
 	configHotFreshResponse       = "ConfigHotRefresh is disabled. " +
-		"Please use the REST API to update the configuration, or enable configHotRefresh"
+		"Please use the REST API to update the configuration, or enable configHotRefresh. "
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -69,9 +70,23 @@ type ValidateConfResponse struct {
...
-func (c *admissionController) mutate(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
-	req := ar.Request
-	namespace := ar.Request.Namespace
+func admissionResponseBuilder(uid string, allowed bool, resultMessage string, patch []byte) *v1beta1.AdmissionResponse {
+	res := &v1beta1.AdmissionResponse{
+		Allowed: allowed,
+		UID:     types.UID(uid),
+		Result: &metav1.Status{
+			Message: resultMessage,
+		},
```
wilfred-s, line 79: If the result message is empty, `len(message) == 0` we should not add it to the response. We should keep the response as clean as possible.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -69,9 +70,23 @@ type ValidateConfResponse struct {
...
+		Patch: patch,
+		PatchType: func() *v1beta1.PatchType {
+			pt := v1beta1.PatchTypeJSONPatch
+			return &pt
+		}(),
```
wilfred-s, line 84: If the patch is empty, i.e. `len(patch) == 0` we should not add the patch and type to the response.
Using a function to generate the variable is a lot of overhead. It will not change using a var defined at the file level is easier.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -81,9 +96,7 @@ func (c *admissionController) mutate(ar *v1beta1.AdmissionReview) *v1beta1.Admis
 		log.Logger().Warn("request kind is not pod", zap.String("uid", uid),
 			zap.String("requestKind", requestKind))
 
-		return &v1beta1.AdmissionResponse{
-			Allowed: true,
-		}
+		return admissionResponseBuilder(uid, true, "", []byte(""))
```
wilfred-s, line 99: The `[]byte("")` should be replaced with a nil. That lines up with the comment earlier about not adding the patch details if not set.
This should be applied to all calls that pass in the empty array.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -213,51 +207,32 @@ func isConfigMapUpdateAllowed(userInfo string) bool {
...
 	if !isConfigMapUpdateAllowed(req.UserInfo.Username) {
-		return &v1beta1.AdmissionResponse{
-			Allowed: false,
-			Result: &metav1.Status{
-				Message: fmt.Sprintf("ConfigHotRefresh is disabled. " +
-					"Please use the REST API to update the configuration, or enable configHotRefresh"),
-			},
-		}
+		return admissionResponseBuilder(uid, false, fmt.Sprintf("ConfigHotRefresh is disabled. "+
+			"Please use the REST API to update the configuration, or enable configHotRefresh"), []byte(""))
```
wilfred-s, line 214: Not introduced here but can we replace this with a constant in the file? The text will never change and generating a new string is just overhead.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -311,20 +286,18 @@ func (c *admissionController) serve(w http.ResponseWriter, r *http.Request) {
...
-			Allowed: false,
-			Result: &metav1.Status{
-				Message: err.Error(),
-			},
+		admissionResponse = admissionResponseBuilder(string(req.UID), false, err.Error(), []byte(""))
+	} else if req != nil {
+		if urlPath == mutateURL {
+			admissionResponse = c.mutate(req)
+		} else if urlPath == validateConfURL {
+			admissionResponse = c.validateConf(req)
```
wilfred-s, line 299: if the deserialiser fails we must not assume that `ar` or `req` are valid and can be used. We should provide a fixed ID as part of the response.

wilfred-s (changes requested):
Bad behaviour that we must fix: any URL that is not one of the supported ones should return a 404 response. 
We should really fix that now in the `serve()` code. Check the path, if it is not `validateConfURL` or `mutateURL` do nothing not even deserialise. Return a fixed message and 404 and stop. We should also not return an admission review in that case.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -46,6 +46,8 @@ const (
 	enableConfigHotRefreshEnvVar = "ENABLE_CONFIG_HOT_REFRESH"
 	yunikornPod                  = "yunikorn"
 	defaultQueue                 = "root.default"
+	configHotFreshResponse       = "ConfigHotRefresh is disabled. " +
```
wilfred-s, line 49: nit: We do not enforce line lengths so this can be one line

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -69,9 +72,27 @@ type ValidateConfResponse struct {
...
+		res.Result = &metav1.Status{
+			Message: resultMessage,
+		}
+	}
+	if len(patch) != 0 {
+		res.Patch = patch
+		res.PatchType = func() *v1beta1.PatchType {
+			pt := v1beta1.PatchTypeJSONPatch
+			return &pt
+		}()
```
wilfred-s, line 90: simpler to just do this with the same result and less overhead:

	pt := v1beta1.PatchTypeJSONPatch
	res.PatchType = &pt

wilfred-s (changes requested):
For the `serve()` method please see the patch I put in slack with a proposal.

wilfred-s (approved):
LGTM
committing the change, thank you for your first contribution!

PR #349, opened 2022-01-10
[YUNIKORN-937] Update eventrecorder usage to new K8S API

anuraagnalluri:
### What is this PR for?
Upgrade usage of EventRecorder to report "k8s.io/api/events/v1beta1".Event events instead of "k8s.io/api/core/v1".Event events. More details on deprecation of core API can be found under Rule 1's Note [here](https://kubernetes.io/docs/reference/using-api/deprecation-policy/).

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-937

Existing unit + e2e tests should pass.

### DIFF
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -422,7 +422,7 @@ func (app *Application) scheduleTasks(taskScheduleCondition func(t *Task) bool) 
 					log.Logger().Warn("init task failed", zap.Error(err))
 				}
 			} else {
-				events.GetRecorder().Event(task.GetTaskPod(), v1.EventTypeWarning, "FailedScheduling", err.Error())
+				events.GetRecorder().Eventf(task.GetTaskPod(), nil, v1.EventTypeWarning, "FailedScheduling", "FailedScheduling", err.Error())
 				log.Logger().Debug("task is not ready for scheduling",
 					zap.String("appID", task.applicationID),
 					zap.String("taskID", task.taskID),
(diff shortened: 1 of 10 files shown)


pkg/cache/application.go
```diff
@@ -422,7 +422,7 @@ func (app *Application) scheduleTasks(taskScheduleCondition func(t *Task) bool)
 					log.Logger().Warn("init task failed", zap.Error(err))
 				}
 			} else {
-				events.GetRecorder().Event(task.GetTaskPod(), v1.EventTypeWarning, "FailedScheduling", err.Error())
+				events.GetRecorder().Eventf(task.GetTaskPod(), nil, v1.EventTypeWarning, "FailedScheduling", "FailedScheduling", err.Error())
```
anuraagnalluri, line 425: Minor difference in new API here: https://pkg.go.dev/k8s.io/client-go@v0.20.11/tools/events#EventRecorder

Using `nil` in place of `related` (2nd param) currently since we did not have concept of secondary object in previous Events API. By same logic, we're duplicating `reason` param to be same as `action` since the latter was also not present in previous API. 

Let me know if there's a better structure I can follow to populate these args correctly in `Eventf`.

craigcondit: This looks fine to me.

craigcondit (approved):
+1. LGTM. Thanks for the contribution @anuraagnalluri!

PR #352, opened 2022-01-12
[YUNIKORN-1014] Update e2e test matrix to include K8s 1.22.x, 1.23.x

craigcondit:
### What is this PR for?
Update the e2e test matrix to include Kubernetes 1.22 and 1.23.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [x] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1014

### Questions:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,7 +36,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        k8s: [v1.21.2, v1.20.7, v1.19.11]
+        k8s: [v1.23.1, v1.22.4, v1.21.2, v1.20.7, v1.19.11]
     steps:
       - uses: actions/setup-go@v2
         with:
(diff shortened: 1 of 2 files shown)


yangwwei:
Just had a discussion with @craigcondit , in the past, we are following the model of supporting 3 major K8s versions per release, but sometimes we are sort of left behind. It is important for us to support the latest k8s versions, we need to improve this. Actually, YK only depends on some pretty core APIs, usually, a release could support many K8s versions. We can revisit our support model in a follow-up community discussion. Approve this for now to have better coverage.

yangwwei:
One potential issue with this is the time needed to run all tests. Even github action parallelize these tasks, it sometimes may be queued and delayed due to resource shortage. The last run took ~10m, not bad. But if we need to keep an eye on the execution time, if it gets worse, we need to revisit this.

yangwwei (approved):

PR #356, opened 2022-01-18
[YUNIKORN-1027] Go: deprecation of io/ioutil

steinsgateted:
### What is this PR for?
Go's ioutil package to be deprecated in 1.16
change ioutil to io or os

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-1027

### Questions:

### DIFF
--- a/pkg/cache/application_graphviz_test.go
+++ b/pkg/cache/application_graphviz_test.go
@@ -1,4 +1,5 @@
 //go:build graphviz
+// +build graphviz
 
 /*
  Licensed to the Apache Software Foundation (ASF) under one
--- a/pkg/cache/application_graphviz_test.go
+++ b/pkg/cache/application_graphviz_test.go
@@ -1,4 +1,4 @@
-// +build graphviz
+//go:build graphviz
 
 /*
  Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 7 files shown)


yangwwei:
Change looks good, not sure I understand codecov's failure, where does that 50% diff come from?

pkg/cache/application_graphviz_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build graphviz
 // +build graphviz
```
pbacsko, line 2: Do we need the old build tag? It will be removed in 1.18 according to https://www.jetbrains.com/help/go/configuring-build-constraints-and-vendoring.html#build-constraints.

I think you can remove the old tag.

pkg/cache/task_graphviz_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build graphviz
 // +build graphviz
```
pbacsko, line 2: Remove the old tag.

pbacsko (changes requested):
I added some minor comments regarding build tags.

pbacsko (approved):
+1
merging this

PR #357, opened 2022-01-21
[YUNIKORN-1038] failure to ignore scheduler pod

wilfred-s:
### What is this PR for?
A missing return causes the admission controller to not ignore the
yunikorn pods. This could cause changes on the yunikorn pod which should
not be made.

### What type of PR is it?
* Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1038

Restart of the scheduler pod with the admission controller running.
Needs manual confirmation


yuchaoran2011 (approved):
I've manually tested this change. LGTM. Waiting for CI results

PR #358, opened 2022-01-21
[YUNIKORN-949] Location of the state dump file should be configurable

anuraagnalluri:
### What is this PR for?
Shim changes for enabling the state dump file should be configurable instead of hardcoded.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-949

* Manually using non-helm deployment.

### DIFF
--- a/deployments/scheduler/scheduler.yaml
+++ b/deployments/scheduler/scheduler.yaml
@@ -47,6 +47,11 @@ spec:
               value: myCluster
             - name: CLUSTER_VERSION
               value: latest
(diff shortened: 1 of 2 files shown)


anuraagnalluri:
No longer needed since we will make state dump output path a partition level config property instead of env var.

PR #359, opened 2022-01-24
[YUNIKORN-1041] Fix binding of dynamic volumes to pod

craigcondit:
### What is this PR for?
Fix binding of dynamic volumes to pods. This is broken after YUNIKORN-872. Also add additional error logging.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1041

Manual testing for now, e2e tests to be added to YUNIKORN-1039.

### Questions:

Error occurring under 0.12.1:

error-0-12-1

Error gone after fix:

no-error-fix

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/tools/cache"
 	podutil "k8s.io/kubernetes/pkg/api/v1/pod"
+	"k8s.io/kubernetes/pkg/controller/volume/scheduling"
 
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/appmgmt/interfaces"
 	schedulercache "github.com/apache/incubator-yunikorn-k8shim/pkg/cache/external"
@@ -365,19 +366,68 @@ func (ctx *Context) bindPodVolumes(pod *v1.Pod) error {
 				zap.String("podName", pod.Name))
 		} else {
 			log.Logger().Info("Binding Pod Volumes", zap.String("podName", pod.Name))
-			boundClaims, claimsToBind, _, err := ctx.apiProvider.GetAPIs().VolumeBinder.GetPodVolumes(assumedPod)
+			boundClaims, claimsToBind, unboundClaimsImmediate, err := ctx.apiProvider.GetAPIs().VolumeBinder.GetPodVolumes(assumedPod)
 			if err != nil {
+				log.Logger().Error("Failed to get pod volumes",
+					zap.String("podName", assumedPod.Name),
+					zap.Error(err))
+				return err
+			}
+			if len(unboundClaimsImmediate) > 0 {
+				err = fmt.Errorf("pod %s has unbound immediate claims", pod.Name)
+				log.Logger().Error("Pod has unbound immediate claims",
+					zap.String("podName", assumedPod.Name),
+					zap.Error(err))
 				return err
 			}
 			node, err := ctx.schedulerCache.GetNodeInfo(assumedPod.Spec.NodeName)
 			if err != nil {
+				log.Logger().Error("Failed to get node info",
+					zap.String("podName", assumedPod.Name),
+					zap.String("nodeName", assumedPod.Spec.NodeName),
+					zap.Error(err))
 				return err
 			}
-			volumes, _, err := ctx.apiProvider.GetAPIs().VolumeBinder.FindPodVolumes(assumedPod, boundClaims, claimsToBind, node)
+			volumes, reasons, err := ctx.apiProvider.GetAPIs().VolumeBinder.FindPodVolumes(assumedPod, boundClaims, claimsToBind, node)
 			if err != nil {
+				log.Logger().Error("Failed to find pod volumes",
+					zap.String("podName", assumedPod.Name),
+					zap.String("nodeName", assumedPod.Spec.NodeName),
+					zap.Int("claimsToBind", len(claimsToBind)),
+					zap.Error(err))
 				return err
 			}
-			return ctx.apiProvider.GetAPIs().VolumeBinder.BindPodVolumes(assumedPod, volumes)
+			if len(reasons) > 0 {
+				sReasons := make([]string, 0)
+				for _, reason := range reasons {
+					sReasons = append(sReasons, string(reason))
+				}
+				sReason := strings.Join(sReasons, ", ")
+				err = fmt.Errorf("pod %s has conflicting volume claims: %s", pod.Name, sReason)
+				log.Logger().Error("Pod has conflicting volume claims",
+					zap.String("podName", assumedPod.Name),
+					zap.String("nodeName", assumedPod.Spec.NodeName),
+					zap.Int("claimsToBind", len(claimsToBind)),
+					zap.Error(err))
+				return err
+			}
+			if volumes.StaticBindings == nil {
+				// convert nil to empty array
+				volumes.StaticBindings = make([]*scheduling.BindingInfo, 0)
+			}
+			if volumes.DynamicProvisions == nil {
+				// convert nil to empty array
+				volumes.DynamicProvisions = make([]*v1.PersistentVolumeClaim, 0)
+			}
+			err = ctx.apiProvider.GetAPIs().VolumeBinder.BindPodVolumes(assumedPod, volumes)
+			if err != nil {
+				log.Logger().Error("Failed to bind pod volumes",
+					zap.String("podName", assumedPod.Name),
+					zap.String("nodeName", assumedPod.Spec.NodeName),
+					zap.Int("dynamicProvisions", len(volumes.DynamicProvisions)),
+					zap.Int("staticBindings", len(volumes.StaticBindings)))
+				return err
+			}
 		}
 	}
 	return nil


pkg/cache/context.go
```diff
@@ -365,19 +366,68 @@ func (ctx *Context) bindPodVolumes(pod *v1.Pod) error {
...
+				sReason := strings.Join(sReasons, ", ")
+				err = fmt.Errorf("pod %s has conflicting volume claims: %s", pod.Name, sReason)
+				log.Logger().Error("Pod has conflicting volume claims",
+					zap.String("podName", assumedPod.Name),
+					zap.String("nodeName", assumedPod.Spec.NodeName),
+					zap.Int("claimsToBind", len(claimsToBind)),
+					zap.Error(err))
+				return err
+			}
+			if volumes.StaticBindings == nil {
```
craigcondit, line 414: This is the actual fix (ensure these slices are empty rather than nil).

pkg/cache/context.go
```diff
@@ -365,19 +366,68 @@ func (ctx *Context) bindPodVolumes(pod *v1.Pod) error {
...
+				// convert nil to empty array
+				volumes.StaticBindings = make([]*scheduling.BindingInfo, 0)
+			}
+			if volumes.DynamicProvisions == nil {
+				// convert nil to empty array
+				volumes.DynamicProvisions = make([]*v1.PersistentVolumeClaim, 0)
+			}
```
craigcondit, line 421: This is the actual fix -- don't allow the slices within volumes to be nil.

pkg/cache/context.go
```diff
@@ -365,19 +366,68 @@ func (ctx *Context) bindPodVolumes(pod *v1.Pod) error {
...
+					zap.Error(err))
+				return err
+			}
+			if len(unboundClaimsImmediate) > 0 {
+				err = fmt.Errorf("pod %s has unbound immediate claims", pod.Name)
+				log.Logger().Error("Pod has unbound immediate claims",
+					zap.String("podName", assumedPod.Name),
+					zap.Error(err))
 				return err
 			}
```
craigcondit, line 382: This is based on upstream K8s; if we have unbound immediate claims, we can't satisfy this.

pkg/cache/context.go
```diff
@@ -365,19 +366,68 @@ func (ctx *Context) bindPodVolumes(pod *v1.Pod) error {
...
+				}
+				sReason := strings.Join(sReasons, ", ")
+				err = fmt.Errorf("pod %s has conflicting volume claims: %s", pod.Name, sReason)
+				log.Logger().Error("Pod has conflicting volume claims",
+					zap.String("podName", assumedPod.Name),
+					zap.String("nodeName", assumedPod.Spec.NodeName),
+					zap.Int("claimsToBind", len(claimsToBind)),
+					zap.Error(err))
+				return err
+			}
```
craigcondit, line 413: This was added based on upstream K8s. If any reasons are returned from FindPodVolumes, there are conflicts and we must fail.

wilfred-s: looks good

yangwwei: Good catch.

```
volumes, reasons, err := ctx.apiProvider.GetAPIs().VolumeBinder.FindPodVolumes(assumedPod, boundClaims, claimsToBind, node)
```
Seems like there will be a case where `err==nil` but `len(reasons) > 0`

> // It returns an error when something went wrong or a list of reasons why the node is
	// (currently) not usable for the pod.

pkg/cache/context.go
yangwwei: why does this fix the problem? 
I did not see anywhere to complain NIL value

yangwwei (changes requested):

yangwwei:
Do you plan to add any UT or e2e test for this?

pkg/cache/context.go
wilfred-s: This prevents the early return in the K8s code. The K8s API code expects an empty array and never a nil for any success case. Even in the K8s test code a nil seems to be replaced with an empty array for success cases if I read the code correctly.

pkg/cache/context.go
```diff
@@ -365,19 +366,68 @@ func (ctx *Context) bindPodVolumes(pod *v1.Pod) error {
 				zap.String("podName", pod.Name))
 		} else {
 			log.Logger().Info("Binding Pod Volumes", zap.String("podName", pod.Name))
-			boundClaims, claimsToBind, _, err := ctx.apiProvider.GetAPIs().VolumeBinder.GetPodVolumes(assumedPod)
+			boundClaims, claimsToBind, unboundClaimsImmediate, err := ctx.apiProvider.GetAPIs().VolumeBinder.GetPodVolumes(assumedPod)
```
wilfred-s, line 369: This same call is made in `AssumePod` line 457 why do we not need the change there?

pkg/cache/context.go
wilfred-s: The nil checks for the `volumes.StaticBindings` and `volumes.DynamicProvisions` have been in the code since k8s 1.12. The real cause is a flow on effect of the change in the `FindPodVolumes` return value as part of a cleanup commit.  That was added in k8s 1.19 and later: 0 length slices are replaced with nil values:
https://gitlab.cncf.ci/kubernetes/kubernetes/commit/2cdc63aeaab603c61b48e95c044f1e4bb2f32548

We need tests for this code path.

wilfred-s (changes requested):
`bindPodVolumes` from the context is called from an event in the task. That task event should have dispatched a failure which I assumed should have failed the whole task which should have triggered a retry etc. 

If there was no failure there is an error handling issue. These changes do not change this at all. Does that mean that volume bind failures are not handled at all?
Having a warning event does not equate to a failure does it?

We should have a test for this

pkg/cache/context.go
craigcondit: > We need tests for this code path.

I don't disagree. However, currently, the entirety of volume binding is disabled in test mode. Mocking out the proper interactions is going to be a pretty major undertaking.

pkg/cache/context.go
craigcondit: Exactly.

pkg/cache/context.go
```diff
@@ -365,19 +366,68 @@ func (ctx *Context) bindPodVolumes(pod *v1.Pod) error {
...
+				log.Logger().Error("Failed to get pod volumes",
+					zap.String("podName", assumedPod.Name),
+					zap.Error(err))
+				return err
+			}
+			if len(unboundClaimsImmediate) > 0 {
+				err = fmt.Errorf("pod %s has unbound immediate claims", pod.Name)
+				log.Logger().Error("Pod has unbound immediate claims",
+					zap.String("podName", assumedPod.Name),
+					zap.Error(err))
```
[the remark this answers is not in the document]
craigcondit: I agree, we probably need to test for this earlier. I'll work on trying to get this working earlier as well. Your suggestion as to where makes sense -- I will investigate that.

craigcondit:
> Do you plan to add any UT or e2e test for this?

Yes, e2e tests will be added in YUNIKORN-1039. Those will take time to develop and this issue is a blocker for 0.12.2 so I don't think we should hold up the fix for the e2e tests to be developed.

pkg/cache/context.go
craigcondit: We probably do. I missed that one.

pkg/cache/context.go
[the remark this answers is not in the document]
craigcondit: Looking at this a bit more, I don't believe we need to worry about this. The existing VolumeBinder plugin gets called as part of pod / node evaluation, so this logic (which comes from PreFilter) should already fire prior to us attempting to bind the node.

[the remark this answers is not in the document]
craigcondit: We should still check it during binding to ensure that the status has not changed though.

pkg/cache/context.go
craigcondit: actually, i don't think we do. This logic will have already run via the PreFilter() / Filter() calls on the upstream VolumeBinder plugin (we call these as part of node fitness checks).

pkg/cache/context.go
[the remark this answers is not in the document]
yangwwei: I see, thanks. Looks like this is already covered in [predict manager](https://github.com/apache/incubator-yunikorn-k8shim/blob/799a42ed9b247815fd344f70e0df27d6b83f90d2/pkg/plugin/predicates/predicate_manager.go#L86), that's good.

yangwwei (approved):
Looks good from my side. Thanks for fixing this!

craigcondit:
> `bindPodVolumes` from the context is called from an event in the task. That task event should have dispatched a failure which I assumed should have failed the whole task which should have triggered a retry etc.
> 
> If there was no failure there is an error handling issue. These changes do not change this at all. Does that mean that volume bind failures are not handled at all? Having a warning event does not equate to a failure does it?

`task.postTaskAllocated()` dispatches a `FailTaskEvent` if `err != nil` is returned from `context.bindPodVolumes()`, so I think we're good there.

pkg/cache/context.go
craigcondit: Given the major surgery required to get more of context.go tested, I'd like to move forward with this PR as-is to unblock 0.12.2 and add e2e tests in YUNIKORN-1039 targeting 1.0.0. @wilfred-s does that sound ok with you?

craigcondit:
Moving forward with this to unblock 0.12.2 release. @wilfred-s, if you have additional concerns, let's address in a follow-up patch for 1.0.0.

pkg/cache/context.go
wilfred-s: yes that is good

PR #362, opened 2022-02-04
[YUNIKORN-1052] Add more cluster-wide metrics to Yunikorn

pbacsko:
### What is this PR for?
Add more cluster-wide metrics to Yunikorn.
We're exposing the "Ready" condition from the shim side on node addition/update. 
https://kubernetes.io/docs/concepts/architecture/nodes/#condition

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1052

### Questions:

### DIFF
--- a/pkg/cache/nodes.go
+++ b/pkg/cache/nodes.go
@@ -114,8 +114,9 @@ func (nc *schedulerNodes) addAndReportNode(node *v1.Node, reportNode bool) {
 			zap.String("nodeLabels", string(nodeLabels)),
 			zap.Bool("schedulable", !node.Spec.Unschedulable))
 
+		ready := common.HasReadyCondition(node)
 		newNode := newSchedulerNode(node.Name, string(node.UID), string(nodeLabels),
-			common.GetNodeResource(&node.Status), nc.proxy, !node.Spec.Unschedulable)
+			common.GetNodeResource(&node.Status), nc.proxy, !node.Spec.Unschedulable, ready)
 		nc.nodesMap[node.Name] = newNode
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/common/si_helper.go
```diff
@@ -159,7 +160,9 @@ func CreateUpdateRequestForUpdatedNode(node Node) si.NodeRequest {
 	// Currently only includes resource in the update request
 	nodeInfo := &si.NodeInfo{
 		NodeID:              node.name,
-		Attributes:          make(map[string]string),
+		Attributes:          map[string]string {
+			"ready":		 strconv.FormatBool(node.ready),
```
pbacsko, line 164: Temporary solution. Either should be in `NodeInfo` or define a constant for it.

pkg/cache/nodes.go
```diff
@@ -212,6 +213,15 @@ func (nc *schedulerNodes) updateNode(oldNode, newNode *v1.Node) {
 		return
 	}
 
+	if schedulerNode, ok := nc.nodesMap[newNode.Name]; ok {
+		schedulerNode.ready = common.HasReadyCondition(newNode)
```
pbacsko, line 217: Updating this field in `SchedulerNode` is probably not necessary, but for the sake of consistency, I decided to do it anyway.

wilfred-s: The `nodesMap` looks broken. I think we have a bigger issue in this code.

wilfred-s: I filed YUNIKORN-1091 for the broken nodesMap update.
We do need this update in the node. This change will clash with that fix.
I also logged YUNIKORN-1090 to remove the common.Node object. That change is impacted by what we have here too.
Moving forward and leaving the fix up for the two new jiras.

wilfred-s: I filed YUNIKORN-1091 for the broken nodesMap update.
We do need this update in the node. This change will clash with that fix.
I also logged YUNIKORN-1090 to remove the common.Node object. That change is impacted by what we have here too.
Moving forward and leaving the fix up for the two new jiras.

wilfred-s (approved):
LGTM

PR #365, opened 2022-02-08
[YUNIKORN-876] Fix pod memory leaks in scheduler cache

craigcondit:
### What is this PR for?
Fix a major memory leak in scheduler cache that causes pods to be stored forever.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-876

Local unit tests pass ; also verified via debugger that removed pods are properly removed from cache.

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -215,6 +215,14 @@ func (ctx *Context) addPodToCache(obj interface{}) {
 		return
 	}
 
+	// if a terminated pod is added to cache, it will
+	// add requested resource to the cached node, causing
+	// the node uses more resources that it actually is,
+	// this can only be fixed after the pod is removed.
+	if utils.IsPodTerminated(pod) {
+		return
+	}
+
 	log.Logger().Debug("adding pod to cache", zap.String("podName", pod.Name))
 	if err := ctx.schedulerCache.AddPod(pod); err != nil {
 		log.Logger().Error("add pod to scheduler cache failed",
@@ -241,11 +249,7 @@ func (ctx *Context) removePodFromCache(obj interface{}) {
 	}
 
 	log.Logger().Debug("removing pod from cache", zap.String("podName", pod.Name))
-	if err := ctx.schedulerCache.RemovePod(pod); err != nil {
-		log.Logger().Debug("failed to remove pod from scheduler cache",
-			zap.String("podName", pod.Name),
-			zap.Error(err))
-	}
+	ctx.schedulerCache.RemovePod(pod)
 }
 
 func (ctx *Context) updatePodInCache(oldObj, newObj interface{}) {
@@ -260,6 +264,11 @@ func (ctx *Context) updatePodInCache(oldObj, newObj interface{}) {
 		return
 	}
 
+	// ignore terminated pods
+	if utils.IsPodTerminated(newPod) {
+		return
+	}
+
 	if err := ctx.schedulerCache.UpdatePod(oldPod, newPod); err != nil {
 		log.Logger().Debug("failed to update pod in cache",
 			zap.String("podName", oldPod.Name),
@@ -271,13 +280,7 @@ func (ctx *Context) updatePodInCache(oldObj, newObj interface{}) {
 func (ctx *Context) filterPods(obj interface{}) bool {
 	switch obj := obj.(type) {
 	case *v1.Pod:
-		// if a terminated pod is added to cache, it will
-		// add requested resource to the cached node, causing
-		// the node uses more resources that it actually is,
-		// this can only be fixed after the pod is removed.
-		// (trigger the delete pod)
-		return utils.GeneralPodFilter(obj) &&
-			!utils.IsPodTerminated(obj)
+		return utils.GeneralPodFilter(obj)
 	default:
 		return false
 	}
--- a/pkg/cache/external/scheduler_cache.go
+++ b/pkg/cache/external/scheduler_cache.go
@@ -217,30 +217,26 @@ func (cache *SchedulerCache) AddPod(pod *v1.Pod) error {
 	defer cache.lock.Unlock()
 
 	currState, ok := cache.podsMap[key]
-	switch {
-	case ok && cache.isAssumedPod(key):
-		if currState.Spec.NodeName != pod.Spec.NodeName {
-			// The pod was added to a different node than it was assumed to.
-			log.Logger().Warn("inconsistent pod location",
-				zap.String("assumedLocation", pod.Spec.NodeName),
-				zap.String("actualLocation", currState.Spec.NodeName))
-
-			// Clean this up.
-			err = cache.removePod(currState)
-			if err != nil {
-				log.Logger().Debug("node not in cache",
-					zap.Error(err))
+	if ok {
+		// pod exists
+		if cache.isAssumedPod(key) {
+			if currState.Spec.NodeName != pod.Spec.NodeName {
+				// pod was added to a different node than it was assumed to
+				log.Logger().Warn("inconsistent pod location",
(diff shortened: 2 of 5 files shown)


pkg/cache/context.go
```diff
@@ -271,13 +284,7 @@ func (ctx *Context) updatePodInCache(oldObj, newObj interface{}) {
...
 	switch obj := obj.(type) {
 	case *v1.Pod:
-		// if a terminated pod is added to cache, it will
-		// add requested resource to the cached node, causing
-		// the node uses more resources that it actually is,
-		// this can only be fixed after the pod is removed.
-		// (trigger the delete pod)
-		return utils.GeneralPodFilter(obj) &&
-			!utils.IsPodTerminated(obj)
+		return utils.GeneralPodFilter(obj)
```
craigcondit, line 287: This logic gets moved to `addPodToCache()` / `updatePodInCache()` as we DO want to allow terminated pods to flow through to `removePodFromCache()`. The old logic filtered these out from all methods, resulting in leaks of terminated pods.

pkg/cache/external/scheduler_cache.go
```diff
@@ -283,34 +281,46 @@ func (cache *SchedulerCache) addPod(pod *v1.Pod) {
...
+	if removeAlloc {
+		// remove pod from any pending or in-progress allocations
+		delete(cache.pendingAllocations, string(pod.UID))
+		delete(cache.inProgressAllocations, string(pod.UID))
+	}
+
+	// remove pod from cache
+	delete(cache.podsMap, string(pod.UID))
+
+	// remove pod from node
```
craigcondit, line 314: Here we make sure to remove from cache regardless so that we don't leak resources.

pkg/cache/external/scheduler_cache.go
```diff
@@ -353,22 +359,15 @@ func (cache *SchedulerCache) ForgetPod(pod *v1.Pod) error {
...
-			return err
-		}
-		delete(cache.assumedPods, key)
-		delete(cache.podsMap, key)
-	default:
-		return fmt.Errorf("pod %v wasn't assumed so cannot be forgotten", key)
-	}
+	delete(cache.assumedPods, key)
+	delete(cache.pendingAllocations, key)
+	delete(cache.inProgressAllocations, key)
```
craigcondit, line 370: Removed most of this logic as it's just horribly broken. We now ensure that we cleanup the cache regardless of its current state, and we log a warning if the pod was assigned to an unexpected node instead of throwing an error.

wilfred-s (changes requested):
The code change itself looks good.
We need some tests for this area. 

The tests in the `scheduler_cache_test.go` does not contain real pod tests. All tests are focussed on the node and pod combination.  It does not cover the pod cache at all.
The `context_test.go` tests do not cover any of the base pod related calls in the context.

Overall that means we have zero real test coverage over the cache pod behaviour.
Running the tests shows less than 40% of the lines covered in the files that handle this cache. That has to increase to make sure we do not regress in this area.

craigcondit:
Added unit tests for both `context` and `scheduler_cache`. Also further simplified some of the error handling in `scheduler_cache`.

craigcondit:
I get a clean e2e test run locally, and build/unit tests pass locally and here. We seem to have some reliability problems with the e2e test runs. I've observed this at times locally, and it seems to be that the kind cluster doesn't always start up cleanly or in a timely manner, and so the initial development namespace fails to create. The failures are not YK-related.

craigcondit:
Rebased on master after YUNIKORN-1074, and fixed breakage affecting plugin mode.

yangwwei (commented):
Other seems good except the place I have commented. Basically, the `scheduler_cache.go` is a sibling of the default scheduler's `cache.go`. This was how we make YK fully compatible with the default scheduler's predicates functions. So I would suggest making them as close as possible. I don't think these places are covered in unit tests, or end-to-end tests, the places I pointed out might be handling some corner cases. It is risky if we introduce some behavior changes because YK is being pretty stable so far.

pkg/cache/external/scheduler_cache.go
```diff
@@ -217,30 +217,21 @@ func (cache *SchedulerCache) AddPod(pod *v1.Pod) error {
...
-				zap.String("actualLocation", currState.Spec.NodeName))
-
-			// Clean this up.
-			err = cache.removePod(currState)
-			if err != nil {
-				log.Logger().Debug("node not in cache",
-					zap.Error(err))
+	if ok {
+		// pod exists
+		if cache.isAssumedPod(key) {
```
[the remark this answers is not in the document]
craigcondit: > in the default scheduler side, it only updates the pod when the pod.sepc.NodeName changes and the pod is assumed. In our code, we do the update no matter if the pod is assumed or not.

Original code then gets out of sync whenever a pod is assumed. There is no reason to do this, and in the case where the pod gets I assumed we are now stale. We should keep the new code. 

> Another thing, when the pod node name is unchanged, and the pod is assumed, it deletes the pod from the cache. [cache.go#489](https://github.com/kubernetes/kubernetes/blob/3866cb91f22da6eb49dab10dd5c33385690b57b4/pkg/scheduler/internal/cache/cache.go#L489). We no longer have this logic anywhere.

This is even worse. Now we lose a pod entirely while we have an active allocation. This also has run-on effects in plug-in mode because we don't explicitly control the bind phase and are reliant on informer updates to keep us synchronized. Again, new code is necessary. 

We are not a carbon-copy of upstream K8s and I'm not willing to keep obviously broken code simply because that's how upstream does it. Caches are meant to reflect the things they cache. Any conditional updates de-synchronize it and make it useless. The original code tries to be entirely too clever by half and resulted in a massive memory leak due to that complexity.

pkg/cache/external/scheduler_cache.go
```diff
@@ -255,26 +246,25 @@ func (cache *SchedulerCache) UpdatePod(oldPod, newPod *v1.Pod) error {
...
-	case ok && !cache.isAssumedPod(key):
-		if currState.Spec.NodeName != newPod.Spec.NodeName {
-			log.Logger().Error("pod updated on a different node than previously added to", zap.String("pod", key))
-			log.Logger().Error("scheduler cache is corrupted and can badly affect scheduling decisions")
-		}
-		if err = cache.updatePod(oldPod, newPod); err != nil {
-			return err
+	if ok {
+		// pod exists and is assumed
+		if !cache.isAssumedPod(key) && currState.Spec.NodeName != newPod.Spec.NodeName {
```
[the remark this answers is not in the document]
craigcondit: See previous reply. Caches need to stay in sync with their original data.

craigcondit:
Updated PR with fix for crash discovered by @brickyard, as well as additional unit tests to bring add/update/remove node methods in scheduler_cache to 100% branch coverage, verifying that patch fixes crash.

At this point we have far better tested code for this than upstream. @yangwwei, can you sign off on this? I know you have some concerns about the differences from upstream, but at this point I think they are theoretical. Let's address in a follow-up JIRA if we discover a regression.

yangwwei (changes requested):
@craigcondit  can you please check if we need to do a set node here as well to avoid NPE? It is similar to what you have fixed in the 3rd commit.

_YUNIKORN-876__Fix_pod_memory_leaks_in_scheduler_cache_by_craigcondit_·_Pull_Request__365_·_apache_incubator-yunikorn-k8shim

yangwwei (approved):
Thanks for the confirmation. +1

PR #368, opened 2022-02-13
[YUNIKORN-979] Add e2e test coverage for the admission controller

lowc1012:
### What is this PR for?
Add the e2e test for the admission controller to increase test coverage.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
[YUNIKORN-979](https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-979)

### Questions:

### DIFF
--- a/test/e2e/admission_controller/admission_controller_test.go
+++ b/test/e2e/admission_controller/admission_controller_test.go
@@ -36,18 +36,19 @@ import (
 var _ = ginkgo.Describe("AdmissionController", func() {
 
 	ginkgo.BeforeEach(func() {
-		kClient = k8s.KubeCtl{}
-		gomega.Expect(kClient.SetClient()).To(gomega.BeNil())
+		kubeClient = k8s.KubeCtl{}
+		gomega.Expect(kubeClient.SetClient()).To(gomega.BeNil())
 	})
 
 	ginkgo.It("Verifying a pod is created in the test namespace", func() {
 
 		ginkgo.By("has 1 running pod whose SchedulerName is yunikorn")
 		sleepPodConfigs := common.SleepPodConfig{Name: sleepPodName, NS: ns}
-		sleepRespPod, err := kClient.CreatePod(common.InitSleepPod(sleepPodConfigs), ns)
+		sleepRespPod, err := kubeClient.CreatePod(common.InitSleepPod(sleepPodConfigs), ns)
+		gomega.Ω(err).ShouldNot(gomega.HaveOccurred())
 
 		// Wait for pod to move into running state
-		err = kClient.WaitForPodBySelectorRunning(ns,
+		err = kubeClient.WaitForPodBySelectorRunning(ns,
 			fmt.Sprintf("app=%s", sleepRespPod.ObjectMeta.Labels["app"]), 10)
 		gomega.Ω(err).ShouldNot(gomega.HaveOccurred())
 		gomega.Ω(sleepRespPod.Spec.SchedulerName).Should(gomega.BeEquivalentTo(constants.SchedulerName))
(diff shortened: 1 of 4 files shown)


wilfred-s:
@lowc1012 can you have a look at the two lint issues?
The kClient is a strange one to get. Not something you would expect. It is linked to not using the Hungarian notation. The go lint code specifically picks up on the `k` which stands for kind in that case. Rename the variable to kubeClient or just to client can be used.

wilfred-s:
One other thing: should we check that the admission controller is up and running by calling the health check before we run these tests or is that called earlier in the e2e tests?

lowc1012:
@wilfred-s Thanks for your explanation.
About the second comment, Do we have the health check API endpoint for admission controller? If not, should I implement it in the PR or create another JIRA for that? Thank you!

lowc1012:
I have no idea why the test for namespaceBlacklist failed.
In the test, I created a pod in kube-system namespace, it should not be mutated by admission controller. 
[YUNIKORN-1084](https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-1084) may address this problem.

wilfred-s:
@craigcondit can you please check if the e2e test failed as expected and that this will work in combination with your changes. If we need changes for YUNIKORN-1084 I think we need to commit this first and add changes to that PR.

craigcondit:
> @craigcondit can you please check if the e2e test failed as expected and that this will work in combination with your changes. If we need changes for YUNIKORN-1084 I think we need to commit this first and add changes to that PR.

Yes, this test will fail currently because the changes from YUNIKORN-1084 have not yet been merged (the existing behavior is not correct, but this test assumes it is). I don't think we need changes to this, but YUNIKORN-1084 needs to go in first to get the admission controller to exhibit the correct behavior.

craigcondit:
@lowc1012 can you re-base this on current master? I think all the admission controller fixes are in now, so hopefully the tests pass.

test/e2e/admission_controller/admission_controller_test.go
```diff
@@ -0,0 +1,86 @@
...
+		err = kubeClient.WaitForPodBySelectorRunning(ns,
+			fmt.Sprintf("app=%s", sleepRespPod.ObjectMeta.Labels["app"]), 10)
+		gomega.Ω(err).ShouldNot(gomega.HaveOccurred())
+		gomega.Ω(sleepRespPod.Spec.SchedulerName).Should(gomega.BeEquivalentTo(constants.SchedulerName))
+	})
+
+	ginkgo.It("Verifying a pod is created on namespace blacklist", func() {
+		ginkgo.By("Create a pod in namespace blacklist")
+		podConfigs := common.SleepPodConfig{Name: sleepPodName, NS: blackNs}
+		pod, err := kubeClient.CreatePod(common.InitSleepPod(podConfigs), blackNs)
```
craigcondit, line 60: Using `InitSleepPod()` here will explicitly set `schedulerName`: `yunikorn`. This prevents the test from passing.

craigcondit (changes requested):

test/e2e/admission_controller/admission_controller_test.go
lowc1012: I see and am sorry for my mistake.
Thank you.

craigcondit (approved):
+1 LGTM.

test/e2e/admission_controller/admission_controller_test.go
craigcondit: No worries! Thanks for the contribution.

PR #369, opened 2022-02-14
[YUNIKORN-1040] add e2e test that re-starts the scheduler pod

anuraagnalluri:
### What is this PR for?
Add an e2e verification for restarting the scheduler pod as discussed in [YUNIKORN-1040](https://issues.apache.org/jira/browse/YUNIKORN-1040). Added as a flow in ginkgo `beforeSuite` in `basicscheduling_test.go`. 

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-1040

Ran e2e test


test/e2e/basic_scheduling/basic_scheduling_test.go
```diff
@@ -70,6 +70,17 @@ var _ = ginkgo.Describe("", func() {
...
+		var podList, err4 = kClient.GetPods(configmanager.YuniKornTestConfig.YkNamespace)
+		Ω(err4).NotTo(HaveOccurred())
+		Ω(podList).NotTo(BeNil())
+		_, err5 := kClient.ScaleDeployment(configmanager.YKScheduler, 0, configmanager.YuniKornTestConfig.YkNamespace)
+		gomega.Ω(err5).NotTo(gomega.HaveOccurred())
+		err6 := kClient.WaitForPodBySelectorTerminated(configmanager.YuniKornTestConfig.YkNamespace, fmt.Sprintf("component=%s", configmanager.SchedulerName), 10)
+		kClient.ScaleDeployment(configmanager.YKScheduler, 1, configmanager.YuniKornTestConfig.YkNamespace)
+		gomega.Ω(err6).NotTo(gomega.HaveOccurred())
+		kClient.WaitForPodBySelectorRunning(configmanager.YuniKornTestConfig.YkNamespace, fmt.Sprintf("component=%s", configmanager.SchedulerName), 10)
+
```
anuraagnalluri, line 83: @wilfred-s Apologies if I missed the mark here, but I couldn't find a better place to include scheduler pod restart than the `BeforeSuite` of `basic_scheduling_test`. Please do advise if you have a better place in mind.

test/e2e/basic_scheduling/basic_scheduling_test.go
```diff
@@ -70,6 +70,16 @@ var _ = ginkgo.Describe("", func() {
 		Ω(err3).NotTo(HaveOccurred())
 		Ω(d).NotTo(BeNil())
 
+		ginkgo.By("Restart scheduler pod")
+		_, err4 := kClient.ScaleDeployment(configmanager.YKScheduler, 0, configmanager.YuniKornTestConfig.YkNamespace)
+		gomega.Ω(err4).NotTo(gomega.HaveOccurred())
+		err5 := kClient.WaitForPodBySelectorTerminated(configmanager.YuniKornTestConfig.YkNamespace, fmt.Sprintf("component=%s", configmanager.YKScheduler), 60)
```
anuraagnalluri, line 76: I'm really not sure why this timeout had to be 60. Waiting for pods to be fully terminated (this does not correspond to "terminating" in k8s state diagram, but rather failing an existence check for any pods with the `selector` labels in the given `namespace`) takes around 10-12 seconds. But even setting 30 here leads to a `timeout exceeded` error. Wasn't sure how to debug this or estimate the value more properly.

test/e2e/basic_scheduling/basic_scheduling_test.go
```diff
@@ -70,6 +70,16 @@ var _ = ginkgo.Describe("", func() {
 		Ω(err3).NotTo(HaveOccurred())
 		Ω(d).NotTo(BeNil())
 
+		ginkgo.By("Restart scheduler pod")
+		_, err4 := kClient.ScaleDeployment(configmanager.YKScheduler, 0, configmanager.YuniKornTestConfig.YkNamespace)
```
anuraagnalluri, line 74: Note k8s has no API call to bounce a pod. I'm simply scaling the replicas in the scheduler deployment to 0 and then back to 1.

test/e2e/framework/helpers/k8s/k8s_utils.go
```diff
@@ -295,14 +308,19 @@ func (k *KubeCtl) ListPods(namespace string, selector string) (*v1.PodList, erro
 }
 
 // Wait up to timeout seconds for all pods in 'namespace' with given 'selector' to enter running state.
-// Returns an error if no pods are found or not all discovered pods enter running state.
-func (k *KubeCtl) WaitForPodBySelectorRunning(namespace string, selector string, timeout int) error {
+// Returns an error if no pods are found when 'wait' is false or not all discovered pods enter running state within the 'timeout' duration.
+// If 'wait' is true, error will not be returned if no pods are found. Pods will be continually listed until there is a non-empty list
+// to iterate over.
+func (k *KubeCtl) WaitForPodBySelectorRunning(namespace string, selector string, timeout int, wait bool) error {
```
anuraagnalluri, line 314: We add this `wait` parameter because other invocations of `WaitForPodBySelectorRunning` are directly after calls to `CreatePod`. This allows for the pod object to be returned by the API server at time of execution. In the newly added code above to restart the scheduler pod, there is a latency between scaling the deployment back to 1 and the ensuing call to create a pod, meaning the object is not readily available to be returned by the API server. In this case, `WaitForPodBySelectorRunning` will just error immediately since there are no pods with the given `selector`, requiring us to change the behavior with a flag.

anuraagnalluri:
Hi @yangwwei , do you have general pointers on how I can debug the failing CI checks? Running the e2e tests locally on this branch passes, but getting the applications in pre-commit checks seem to fail.

wilfred-s (changes requested):
The scale up and scale down are way to nice to the scheduler. We want to kill the pod that runs the scheduler and then see if the pod gets rescheduled again. The admission controller should not be killed.
The fact that the scheduler runs as a deployment should cause the scheduler pod to be recreated and scheduled again.
We can follow these high level steps:
```
kClient.GetPodNamesFromNS("YK Namespace")
get the pod named "yunikorn-scheduler-*"
kClient.DeletePod("YK scheduler", "YK Namespace")
```
Even after that is done and the delete pod has returned without an error we should see a new scheduler pod in ready state within a short amount of time.

anuraagnalluri:
Thanks @wilfred-s, incorporated your suggestions and ran e2e tests locally with success. Not sure why the pre-commit checks seem to fail on getting the applications in CI. Any way I can get access to the cluster these run on and debug?

yangwwei:
Sorry for the late response, I was on vacation last week so emails/notifications got backlogged.  For this issue, @ronazhan can you share your thoughts?

I am not entirely sure what is the best way to debug this. The error shown up there:

```
Unexpected error:
      <*url.Error | 0xc0004c0210>: {
          Op: "Get",
          URL: "http://localhost:9080/ws/v1/apps",
          Err: {s: "EOF"},
      }
      Get "http://localhost:9080/ws/v1/apps": EOF
  occurred
```

that seems like the REST endpoint could not be accessed. Maybe this is because the scheduler pod gets restarted and we need to redo the port-forwarding? Here is how it was done during the initial setup: https://github.com/apache/incubator-yunikorn-k8shim/blob/a61fc0052c07e510503853db74030290bbda562b/scripts/run-e2e-tests.sh#L193-L199.

wilfred-s:
I don't know what happened with my last update. I must have closed a browser window without saving.
I think that @yangwwei is right the forward of the port has not been recreated after the restart and that will cause the issue for the rest of the tests.

I would also suggest that we move this restart test to its own test suite. Instead of making this part of basic scheduling and causing those tests to fail we should create a "recovery & restart" suite. Move this restart test in there and we can then extend it with e2e recover tests.

Using some more advanced ginkgo tricks we can then make sure these restart or recovery tests are run separate from all the others. Adding labels and splitting the ginkgo run into two runs would allow us to be more destructive in tests. I do think that it requires gingkgo v2 with these test labels. @ronazhan please some input on ginkgo v2 change also

ronazhan:
Yes, @wilfred-s and @yangwwei are right that the port-forward connection needs to be restarted as well after the new scheduler pod is brought up. I don't believe that port-forward is available using the k8s go-client library, so this might have to be done as a shell command

I haven't looked much into the ginkgo v2 changes, but the recovery suite can be separated by using `--skip-package`  flag. Then, it can be run separately than the other test packages

anuraagnalluri:
@yangwwei @wilfred-s @ronazhan Thank you for your inputs. I’ve made most of your suggestions and the e2e tests are passing locally again.

However, the CI checks seem to be hanging now. Perhaps something is off with the port-forward I added in k8s_utils — I’ll debug tomorrow when I can view the log output on the checks.

yangwwei:
Looks like the CI is failing after the recovery test suite, could it be possible there is still something wrong with the port-forwarding logic? Can we collect both stdout and stderr in `PortForwardService` while executing `exec.Command`: , e.g using https://pkg.go.dev/os/exec#Cmd.CombinedOutput to get more info?

anuraagnalluri:
@yangwwei Right, something is definitely off about the port-forwarding. If I use `exec.Output` or `exec.CombinedOutput`, I get a failure likely due to it being a blocking operation (which is why I went the route of using `exec.Start`). But this is swallowing the following error:
```
 [91mUnexpected error:
      <*exec.ExitError | 0x1400054e000>: {
          ProcessState: {
              pid: 69190,
              status: 256,
              rusage: {
                  Utime: {
                      Sec: 0,
                      Usec: 62424,
                      Pad_cgo_0: [0, 0, 0, 0],
                  },
                  Stime: {
                      Sec: 0,
                      Usec: 20745,
                      Pad_cgo_0: [0, 0, 0, 0],
                  },
                  Maxrss: 45203456,
                  Ixrss: 0,
                  Idrss: 0,
                  Isrss: 0,
                  Minflt: 3441,
                  Majflt: 34,
                  Nswap: 0,
                  Inblock: 0,
                  Oublock: 0,
                  Msgsnd: 17,
                  Msgrcv: 10,
                  Nsignals: 76,
                  Nvcsw: 67,
                  Nivcsw: 1116,
              },
          },
          Stderr: nil,
      }
      exit status 1
```

I thought about using `client-go/tools` for a go-native implementation as explained here: https://gianarb.it/blog/programmatically-kube-port-forward-in-go . But if we put the termination and port-forwarding logic in goroutines and use `sync/WaitGroup`, then this would still be a blocking operation, right?

Happy to look in to any other suggestions.

yangwwei:
hi @anuraagnalluri  thanks for the updates.

The error message is still not that informative, I am still unclear why this port-forward command would fail. Any way to get more meaningful error messages? Have you tried to run the same port forward code locally? That works perfectly fine?

The code in https://gianarb.it/blog/programmatically-kube-port-forward-in-go could be another approach, I think it won't block the main process

```
// I just wait until the readyCh tells me that the connection is up and running
select {
case <-readyCh:
    break
}
println("Port forwarding is ready to get traffic. have fun!")
```

actually it would be good to give this approach a try, because IIUC this way we can get rid of changes in the github action:

```
gosec:
    # To specify a set of rules to explicitly exclude.
    # Available rules: https://github.com/securego/gosec#available-rules
    excludes:
      - G204
```

ronazhan:
@anuraagnalluri Could this be because the previous port forward command was not killed?

anuraagnalluri:
@ronazhan That was one suspicion I had which is why I port-forwarded the service instead of the pod. The pod has a unique UUID post restart, but the service remains static. Since the command would effectively be for the same service, would we have to kill the old process and start a new one? It's difficult to verify what may be causing `fallback_test` to fail locally, because even when rebuilding the images on `master` and running the test, it reliably fails/times out [here](https://github.com/apache/incubator-yunikorn-k8shim/blob/master/test/e2e/state_aware_app_scheduling/fallback_test.go#L80). This happens regardless of whether `port-forward` proc is running or not on `master` and my task branch.

@yangwwei Makes sense, I think it would be good for removing gosec exclusion. I think the actual code the blog refers to [does block the main process](https://github.com/gianarb/kube-port-forward/blob/master/main.go#L111) until [`stopCh` is closed ](https://github.com/gianarb/kube-port-forward/blob/master/main.go#L79-L80). But we can likely remove these from our implementation to make it non-blocking, right?

yangwwei:
> But we can likely remove the syncs from our implementation to make it non-blocking, right?

Yeah, seems like that can be removed, please give a try! thanks!

anuraagnalluri:
> > But we can likely remove the syncs from our implementation to make it non-blocking, right?
> 
> Yeah, seems like that can be removed, please give a try! thanks!

@yangwwei Used `client-go` as discussed. Still consistently failing on port-forwarding, but error is clearer. We're port-forwarding the new scheduler pod right after it comes up (with unique UUID) , so I'm not sure how `9080` could already be bound.

yangwwei:
I think it is because the last port-forwarding wasn't killed so the port was occupied, like @ronazhan mentioned in [this comment](https://github.com/apache/incubator-yunikorn-k8shim/pull/369#issuecomment-1055035338). I have tried the following steps locally:

1. start YK
2. do a port-forwarding
3. kill YK pod

and I get

```
kubectl port-forward svc/yunikorn-service 9889:9889
Forwarding from 127.0.0.1:9889 -> 9889
Forwarding from [::1]:9889 -> 9889
Handling connection for 9889
Handling connection for 9889
E0304 10:12:28.165342   65184 portforward.go:400] an error occurred forwarding 9889 -> 9889: error forwarding port 9889 to pod a1c96c399c12d2a7b14fe58cbe9f320c1a251f80e3bad71e56010890762297e1, uid : Error: No such container: a1c96c399c12d2a7b14fe58cbe9f320c1a251f80e3bad71e56010890762297e1
E0304 10:12:28.166506   65184 portforward.go:400] an error occurred forwarding 9889 -> 9889: error forwarding port 9889 to pod a1c96c399c12d2a7b14fe58cbe9f320c1a251f80e3bad71e56010890762297e1, uid : Error: No such container: a1c96c399c12d2a7b14fe58cbe9f320c1a251f80e3bad71e56010890762297e1
Handling connection for 9889
```

the port is still occupied and I got the following exception if I try to do port-forwarding again:

```
kubectl port-forward svc/yunikorn-service 9889:9889
Unable to listen on port 9889: Listeners failed to create with the following errors: [unable to create listener: Error listen tcp4 127.0.0.1:9889: bind: address already in use unable to create listener: Error listen tcp6 [::1]:9889: bind: address already in use]
error: unable to listen on any of the requested ports: [{9889 9889}]
```

anuraagnalluri:
@yangwwei @ronazhan Right, this makes sense now. Thank you both for your inputs. I forgot that even though we're port-forwarding the pod in the newly added code, the `run-e2e-tests.sh` initiates a service level port-forward that will continue to conflict as you guys mentioned. Will kill the initial port-forward and see if that allows the checks to pass.

anuraagnalluri:
@yangwwei Killing the initial port-forward now, but consistently seeing the following error on 1-2 CI checks (different ones each time):
```
admission webhook "admission-webhook.yunikorn.validate-conf" denied the request: Post "http://yunikorn-service:9080/ws/v1/validate-conf": dial tcp 10.96.250.52:9080: connect: connection refused
  occurred
```
I haven't been able to reproduce this locally. Also the log output from `kClient.PortForwardYkSchedulerPod()` is as expected, since we can see the following at the end of `recovery_and_restart`:
``` 
Forwarding from 127.0.0.1:9080 -> 9080
Forwarding from [::1]:9080 -> 9080
Port-forwarding traffic for yunikorn-scheduler...•
```
Any further tips on troubleshooting?

yangwwei:
Looks like we are close to sorting this out. The remaining issue is an intermittent one.  The failed tests were running right after the restart testing, could it be possible that the services couldn't be fully started or the port-forwarding wasn't fully completed? Maybe we can add some code to wait until the port is accessible before moving to the next test case.

Essentially, we will need to add more coverage in this test suite, we need to verify things like schedule some stuff, restart and make sure the scheduler doesn't lose track of the running apps. In order to run all these verifications, we still need to make sure services are all good.

anuraagnalluri:
@yangwwei Added more verifications as you suggested and checks seem to be passing now. Similar to `basicscheduling_test`, we deploy a sleep pod. After restarting the scheduler, killing previous port-forward, and re-issuing another port-forward command, we get the running applications and verify specific allocation properties.

test/e2e/framework/helpers/k8s/k8s_utils.go
```diff
@@ -141,6 +191,61 @@ func (k *KubeCtl) GetPodNamesFromNS(namespace string) ([]string, error) {
...
+	stream := genericclioptions.IOStreams{
+		In:     os.Stdin,
+		Out:    os.Stdout,
+		ErrOut: os.Stderr,
+	}
+
+	go func() {
+		schedulerPodName, err := k.GetSchedulerPod()
+		if err != nil {
+			panic(err)
```
yangwwei, line 224: Instead of panicking out, when there is an error happened, can we notify the main process and exit with an error? Then the caller can catch this exception and print some more meaningful error messages.

test/e2e/framework/helpers/k8s/k8s_utils.go
```diff
@@ -141,6 +191,61 @@ func (k *KubeCtl) GetPodNamesFromNS(namespace string) ([]string, error) {
...
+				},
+			},
+			LocalPort: portForwardPort,
+			PodPort:   portForwardPort,
+			Streams:   stream,
+			StopCh:    stopCh,
+			ReadyCh:   readyCh,
+		})
+		if err != nil {
+			panic(err)
```
yangwwei, line 241: Same as above

test/e2e/framework/helpers/k8s/k8s_utils.go
```diff
@@ -112,6 +138,30 @@ func (k *KubeCtl) GetPod(name, namespace string) (*v1.Pod, error) {
...
+	for _, podName := range podNameList {
+		if strings.Contains(podName, configmanager.YKScheduler) {
+			return podName, nil
+		}
+	}
+	return "", errors.New("YK scheduler pod not found")
+}
+
+func (k *KubeCtl) KillKubectlProcess() error {
+	cmd := exec.Command("pkill", "-SIGINT", "kubectl")
```
yangwwei, line 155: I noticed the portforward has a [close](https://github.com/kubernetes/client-go/blob/2f52a105e63e9ac7affb1a0f533c62883611ba81/tools/portforward/portforward.go#L412) function, is it possible to leverage that instead of killing kubectl process? The concern is in the future, if the testing env has some other kubectl processes, they could be killed by this as a side effect.

yangwwei (changes requested):
Oh, cool. this is working now! thanks!
Almost look good to me, I have some suggestions with the latest commit, please take a look, thank you @anuraagnalluri !

test/e2e/framework/helpers/k8s/k8s_utils.go
anuraagnalluri: @yangwwei Definitely a valid concern. That function requires a `PortForwarder` and I couldn't figure out how to get an existing one. Only method I found returning that struct was [New](https://github.com/kubernetes/client-go/blob/master/tools/portforward/portforward.go#L156) which creates a new one.

yangwwei: I am not sure if ginkgo has any mechanism that can be used to share context between test suites, can you check that? If not, maybe an option is to use a singleton pattern for the port forward service? What do you think?

anuraagnalluri: `KillKubectlProcess` is invoked to terminate the initial  port-forward [here](https://github.com/apache/incubator-yunikorn-k8shim/blob/master/scripts/run-e2e-tests.sh#L193-L195). For this running port-forward process, I'm not sure how to use [client-go/portforward](https://github.com/kubernetes/client-go/blob/master/tools/portforward/portforward.go) to get the `PortForwarder` instance so we can close its listeners.

We can likely use singleton pattern to maintain a `PortForwarder` once we re-run port-forwarding from this client library. But not sure how to target the main use-case of terminating the initial port-forward regardless of our ability to share context. If we have the ability to share context, can you see that being helpful somehow to get the existing `PortForwarder` (or at least its listeners) in this scenario?

yangwwei: The initial port forwarding code was actually the last step of the installation, I think we can remove it, and use your latest code to create the port forward in golang code. Then we either create a shared context between test suits or use a singleton pattern to maintain one instance to do easier creating/stopping port forward. That might be a more consistent way to manage this. What do you think?

anuraagnalluri:
@yangwwei Apologies for the relative inactivity on this PR, just getting back to it now. Addressed most of your concerns and am going with the singleton approach since I could not find a way to share context between different ginkgo test suites (only within a suite via `Describe` block). Since ginkgo also has the ability to parallelize within a suite, `It` blocks potentially run within their own "containers". Therefore, if port-forwarding is managed by the go runtime, it should be in the setup for all tests now that issue REST calls to the scheduler svc.

If we can verify that all suites execute in independent runtimes (which seems plausible from the log output of the checks), we don't even need to follow singleton pattern since each runtime have its own port forwarder.

With the changes I currently have, some checks intermittently fail due to not being able to get allocations post scheduler restart. Many times, all checks are passing, but there is not 100% pass rate when re-running them. I've printed some debug information which shows that we can see the sleep-pod in the dev namespace for both cases, but the allocations are empty in the failing cases. You can search for "appsInfo allocations are" in failing checks to verify this.

Any thoughts on what the cause of empty allocations when querying `/ws/v1/apps` upon scheduler restart may be? Haven't been able to reproduce the error locally.

yangwwei:
hi @anuraagnalluri  thanks for the updates! Appreciated!
It is sometimes not so easy to figure out what caused the intermittent issues, pls allow me to look into today and tomorrow, I will get back to you on this. Thanks for the efforts going this far! Appreciated!

yangwwei:
hi @anuraagnalluri 

There are a couple of things:

1.  The unit test failure was unrelated to your patch, looks like it was because [here](https://github.com/apache/incubator-yunikorn-k8shim/blob/4fa4544c07e8820031c7acc4dc05fd068c0482c5/pkg/cache/placeholder_manager.go#L154) we need to use a read lock when to get the cleanupTime.  We need a separate JIRA to get this fixed. For this one, as long as we can have a good UT run, we are good.
2. The failed e2e tests are with `--plugin` option, which means they are testing the scheduler-plugin mode. The panic happens at `  	/home/runner/work/incubator-yunikorn-k8shim/incubator-yunikorn-k8shim/test/e2e/recovery_and_restart/recovery_and_restart_test.go:120 +0x156c` which was because the allocations list retrieved from the scheduler was empty (we can also see that from the debug output). Have you looked at the history if it was always failing under the plugin mode? if that's the case, I suggest rerunning this simple scenario in plugin mode and see if it introduces a different behavior after scheduler restart.

anuraagnalluri:
@yangwwei Thanks for your input. I came to a similar realization that intermittent failures would occur on the plugin mode. I did locally spin up scheduler in plugin mode and ran the `recovery_and_restart` test suite, but I was not able to reproduce the error we see in failing checks here.

Specifically, the suite passes and allocations list is non-empty. I'll keep looking in to this issue.

yangwwei:
hi @anuraagnalluri could you please rebase your changes to the latest master? YK now is an Apache TLP, so we have renamed our repos with some related code changes, hope those won't affect this PR.
I also spoke with @ronazhan earlier today, he has lots of expertise on e2e tests, he will help to take a look at this issue as well. Thanks!

anuraagnalluri:
@yangwwei Done, and changed necessary imports. Thanks for getting another pair of eyes on this. I was able to reproduce the error locally a couple times in plugin mode, but am still unsure why the allocations list is empty. 

When I ran in to the same failure as we see in CI checks, I was able to verify that the applicationID of the sleepjob pod belongs to the newly added `recovery_and_restart` suite and _not_ `basic_scheduling_test`. My initial thought was that a "completed" sleepjob with 0 allocations from a previous test could have been picked up, but this is not the case (as that test also tears down the namespace in cleanup). I could see the sleeppod was in "Running" state and ultimately could not identify any metadata differences in the failing case vs. when it's deployed in passing test runs.

Is it possible that plugin-mode logic could specifically affect this behavior in a way normal mode cannot?

test/e2e/recovery_and_restart/recovery_and_restart_test.go
```diff
@@ -0,0 +1,153 @@
...
+	})
+
+	ginkgo.It("Verify_Pod_Alloc_Props", func() {
+		appsInfo, err := restClient.GetAppInfo(sleepRespPod.ObjectMeta.Labels["applicationId"])
+		gomega.Ω(err).NotTo(gomega.HaveOccurred())
+		gomega.Ω(appsInfo).NotTo(gomega.BeNil())
+		ginkgo.By("Verify the pod allocation properties")
+		gomega.Ω(appsInfo["allocations"]).NotTo(gomega.BeNil())
+		for key, val := range appsInfo {
+			fmt.Println("Key:", key, "=>", "Element:", val)
```
anuraagnalluri, line 109: Remove this, for loop block on 113-115, and line 116 after resolving issue where allocations are dropped upon scheduler restart and before merging.

ronazhan:
> Specifically, the suite passes and allocations list is non-empty.

`appsInfo allocations are: []`

@anuraagnalluri Just to clarify, the remaining failure is due to the test expecting a nil object representing an application's allocations, but instead it receives an allocation list of length 0? 

This seems like a minor acceptable behavior change, but @yangwwei can you help confirm this to be expected now? Ideally, this should be consistent between plugin and regular Yunikorn versions. 

For now, I would recommend adding an if statement check to keep the test backwards compatible with previous versions. If the allocations object is not nil, validate that the allocations length is 0.

anuraagnalluri:
> @anuraagnalluri Just to clarify, the remaining failure is due to the test expecting a nil object representing an application's allocations, but instead it receives an allocation list of length 0?

@ronazhan Appreciate your help! This is not quite the issue. The issue is that there is a sleep pod deployed in development namespace [here](https://github.com/apache/yunikorn-k8shim/blob/82302baf7cae11b397e597bd28ba695048169dfc/test/e2e/recovery_and_restart/recovery_and_restart_test.go#L75), so the test actually expects an allocation list of length 1 [here](https://github.com/apache/yunikorn-k8shim/blob/82302baf7cae11b397e597bd28ba695048169dfc/test/e2e/recovery_and_restart/recovery_and_restart_test.go#L117).

Most of the time, this works and we can see the corresponding allocation. Sometimes, but only on plugin mode, this allocation list that's retrieved is of length 0 despite the sleep pod being present/running in the dev namespace. This causes an error when trying to index the first element of the list with `[0]`, but it's unclear why the corresponding allocation isn't showing in the `appsInfo` var which is populated with a `GetAppInfo` call [here](https://github.com/apache/yunikorn-k8shim/blob/82302baf7cae11b397e597bd28ba695048169dfc/test/e2e/recovery_and_restart/recovery_and_restart_test.go#L103).

It's worth noting that the only difference between this and [basic_scheduling_test behavior](https://github.com/apache/yunikorn-k8shim/blob/82302baf7cae11b397e597bd28ba695048169dfc/test/e2e/basic_scheduling/basic_scheduling_test.go#L105-L121) is that we [bounce the scheduler pod](https://github.com/apache/yunikorn-k8shim/blob/82302baf7cae11b397e597bd28ba695048169dfc/test/e2e/recovery_and_restart/recovery_and_restart_test.go#L85) prior to making these checks. We want to ensure that restarting the scheduler won't drop allocations or affect any normal functionality. But this does not seem to be the case 100% of the time specifically in plugin mode.

craigcondit:
> The issue is that there is a sleep pod deployed in development namespace [here](https://github.com/apache/yunikorn-k8shim/blob/82302baf7cae11b397e597bd28ba695048169dfc/test/e2e/recovery_and_restart/recovery_and_restart_test.go#L75), so the test actually expects an allocation list of length 1 [here](https://github.com/apache/yunikorn-k8shim/blob/82302baf7cae11b397e597bd28ba695048169dfc/test/e2e/recovery_and_restart/recovery_and_restart_test.go#L117).
> 
> Most of the time, this works and we can see the corresponding allocation. Sometimes, but only on plugin mode, this allocation list that's retrieved is of length 0 despite the sleep pod being present/running in the dev namespace. This causes an error when trying to index the first element of the list with `[0]`, but it's unclear why the corresponding allocation isn't showing in the `appsInfo` var which is populated with a `GetAppInfo` call [here](https://github.com/apache/yunikorn-k8shim/blob/82302baf7cae11b397e597bd28ba695048169dfc/test/e2e/recovery_and_restart/recovery_and_restart_test.go#L103).
> 
> It's worth noting that the only difference between this and [basic_scheduling_test behavior](https://github.com/apache/yunikorn-k8shim/blob/82302baf7cae11b397e597bd28ba695048169dfc/test/e2e/basic_scheduling/basic_scheduling_test.go#L105-L121) is that we [bounce the scheduler pod](https://github.com/apache/yunikorn-k8shim/blob/82302baf7cae11b397e597bd28ba695048169dfc/test/e2e/recovery_and_restart/recovery_and_restart_test.go#L85) prior to making these checks. We want to ensure that restarting the scheduler won't drop allocations or affect any normal functionality. But this does not seem to be the case 100% of the time specifically in plugin mode.

There's a few things I can see that are issues with the test right away. For one thing, you've structured this as multiple integration tests, when in fact there is really only one. All of the restart / re-forward behavior should go into the setup method, and not be a separate IT test (Ginkgo doesn't guarantee test ordering, so the current structure is brittle, even if it appears to execute correctly).

More importantly, there's nothing I can see here that waits for scheduler recovery to complete. That's not a simple thing to do, but you might try scheduling a second pod (after the restart) and wait for it to be running. Once that happens, it should be safe to query for the original pod.

test/e2e/recovery_and_restart/recovery_and_restart_test.go
```diff
@@ -0,0 +1,161 @@
...
+		gomega.Ω(err).NotTo(gomega.HaveOccurred())
+		gomega.Ω(ns1.Status.Phase).To(gomega.Equal(v1.NamespaceActive))
+
+		ginkgo.By("Deploy the sleep pod to the development namespace")
+		sleepRespPod, err = kClient.CreatePod(common.InitSleepPod(sleepPodConfigs), dev)
+		gomega.Ω(err).NotTo(gomega.HaveOccurred())
+		// Wait for pod to move to running state
+		err = kClient.WaitForPodBySelectorRunning(dev,
+			fmt.Sprintf("app=%s", sleepRespPod.ObjectMeta.Labels["app"]),
+			10)
```
craigcondit, line 82: this should likely be more like 30-60 seconds

test/e2e/recovery_and_restart/recovery_and_restart_test.go
```diff
@@ -0,0 +1,161 @@
...
+		err = kClient.PortForwardYkSchedulerPod()
+		Ω(err).NotTo(gomega.HaveOccurred())
+
+		ginkgo.By("Deploy 2nd sleep pod to the development namespace")
+		sleepRespPod2, err := kClient.CreatePod(common.InitSleepPod(sleepPod2Configs), dev)
+		gomega.Ω(err).NotTo(gomega.HaveOccurred())
+		// Wait for pod to move to running state
+		err = kClient.WaitForPodBySelectorRunning(dev,
+			fmt.Sprintf("app=%s", sleepRespPod2.ObjectMeta.Labels["app"]),
+			10)
```
craigcondit, line 106: this as well, probably 60 seconds, since it may take some time for recovery to complete.

craigcondit (changes requested):
A couple minor issues with timing.

test/e2e/recovery_and_restart/recovery_and_restart_test.go
anuraagnalluri: Can change but I kept the 10 seconds to mirror basic_scheduling_test. Even for deployment of the 2nd pod, we are already portforwarding a running scheduler pod. That seems to be the only precondition for running a 10 second wait in the other test. Since the precondition is the same here, do we want to introduce a different timeout value?

Happy to change to the suggested 30-60 seconds if you still think it’s a good idea.

anuraagnalluri:
@craigcondit Thanks a lot for your inputs. You're absolutely right about the multi-integ test structure. I realize Ginkgo runs its `It` blocks in independent "containers" with no guarantees on ordering. I also wrongly assumed that a "running" scheduler guarantees its state is updated, but it makes sense that a "running" second deployment would be a better verification for that.

I believe the checks are working now and am willing to change the timeouts if you think they should diverge from `basic_scheduling_test`. Thanks a lot for your review :)

craigcondit:
I think the timeout for the final pod status should probably be 30-60 seconds. We need to allow time for recovery and that might take a little while. A normal pod with an already running scheduler will probably be scheduled in 10 seconds reliably but one on a just-started and not yet recovered one may not.

craigcondit (approved):
+1 LGTM. I'll merge this in shortly.

PR #371, opened 2022-02-21
[YUNIKORN-1076] Emit K8s event for invalid Task Group annotation

steinsgateted:
### What is this PR for?
add pod event  for `GetTaskGroupsFromAnnotation` error

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-1076

### Questions:

### DIFF
--- a/pkg/appmgmt/general/general.go
+++ b/pkg/appmgmt/general/general.go
@@ -32,11 +32,11 @@ import (
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/client"
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/common"
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/constants"
+	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/events"
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/utils"
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/conf"
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/log"
 	"github.com/apache/incubator-yunikorn-scheduler-interface/lib/go/si"
-
 	"go.uber.org/zap"
 )
 
@@ -146,6 +146,8 @@ func (os *Manager) getAppMetadata(pod *v1.Pod) (interfaces.ApplicationMetadata, 
 				zap.String("namespace", pod.Namespace),
 				zap.String("name", pod.Name),
 				zap.Error(err))
+			events.GetRecorder().Eventf(pod, nil, v1.EventTypeWarning, "TaskGroupFormatError", "TaskGroupFormatError",
+				"unable to get taskGroups for pod, pod namespace: %s, pod name: %s, reason: %s", appID, pod.Namespace, pod.Name, err.Error())
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


wilfred-s:
I am happy with the event added.

@yuchaoran2011 & @HuangTing-Yao:
However when I look at the original jira [YUNIKORN-1076](https://issues.apache.org/jira/browse/YUNIKORN-1076) the request was to look at rejecting the pod. The point that we pull out all this task group detail from the pod is part of the `addPod()` processing in the manager. We cannot reject the pod at that point as we do not have an application yet. The `addPod()` will create a new application. At the moment that even happens if the parsing fails. Having the application is again not enough. The application state machine does not allow rejecting an application in the new state. It must be submitted first and then the core can reject it. There is no internal way to reject the application. We can also not return the parse failure from `getAppMetadata()` as that would cause it to be interpreted as a task which could have all kinds of side effects.

If we are happy with just the event on the pod we can leave it at this. 
I do think that we need to look at this in more detail, however that will end up being a much larger and way more complex change.

yuchaoran2011:
@wilfred-s Thanks for your thoughts. We need to think more about it. Not sure if rejecting an app simply because its gang annotation has issues is the best UX. We should change the title of this JIRA to something like "Emit events when task group annotation is invalid" to indicate its smaller scope, or create a sub-task for this PR

HuangTing-Yao:
Agree with @wilfred-s and @yuchaoran2011 
Just like @yuchaoran2011 said, We need rename the JIRA and create another one for application level.
I think we need more discussion about the solution.

wilfred-s:
Based on the comments I will commit this making it just an event emit jira.
I have created a follow up jira to look at the reject side of the original jira

pkg/appmgmt/general/general.go
```diff
@@ -146,6 +146,8 @@ func (os *Manager) getAppMetadata(pod *v1.Pod) (interfaces.ApplicationMetadata,
 				zap.String("namespace", pod.Namespace),
 				zap.String("name", pod.Name),
 				zap.Error(err))
+			events.GetRecorder().Eventf(pod, nil, v1.EventTypeWarning, "TaskGroupFormatError", "TaskGroupFormatError",
+				"unable to get taskGroups for pod, pod namespace: %s, pod name: %s, reason: %s", appID, pod.Namespace, pod.Name, err.Error())
```
wilfred-s, line 150: I think we have a discrepancy here between the number of args passed in and the formatting directives in the string:
```
unable to get taskGroups for pod, pod namespace: <appID>, pod name: <pod.Namespace>, reason: pod.Name
```
I think there is a `%s` missing between "pod" and the first comma

wilfred-s (changes requested):

pkg/appmgmt/general/general.go
steinsgateted: Sorry I made a mistake. Modified. Thanks.

wilfred-s (approved):
LGTM

PR #375, opened 2022-03-01
[YUNIKORN-1100] Fix scheduler cache inconsistencies

craigcondit:
### What is this PR for?
Fix leaks and inconsistencies in the shim scheduler cache.

Many of these issues have been shown to lead to scheduling failures (such as hitting max Pod limits) because we don't properly cleanup allocations.

- Made Add/Update/Remove Pod/Node handlers idempotent and removed error returns wherever possible
- Ensure RemoveNode properly removes any associated Pod allocations
- Added debugging hooks to pre/post handlers to dump state of scheduler cache
- Handle terminated pods in a similar fashion to removed Pods
- Fix leaking of assumed pods
- Fix duplicate addition of pods to NodeInfo.Pods structure
- Track pod -> node assignments separately from NodeInfo as NodeInfo stores pods in a list and doesn't check for duplicates
- Refactored Add/Update methods to share logic wherever possible
- Log both podKey and podName when updates are made

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1100

Unit tests updated and debug logs show that we are no longer leaking objects.

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -165,11 +165,7 @@ func (ctx *Context) updateNode(oldObj, newObj interface{}) {
 	}
 
 	// update secondary cache
-	if err := ctx.schedulerCache.UpdateNode(oldNode, newNode); err != nil {
-		log.Logger().Error("unable to update node in scheduler cache",
-			zap.Error(err))
-		return
-	}
+	ctx.schedulerCache.UpdateNode(oldNode, newNode)
 
 	// update primary cache
 	ctx.nodes.updateNode(oldNode, newNode)
@@ -194,11 +190,7 @@ func (ctx *Context) deleteNode(obj interface{}) {
 
 	// delete node from secondary cache
 	log.Logger().Debug("delete node from cache", zap.String("nodeName", node.Name))
-	if err := ctx.schedulerCache.RemoveNode(node); err != nil {
-		log.Logger().Error("unable to delete node from scheduler cache",
-			zap.Error(err))
-		return
-	}
+	ctx.schedulerCache.RemoveNode(node)
 
 	// delete node from primary cache
 	ctx.nodes.deleteNode(node)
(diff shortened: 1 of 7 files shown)


pkg/cache/context.go
```diff
@@ -472,25 +457,26 @@ func (ctx *Context) AssumePod(name string, node string) error {
...
 	defer ctx.lock.Unlock()
 
 	if pod, ok := ctx.schedulerCache.GetPod(name); ok {
 		log.Logger().Debug("forget pod", zap.String("pod", pod.Name))
-		return ctx.schedulerCache.ForgetPod(pod)
+		ctx.schedulerCache.ForgetPod(pod)
+		return
 	}
 	log.Logger().Debug("unable to forget pod",
 		zap.String("reason", fmt.Sprintf("pod %s not found in scheduler cache", name)))
```
wilfred-s, line 479: NIT: Can we clean up this logging to just be:
```
log.Logger().Debug("unable to forget pod: not found in cache", zap.String("pod", name))
```

craigcondit: Updated.

yuchaoran2011 (commented):
Is there a point leaving UpdateNode and UpdatePod two-parameter functions? The first parameter (oldNode and oldPod) are never used

craigcondit:
> Is there a point leaving UpdateNode and UpdatePod two-parameter functions? The first parameter (oldNode and oldPod) are never used

Good point. Updated patch to remove the unnecessary parameter.

yuchaoran2011 (approved):
LGTM

wilfred-s (approved):
LGTM

HuangTing-Yao (approved):
LGTM

PR #376, opened 2022-03-04
[YUNIKORN-1101] idempotent remove application and remove task in context

youxuan0714:
### What is this PR for?
The shim context RemoveApplicationInternal and RemoveTask should not return an error if the application or task is not found.

I set the return value of RemoveApplicationInternal and RemoveTask to nil, so that no matter whether the task or application to be deleted exists or not, it will not return error.
Modify TestRemoveApplicationInternal and TestRemoveTask in context.go as expected.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
(https://issues.apache.org/jira/browse/YUNIKORN-1101)

use "make test" command
### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -692,8 +692,11 @@ func (ctx *Context) RemoveTask(appID, taskID string) error {
 	ctx.lock.RLock()
 	defer ctx.lock.RUnlock()
 	if app, ok := ctx.applications[appID]; ok {
-		app.removeTask(taskID)
-		return nil
+		err := app.removeTask(taskID)
+		if err != nil {
+			log.Logger().Info(" not found, it has been deleted",
+				zap.String("task", taskID))
+		}
 	}
 	return nil
 }
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -657,11 +657,14 @@ func (ctx *Context) RemoveApplication(appID string) error {
 	return fmt.Errorf("application %s is not found in the context", appID)
 }
 
-func (ctx *Context) RemoveApplicationInternal(appID string) error {
+func (ctx *Context) RemoveApplicationInternal(appID string) {
 	ctx.lock.Lock()
 	defer ctx.lock.Unlock()
+	if _, exist := ctx.applications[appID]; !exist {
+		log.Logger().Debug("Attempted to remove non-existent application", zap.String("appID", appID))
+		return
+	}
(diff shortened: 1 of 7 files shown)


pkg/cache/context.go
```diff
@@ -660,11 +660,8 @@ func (ctx *Context) RemoveApplication(appID string) error {
 func (ctx *Context) RemoveApplicationInternal(appID string) error {
 	ctx.lock.Lock()
 	defer ctx.lock.Unlock()
-	if _, exist := ctx.applications[appID]; exist {
-		delete(ctx.applications, appID)
-		return nil
-	}
-	return fmt.Errorf("application %s is not found in the context", appID)
+	delete(ctx.applications, appID)
```
0yukali0, line 663: I think we can treat the message that shows the application doesn't exist as a log information.
Like you did in RemoveTask function.
`
if err != nil {
  log.Logger().Info(" not found, it has been deleted",
  zap.String("Application", appID))
  }
`

0yukali0: Hi @wilfred-s , should we keep the message as a log to remind user or just skip the the massage?

wilfred-s: I would remove the whole return value from the call. We do not return errors as part of a `deleteNode` in the context as an example.

We can log a message, however we should not log it at a INFO level. DEBUG would be more appropriate for the log level. The message should then just say that we're removing the application or the task. Independent of the fact that the delete really happened. Again check `deleteNode` changes in YUNIKORN-1100 (#375)

wilfred-s: `RemoveApplicationInternal` `RemoveTask` and `removeTask` all should not return an error. Just be silent and execute the remove

0yukali0: Thanks @wilfred-s, i got it.
Hi @youxuan0714,   please move the message to debug level and change these functions with no error return.
You can review similar modification in YUNIKORN-1100(#375).

youxuan0714: Ok, I will try again, thanks.

wilfred-s (approved):
LGTM
@0yukali0 please check again

0yukali0 (commented):
LGTM too.

0yukali0 (approved):

PR #377, opened 2022-03-04
[YUNIKORN-1102] shim context getTask error is ignored

9501sam:
### What is this PR for?
modify context.getTask() and provide more detail about the reason of error:
* conversion error
* task not found error
and add unit test TestGetTask() for context.getTask()

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1102

using ```make test``` command

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -703,12 +703,14 @@ func (ctx *Context) RemoveTask(appID, taskID string) error {
 func (ctx *Context) getTask(appID string, taskID string) (*Task, error) {
 	ctx.lock.RLock()
 	defer ctx.lock.RUnlock()
-	if app, ok := ctx.applications[appID]; ok {
+	if app := ctx.GetApplication(appID); app != nil {
 		if managedTask, err := app.GetTask(taskID); err == nil {
 			if task, valid := managedTask.(*Task); valid {
 				return task, nil
 			}
+			return nil, fmt.Errorf("task %s cannot convert to *Task", taskID)
 		}
+		return nil, fmt.Errorf("task %s is not found in application %s", taskID, appID)
 	}
 	return nil, fmt.Errorf("application %s is not found in context", appID)
 }
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -700,19 +700,24 @@ func (ctx *Context) RemoveTask(appID, taskID string) error {
 	return fmt.Errorf("application %s is not found in the context", appID)
 }
 
-func (ctx *Context) getTask(appID string, taskID string) (*Task, error) {
+func (ctx *Context) getTask(appID string, taskID string) *Task {
 	ctx.lock.RLock()
 	defer ctx.lock.RUnlock()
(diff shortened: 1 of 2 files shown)


9501sam:
I think the conversion of ```managedTask``` would be always success.
Should I remove the conversion check ?
like:
```go
func (ctx *Context) getTask(appID string, taskID string) (*Task, error) {
	ctx.lock.RLock()
	defer ctx.lock.RUnlock()
	if app := ctx.GetApplication(appID); app != nil {
		if managedTask, err := app.GetTask(taskID); err == nil {
			return task, nil
		}
		return nil, fmt.Errorf("task %s is not found in application %s", taskID, appID)
	}
	return nil, fmt.Errorf("application %s is not found in context", appID)
}
```

wilfred-s (changes requested):
the `context.getTask()` should only return a Task, not an error. The returned error value does not add anything. In case we cannot find the task or there is an error  converting a nil Task should be returned.
All callers of getTask need to be check for a nil Task.

> I think the conversion of `managedTask` would be always success. Should I remove the conversion check ? like:

The conversion should never fail. However based on the fact that we currently still return a `ManagedTask` from `app.GetTask()` we should do the conversion. If the conversion fails we log a DEBUG message and return a nil Task.

9501sam:
OK, I understand !

pkg/cache/context.go
```diff
@@ -700,17 +700,24 @@ func (ctx *Context) RemoveTask(appID, taskID string) error {
 	return fmt.Errorf("application %s is not found in the context", appID)
 }
 
-func (ctx *Context) getTask(appID string, taskID string) (*Task, error) {
+func (ctx *Context) getTask(appID string, taskID string) *Task {
 	ctx.lock.RLock()
 	defer ctx.lock.RUnlock()
-	if app, ok := ctx.applications[appID]; ok {
+	if app := ctx.GetApplication(appID); app != nil {
```
wilfred-s, line 706: Looking at what we log in case of an issue I think we have a flow issue. If the task conversion fails we log 3 things: conversion failed, task not found and app not found. Really confusing. Go coding guidelines also state to prefer to return early and do not indent the correct case, indent the error handling
That would get a better user experience would be to just log the most specific message.
```
app := ctx.GetApplication(appID)
if app == nil {
    log.Logger().....
    return nil
}
managedTask, err := app.GetTask(taskID)
if err != nil {
    log.Logger().....
    return nil
}
task, valid := managedTask.(*Task)
if !valid {
    log.Logger().....
    return nil
}
return task
```

wilfred-s (changes requested):

pkg/cache/context.go
9501sam: OK, that is really more clean, thank you

wilfred-s (approved):
LGTM

PR #380, opened 2022-03-09
[YUNIKORN-1090] remove common.Node from the shim

manirajv06:
### What is this PR for?

Enhanced si_helper methods to create/update/delete node request and use these methods directly instead of using common.Node method wrappers to avoid the unnecessary object creations. Also re-used the methods in other place as well.

### What type of PR is it?
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1090

### Questions:


pkg/cache/nodes.go
```diff
@@ -213,21 +212,22 @@ func (nc *schedulerNodes) updateNode(oldNode, newNode *v1.Node) {
 		return
 	}
 
-	if schedulerNode, ok := nc.nodesMap[newNode.Name]; ok {
-		schedulerNode.ready = common.HasReadyCondition(newNode)
+	schedulerNode, ok := nc.nodesMap[newNode.Name]
+	if ok {
+		schedulerNode.ready = hasReadyCondition(newNode)
```
pbacsko, line 217: The logic here changed slightly. We only send out the update to the proxy if the node is found in the `nodesMap`. I don't think it has substantial effects, just wondering if this is better?

cc @wilfred-s

pbacsko:
Mostly LGTM, I added two minor comments.

pkg/common/si_helper.go
```diff
@@ -177,14 +184,11 @@ func CreateUpdateRequestForUpdatedNode(node Node) si.NodeRequest {
 	return request
 }
 
-func CreateUpdateRequestForDeleteNode(node Node) si.NodeRequest {
+func CreateUpdateRequestForDeleteNode(nodeID string, action si.NodeInfo_ActionFromRM) si.NodeRequest {
```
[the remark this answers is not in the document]
manirajv06: "delete" as generic term for drain or decommission, but it does restore as well. hence named it as DeleteOrRestore. Hope this is ok.

pkg/cache/nodes.go
manirajv06: Yes, it should not have any effects. I don't see a situation wherein entry doesn't exist in `nodesMap `during "update"

pkg/cache/nodes.go
```diff
@@ -213,21 +212,22 @@ func (nc *schedulerNodes) updateNode(oldNode, newNode *v1.Node) {
 		return
```
wilfred-s, line 212: we must update the resources in the cachedNode, might need locking and a setter method:
`cachedNode.capacity = common.GetNodeResource(&newNode.Status)`

pkg/cache/nodes.go
wilfred-s: I do think we need to carefully look at what we do here (referencing new line numbers):
* line 195: we get the node out of the nodesMap store in cachedNode
* line 215: we retrieve the same node again, we already know it CANNOT be nil and ok is `true, we should re-use cachedNode
* line 217: should always be executed, even if resources do not change, should this use a locked setter method?

pkg/cache/nodes.go
```diff
@@ -213,21 +212,22 @@ func (nc *schedulerNodes) updateNode(oldNode, newNode *v1.Node) {
...
 	}
 
-	if schedulerNode, ok := nc.nodesMap[newNode.Name]; ok {
-		schedulerNode.ready = common.HasReadyCondition(newNode)
+	schedulerNode, ok := nc.nodesMap[newNode.Name]
+	if ok {
+		schedulerNode.ready = hasReadyCondition(newNode)
 		if !schedulerNode.ready {
 			log.Logger().Debug("Node is not ready", zap.String("Node name", newNode.Name))
 		}
```
wilfred-s, line 220: we can log the state of the node always

pkg/cache/nodes.go
```diff
@@ -213,21 +212,22 @@ func (nc *schedulerNodes) updateNode(oldNode, newNode *v1.Node) {
...
+		request := common.CreateUpdateRequestForUpdatedNode(newNode.Name, common.GetNodeResource(&newNode.Status),
```
wilfred-s, line 221: should use the cachedNode, does this event need to be generated when ready changes without resource changes?

pkg/cache/nodes.go
```diff
@@ -213,21 +212,22 @@ func (nc *schedulerNodes) updateNode(oldNode, newNode *v1.Node) {
...
+			schedulerNode.occupied, schedulerNode.ready)
+		log.Logger().Info("report updated nodes to scheduler", zap.Any("request", request))
+		if err := nc.proxy.UpdateNode(&request); err != nil {
+			log.Logger().Info("hitting error while handling UpdateNode", zap.Error(err))
+		}
 	} else {
 		log.Logger().Error("Unable to find scheduler node in nodes map", zap.String("node name",
```
wilfred-s, line 228: can be removed node is always found

pkg/cache/nodes.go
```diff
@@ -259,3 +258,12 @@ func (nc *schedulerNodes) schedulerNodeEventHandler() func(obj interface{}) {
 		}
 	}
 }
+
+func hasReadyCondition(node *v1.Node) bool {
+	for _, condition := range node.Status.Conditions {
```
wilfred-s, line 263: Can `node` or `Status` be nil? Rather be careful than rely on the externally defined object always being complete.

pkg/common/si_helper.go
```diff
@@ -133,18 +134,23 @@ func CreateReleaseAllocationRequestForTask(appID, allocUUID, partition, terminat
 	return result
 }
 
-func CreateUpdateRequestForNewNode(node Node) si.NodeRequest {
+func CreateUpdateRequestForNewNode(nodeID string, capacity *si.Resource, occupied *si.Resource,
```
wilfred-s, line 137: Add comment to the exported function:
```
// CreateUpdateRequestForNewNode builds a NodeRequest ....
```
Should do the same for all si_helper functions changed here.

pkg/common/si_helper.go
```diff
@@ -133,18 +134,23 @@ func CreateReleaseAllocationRequestForTask(appID, allocUUID, partition, terminat
...
+	existingAllocations []*si.Allocation, labels string, ready bool) si.NodeRequest {
 	// Use node's name as the NodeID, this is because when bind pod to node,
 	// name of node is required but uid is optional.
 	nodeInfo := &si.NodeInfo{
-		NodeID:              node.name,
-		SchedulableResource: node.capacity,
+		NodeID:              nodeID,
+		SchedulableResource: capacity,
+		OccupiedResource:    occupied,
 		// TODO is this required?
```
wilfred-s, line 145: Please remove this TODO

pkg/common/si_helper.go
```diff
@@ -156,15 +162,16 @@ func CreateUpdateRequestForNewNode(node Node) si.NodeRequest {
 	return request
 }
 
-func CreateUpdateRequestForUpdatedNode(node Node) si.NodeRequest {
+func CreateUpdateRequestForUpdatedNode(nodeID string, capacity *si.Resource, occupied *si.Resource,
+	ready bool) si.NodeRequest {
 	// Currently only includes resource in the update request
```
wilfred-s, line 167: Please update this comment as it is incorrect

pkg/common/si_helper_test.go
```diff
@@ -18,14 +18,21 @@
...
 	"testing"
 
+	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/constants"
 	"github.com/apache/incubator-yunikorn-scheduler-interface/lib/go/common"
+	"github.com/apache/incubator-yunikorn-scheduler-interface/lib/go/si"
+
 	"gotest.tools/assert"
+
 	v1 "k8s.io/api/core/v1"
 	apis "k8s.io/apimachinery/pkg/apis/meta/v1"
```
wilfred-s, line 31: Can we re-order as we are cleaning up, 3 groups separated by an empty line:
- group system imports
- group other imports
- group yunikorn imports

wilfred-s (changes requested):
One other main point for the `updateNode`:
* line 203-208: if we need to drain or restore we send an update to the core to do that and keep processing: we again look up the node during the calls in the nodesMap (WHY ???) and do some checks. That needs to be optimised out and simplified (remove functions drainNode and restoreNode) We also do not need an `if..else` construct as the checks can never be true at the same time:
```
	if !oldNode.Spec.Unschedulable && newNode.Spec.Unschedulable {
		&& cachedNode.getNodeState() == events.States().Node.Healthy {
			dispatcher.Dispatch(CachedSchedulerNodeEvent{
				NodeID: cachedNode.name,
				Event:  events.DrainNode,
			})
	}
	if oldNode.Spec.Unschedulable && !newNode.Spec.Unschedulable
		&& cachedNode.getNodeState() == events.States().Node.Draining {
			dispatcher.Dispatch(CachedSchedulerNodeEvent{
				NodeID: cachedNode.name,
				Event:  events.RestoreNode,
			})
	}
```

pkg/cache/nodes.go
manirajv06: Taken care here itself even though YUNIKORN-1091 has been filed for the same

manirajv06:
Observed few more things while doing this clean up:

1. I can see few more places where SchedulerNode properties gets updates without having lock mechanism in place. For example, `schedulerNode.occupied = common.Add(schedulerNode.occupied, resource)` in `updateNodeOccupiedResources` method
2. I think we should start exploring and see if `sync.Map` makes sense instead of using `map` given the no. of reads and writes it is going through.

Please share your views.

wilfred-s:
> Observed few more things while doing the clean up:
> 
> 1. I can see few more places where SchedulerNode properties gets updated without having proper lock mechanism in place. For example, `schedulerNode.occupied = common.Add(schedulerNode.occupied, resource)` in `updateNodeOccupiedResources` method. We should call `setOccupiedResource` here.

We should look at all those and we might need to get this fixed throughout. A follow up jira would be the right approach here.

> 2. I think we should start exploring and see if `sync.Map` makes sense instead of using `map` given the no. of reads and writes it is going through.

The number of writes is minimal. Node updates should not happen at a high rate, not even in large clusters. Reads do occur more often. Optimising the performance of this access would most likely not affect the overall performance. `sync.Map` also has a different behaviour when ranging over entries which could cause small but really difficult to track down bugs.

> 3. Among multiple conditions discussed in https://kubernetes.io/docs/concepts/architecture/nodes/#condition, only "Ready" has been used. Should we use other conditions as well to determine a generic `isNodeHealthy` factor and eventually passing to core rather than depending on only one condition?

That is surely worth considering. We should file a follow up jira and assess what we can do in this area.

wilfred-s (approved):
LGTM
the other points mentioned we should fix via follow up jiras

PR #381, opened 2022-03-09
[YUNIKORN-638] Make placeholder image configurable

pbacsko:
### What is this PR for?
The placeholder image is hard-coded and pulled from a central repository.
In an air gapped environment, this might not be feasible and being able to change it is necessary.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-638

### Questions:

### DIFF
--- a/pkg/cache/placeholder.go
+++ b/pkg/cache/placeholder.go
@@ -20,16 +20,14 @@ package cache
 
 import (
 	"fmt"
-	"os"
 
-	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/apis/yunikorn.apache.org/v1alpha1"
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/constants"
 	"github.com/apache/incubator-yunikorn-k8shim/pkg/common/utils"
-	"github.com/apache/incubator-yunikorn-k8shim/pkg/log"
+	"github.com/apache/incubator-yunikorn-k8shim/pkg/conf"
 )
 
 // MUST: run the placeholder pod as non-root user
@@ -41,7 +39,6 @@ import (
 // launch all the placeholder pods.
 var runAsUser int64 = 1000
 var runAsGroup int64 = 3000
-var placeHolderImage string
 
 type Placeholder struct {
 	appID         string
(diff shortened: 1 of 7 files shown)


pkg/cache/placeholder.go
```diff
@@ -104,3 +108,20 @@ func (p *Placeholder) String() string {
 	return fmt.Sprintf("appID: %s, taskGroup: %s, podName: %s/%s",
 		p.appID, p.taskGroupName, p.pod.Namespace, p.pod.Name)
 }
+
+func getPlaceHolderImage() string {
+	if placeHolderImage == "" {
+		imageEnvSetting := os.Getenv(constants.PlaceHolderImageEnvVar)
```
craigcondit, line 114: Let's use the standard configuration mechanism we already have in place rather than ENV vars here (those should be handled in the Docker container config and passed as arguments). Be sure to update both the shim and plugin setups.

pkg/cache/placeholder.go
```diff
@@ -104,3 +108,20 @@ func (p *Placeholder) String() string {
...
+		if imageEnvSetting != "" {
+			placeHolderImage = imageEnvSetting
+			log.Logger().Info("Using placeholder image from environment variable", zap.String("imageName",
```
craigcondit, line 117: I think we don't need to log here but can just use the config printout that comes in from the shim startup already.

craigcondit (changes requested):
Let's use the same configuration pattern we use for everything else.

wilfred-s:
One thought: what happens if an image is specified that does not exists?
What do we show in the logs, what happens to the application and the gang.
We can fail but we should complain loudly if and when it happens.

Can we test for this? maybe in the e2e tests?

pbacsko:
> One thought: what happens if an image is specified that does not exists? What do we show in the logs, what happens to the application and the gang. We can fail but we should complain loudly if and when it happens.
> 
> Can we test for this? maybe in the e2e tests?

I think this would be a legit testcase.

craigcondit (approved):
+1, LGTM. Also, please add a PR for the release repo to update the helm charts with the new attribute. This acts as user-visible documentation for the property as well as an easy way to configure it.

pbacsko:
> +1, LGTM. Also, please add a PR for the release repo to update the helm charts with the new attribute. This acts as user-visible documentation for the property as well as an easy way to configure it.

There's also a PR for that: https://github.com/apache/incubator-yunikorn-release/pull/83

wilfred-s (approved):
LGTM
Incorrect image details will cause the placeholder pods to fail on startup. It will be clear what is happening. Skipping testing of this feature for now.

PR #382, opened 2022-03-09
[YUNIKORN-1108] Follow-up: Fix race conditions in dispatcher_test.go

pbacsko:
### What is this PR for?
Follow-up on YUNIKORN-1108. Creating a new dispatcher for every test case.
Removed the unnenecessary nil check in initDispatcher().

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1108

### Questions:

### DIFF
--- a/pkg/dispatcher/dispatcher.go
+++ b/pkg/dispatcher/dispatcher.go
@@ -58,7 +58,6 @@ type Dispatcher struct {
 	handlers  map[EventType]func(interface{})
 	running   atomic.Value
 	lock      sync.RWMutex
-	asyncStop chan struct{}
 }
 
 func initDispatcher() {
@@ -69,7 +68,6 @@ func initDispatcher() {
 		stopChan:  make(chan struct{}),
 		running:   atomic.Value{},
 		lock:      sync.RWMutex{},
-		asyncStop: make(chan struct{}),
 	}
 	dispatcher.setRunning(false)
 	DispatchTimeout = conf.GetSchedulerConf().DispatchTimeout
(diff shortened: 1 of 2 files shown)


wilfred-s (changes requested):
In a number of tests we manipulate the dispatcher and use defer calls to fix up the exact changes made one at a time. This could be simplified to just replacing the whole dispatcher with a clean one via a defer call:
```
initDispatcher()
defer initDispatcher()
```
We do not need to track or reset things one by one as we know the dispatcher will be thrown away. Simplifies the code even further. Cleanup diff: +5, -15 lines if I count correctly.

pbacsko:
> In a number of tests we manipulate the dispatcher and use defer calls to fix up the exact changes made one at a time. This could be simplified to just replacing the whole dispatcher with a clean one via a defer call:
> 
> ```
> initDispatcher()
> defer initDispatcher()
> ```
> 
> We do not need to track or reset things one by one as we know the dispatcher will be thrown away. Simplifies the code even further. Cleanup diff: +5, -15 lines if I count correctly.

I ended up with only -15 :) pls check.

pbacsko:
Wow, we still have data races. Let me investigate this further...

pbacsko:
`TestExceedAsyncDispatchLimit()` doesn't even run on its own. There are more problems in the test module.

wilfred-s (approved):
LGTM

PR #384, opened 2022-03-11
[YUNIKORN-1112] Use K8s ENV vars to locate yunikorn service from within admission controller

craigcondit:
### What is this PR for?
In certain networking environments (such as EKS with Calico), the yunikorn-service DNS entry is unavailable within the admission controller when running in the host network. However, we can use Kubernetes ENV vars to locate the service. 

This patch uses `YUNIKORN_SERVICE_SERVICE_HOST` and `YUNIKORN_SERVICE_SERVICE_PORT` in preference to `SCHEDULER_SERVICE_ADDRESS` if present.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1112

Verified with local deployment using the new syntax.

### Questions:

### DIFF
--- a/pkg/plugin/admissioncontrollers/webhook/webhook.go
+++ b/pkg/plugin/admissioncontrollers/webhook/webhook.go
@@ -37,6 +37,8 @@ const (
 	HTTPPort                             = 9089
 	policyGroupEnvVarName                = "POLICY_GROUP"
 	schedulerServiceAddressEnvVarName    = "SCHEDULER_SERVICE_ADDRESS"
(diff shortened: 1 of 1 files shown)


craigcondit:
Withdrawing patch as this will not survive service restarts.

PR #385, opened 2022-03-13
[YUNIKORN-1103]Support fetching queue name from pod annotation yunikorn.apache.org/queue

HuangTing-Yao:
### What is this PR for?
we can only get queue name from pod label before, now we can also get queue name from pod annotation.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1103

UT test

### Questions:


yuchaoran2011 (approved):
LGTM. Please also create a JIRA for the website change to explain that label-based queue name takes precedence over the annotation-based one

yangwwei (approved):
Looks good to me as well. 
Just one thing that as we are moving our constants to https://github.com/apache/incubator-yunikorn-scheduler-interface/blob/master/scheduler-interface-spec.md#following-are-constant-of-spec, shall we do the same for label/annotation stuff?

yangwwei:
@HuangTing-Yao pls create a JIRA to track the doc changes. 
For the SI side, I think I am going to create another JIRA to move constants to SI, to avoid dup code. Thanks

HuangTing-Yao:
> LGTM. Please also create a JIRA for the website change to explain that label-based queue name takes precedence over the annotation-based one

Hi @yuchaoran2011 , I think we can explain it in [Yunikorn-596](https://issues.apache.org/jira/browse/YUNIKORN-596)
In Yunikorn-596, we can have a independence page to introduce all the labels and annotations in Yunikorn, and also explain label-based and annotation-based queue name.

PR #386, opened 2022-03-14
[YUNIKORN-1105] Use absolute units for memory

craigcondit:
### What is this PR for?
Switch to using absolute units (bytes) for memory instead of megabytes (10^6). This avoids rounding errors within the scheduler.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1105

unit tests updated

### Questions:
* [x] - There is breaking changes for older versions.
* [x] - It needs documentation.

### DIFF
--- a/pkg/cache/application_test.go
+++ b/pkg/cache/application_test.go
@@ -664,12 +664,11 @@ func TestSetTaskGroupsAndSchedulingPolicy(t *testing.T) {
 	assert.Equal(t, tg2.Tolerations[0].Effect, v1.TaintEffectNoSchedule)
 	assert.Equal(t, tg2.Tolerations[0].TolerationSeconds, &duration)
 
-	// This is a weird calculation for memory as the request is rounded up before multiplying.
-	// TG1: 500Mi, 10 members: each member 525M -> total 5250M
-	// TG2: 1000Mi, 20 members: each member 1049M -> total 20980
-	// overall usage 5250M + 20980M -> 26230M. This will also be the queue usage so the correct handling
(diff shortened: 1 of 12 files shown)


wilfred-s:
With the fix for YUNIKORN-1121 we do not need any changes in `pkg/shim/scheduler_test.go` The tests pass without changes and it shows correctly that the tasks are not seen as _PodQOSBestEffort_ and we can schedule any pod size.
Should we revert the changes for that file from this PR?

wilfred-s:
With the fix for YUNIKORN-1121 we do not need any changes in `pkg/shim/scheduler_test.go` The tests pass without changes and it shows correctly that the tasks are not seen as _PodQOSBestEffort_ and we can schedule any pod size.
Should we revert the changes for that file from this PR?

craigcondit:
> Should we revert the changes for that file from this PR?

Personally I would still keep them, as the new versions are closer to reality. I think the size difference emphasizes that memory is in bytes, and 100 bytes or so as the previous tests used doesn't make much sense.

wilfred-s (approved):
LGTM

wilfred-s:
Filed a follow up jira to test the `PodQOSBestEffort` in an e2e test [YUNIKORN-1126](https://issues.apache.org/jira/browse/YUNIKORN-1126)

PR #387, opened 2022-03-16
[YUNIKORN-1123] UpdateNode may cause the scheduler to crash

manirajv06:
### What is this PR for?
Sending UpdateNode request without empty nodeAttributes map to decommission the node causes panic on the core side. This bug has been caused while doing clean up as discussed in https://issues.apache.org/jira/browse/YUNIKORN-1090. This pr fixes the problem just by reverting that piece to its original state. 

### What type of PR is it?
* [ ] - Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1123

### Questions:


pbacsko (approved):
+1 LGTM

yuchaoran2011 (approved):

PR #390, opened 2022-03-17
[YUNIKORN-1124] Avoid passing empty nodeAttributes in UpdateNode request

manirajv06:
### What is this PR for?
Shim do not pass the following:
1. Empty NodeAttributes map as part of UpdateRequest (for delete, restore & decommissioning actions). 
2. Removed OccupiedResource and labels from UpdateRequest (create action) as they are not required on the core side

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1124

### Questions:

### DIFF
--- a/pkg/cache/node.go
+++ b/pkg/cache/node.go
@@ -167,8 +167,7 @@ func (n *SchedulerNode) handleNodeRecovery(event *fsm.Event) {
 		zap.String("nodeID", n.name),
 		zap.Bool("schedulable", n.schedulable))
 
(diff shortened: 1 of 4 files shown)


wilfred-s (approved):
LGTM

I was not sure about `OccupiedResource` at first but the node passed in is a "new" node which will not have occupied set ever so it can be removed.

PR #400, opened 2022-03-29
[YUNIKORN-1152] Make sure that --placeHolderImage switch is always properly set

pbacsko:
### What is this PR for?
The command line switch -placeHolderImage can be empty in some cases. Make sure that it's always specified.
Also, this PR removes unnecessary constant duplications from Dockerfiles. Default values are already defined at the source code level.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1152

### Questions:

### DIFF
--- a/deployments/image/plugin/Dockerfile
+++ b/deployments/image/plugin/Dockerfile
@@ -28,11 +28,26 @@ ADD start-yunikorn-plugin.sh /opt/yunikorn/bin/start-yunikorn-plugin.sh
 ADD scheduler-config.yaml /opt/yunikorn/work/scheduler-config.yaml
 RUN chmod 755 /opt/yunikorn/bin/*
 
+ENV CLUSTER_ID "mycluster"
+ENV CLUSTER_VERSION "latest"
+ENV POLICY_GROUP "queues"
+ENV SCHEDULING_INTERVAL "1s"
+ENV LOG_LEVEL "0"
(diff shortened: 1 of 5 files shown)


craigcondit (changes requested):
I don't think we should remove the values here. They act as clear documentation, and the replacement pattern results in values with embedded spaces potentially being interpreted improperly. It's also best practice in Dockerfiles to provide default values for all ENV vars used.

Can we update the PR to add the missing attribute but leave everything else as-is?

deployments/image/plugin/Dockerfile
```diff
@@ -28,26 +28,12 @@ ADD start-yunikorn-plugin.sh /opt/yunikorn/bin/start-yunikorn-plugin.sh
 ADD scheduler-config.yaml /opt/yunikorn/work/scheduler-config.yaml
 RUN chmod 755 /opt/yunikorn/bin/*
 
-ENV CLUSTER_ID "mycluster"
```
craigcondit, line 31: All of these removals should be reverted.

craigcondit (approved):
+1 LGTM (pending pre-commit checks).

PR #401, opened 2022-03-30
[YUNIKORN-1165] Ensure StartPodAllocation() only matches selected node

craigcondit:
### What is this PR for?
Fixes an issue with plugin mode where multiple nodes are returned to the scheduler as matches for a pod.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1165

Unit tests added to cover this case.

### Questions:


wilfred-s (approved):
LGTM
codecov is wrong the context call directly calls the cache so the change is covered.

PR #404, opened 2022-04-05
[YUNIKORN-1162] Ensure occupied node resources are updated on recovery

craigcondit:
### What is this PR for?
Fixes an issue where node recovery doesn't account for occupied resources.

* Revert fix in YUNIKORN-1159
* Ensure common.CreateUpdateRequestForNewNode() handles occupied resources
* Add more debug logging around node recovery
* Fix race condition in occupied resource update by always adding rather than setting
* Avoid transitioning node to Recovering more than once
* Increase recovery timeout to avoid possible failure on large clusters

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1162

### Questions:

### DIFF
--- a/pkg/shim/scheduler.go
+++ b/pkg/shim/scheduler.go
@@ -213,7 +213,7 @@ func (ss *KubernetesShim) recoverSchedulerState(e *fsm.Event) {
 				recoverableAppManagers = append(recoverableAppManagers, m)
 			}
 		}
-		if err := ss.context.WaitForRecovery(recoverableAppManagers, 30*time.Second); err != nil {
+		if err := ss.context.WaitForRecovery(recoverableAppManagers, 5*time.Minute); err != nil {
 			// failed
 			log.Logger().Error("scheduler recovery failed", zap.Error(err))
 			dispatcher.Dispatch(ShimSchedulerEvent{
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -108,7 +108,6 @@ func (ctx *Context) AddSchedulingEventHandlers() {
 	ctx.apiProvider.AddEventHandler(&client.ResourceEventHandlers{
 		Type:     client.PodInformerHandlers,
 		FilterFn: nodeCoordinator.filterPods,
-		AddFn:    nodeCoordinator.addPod,
 		UpdateFn: nodeCoordinator.updatePod,
 		DeleteFn: nodeCoordinator.deletePod,
 	})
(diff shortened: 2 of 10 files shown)


pkg/shim/scheduler.go
```diff
@@ -213,7 +213,7 @@ func (ss *KubernetesShim) recoverSchedulerState(e *fsm.Event) {
 				recoverableAppManagers = append(recoverableAppManagers, m)
 			}
 		}
-		if err := ss.context.WaitForRecovery(recoverableAppManagers, 30*time.Second); err != nil {
+		if err := ss.context.WaitForRecovery(recoverableAppManagers, 5*time.Minute); err != nil {
```
craigcondit, line 216: Increased this to avoid potential issues on large clusters. However, the only way a node should fail to transition is if the scheduler is very badly misconfigured.

craigcondit:
Rebased on master to pull in e2e fixes from #406.

pkg/shim/scheduler.go
yuchaoran2011: Around mid 2021, I used to observe that recovery could fail relatively frequently. When it did, it never recovered and we could benefit from knowing the failure sooner rather than later. That's why @chenya-zhang introduced this change to fail really quickly (in 30 secs). A ton of things have changed over the last year, 5 minutes should be fine. But I'll keep an eye on what if any failed recovery scenarios could occur

yuchaoran2011 (approved):
LGTM. Thanks for the fixes

pkg/shim/scheduler.go
craigcondit: I also debated what to use here for a timeout. I went through the code paths that the recover event goes through, and it seems like the only error that can occur is if the RM is not properly registered with the shim (which should never happen). If it were to happen, it would fail regardless of how long a timeout we provide.

My reasoning for a longer timeout is that while unlikely, it's not impossible that we fail to finish recovery with a large enough cluster. Also, we log every second how many nodes are waiting on recovery, so if recovery fails we should be able to see this in the logs pretty clearly.

Personally, I'd rather not have a timeout at all and just do the recovery synchronously, but that's a more invasive change at this point...

PR #405, opened 2022-04-06
[YUNIKORN-1155] Placeholder timeout should not be initialized to the default value during recovery

pbacsko:
### What is this PR for?
When Yunikorn is restarted, the placeholder timeout is reset to the original value, regardless of how much time the placeholder had already spent in pending/running state. This PR fixes this issue by sending the creationTime attribute of the pod to the core which can make a proper calculation about the remaining time.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1155

### Questions:

### DIFF
--- a/pkg/appmgmt/general/general.go
+++ b/pkg/appmgmt/general/general.go
@@ -113,7 +113,7 @@ func (os *Manager) getTaskMetadata(pod *v1.Pod) (interfaces.TaskMetadata, bool) 
 	}, true
 }
 
-func (os *Manager) getAppMetadata(pod *v1.Pod) (interfaces.ApplicationMetadata, bool) {
+func (os *Manager) getAppMetadataInternal(pod *v1.Pod, recovery bool) (interfaces.ApplicationMetadata, bool) {
 	appID, err := utils.GetApplicationIDFromPod(pod)
 	if err != nil {
 		log.Logger().Debug("unable to get application for pod",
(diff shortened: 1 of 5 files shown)


pkg/appmgmt/general/general.go
```diff
@@ -162,9 +168,14 @@ func (os *Manager) getAppMetadata(pod *v1.Pod) (interfaces.ApplicationMetadata,
 		TaskGroups:                 taskGroups,
 		OwnerReferences:            ownerReferences,
 		SchedulingPolicyParameters: schedulingPolicyParams,
+		CreationTime:               creationTime,
 	}, true
 }
 
+func (os *Manager) getAppMetadata(pod *v1.Pod) (interfaces.ApplicationMetadata, bool) {
```
pbacsko, line 175: I believe it would be cleaner to have only this method with (pod, bool) arguments, but for now I wanted to avoid too many changes. This is a potential TODO.

pkg/cache/context.go
```diff
@@ -612,6 +613,9 @@ func (ctx *Context) AddApplication(request *interfaces.AddApplicationRequest) in
 		ctx.apiProvider.GetAPIs().SchedulerAPI)
 	app.setTaskGroups(request.Metadata.TaskGroups)
 	app.setTaskGroupsDefinition(request.Metadata.Tags[constants.AnnotationTaskGroups])
+	if request.Metadata.CreationTime != 0 {
+		app.tags["creationTime"] = strconv.FormatInt(request.Metadata.CreationTime, 10)
```
pbacsko, line 617: I haven't made SI changes, so passing this as a tag.

wilfred-s (approved):
LGTM

PR #412, opened 2022-04-23
[YUNIKORN-966] Retrieve the username from the SparkApp CRD

steinsgateted:
### What is this PR for?
get the user label on the SparkApp CRD objects

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-966

### Questions:


pkg/common/constants/constants.go
```diff
@@ -50,6 +50,9 @@ const CPU = "vcore"
 const SparkLabelAppID = "spark-app-selector"
 const SparkLabelRole = "spark-role"
 const SparkLabelRoleDriver = "driver"
+const SparkDefaultUser = "root"
```
craigcondit, line 53: Spark should not have a different user than YuniKorn. The default YuniKorn user is "nobody" - see constants.go.

pkg/appmgmt/sparkoperator/spark.go
```diff
@@ -124,3 +128,21 @@ func (os *Manager) deleteApplication(obj interface{}) {
 	log.Logger().Info("spark app deleted", zap.Any("SparkApplication", app))
 	os.amProtocol.NotifyApplicationComplete(app.Status.SparkApplicationID)
 }
+
```
craigcondit, line 131: We need to ensure that the YuniKorn label takes precedence if specified. Otherwise we create a security hole if we have an environment where the username label is forced by policy (such as an admission controller that checks security credentials). Allowing Spark to override this is wrong.

craigcondit (changes requested):
Looks good for the most part, but need a few things:
- Use 'nobody' as the default user as this is the YuniKorn default
- YuniKorn username label should override Spark, not the other way around
- Needs unit tests.

steinsgateted:
@craigcondit Thanks for your review.
I would like some clarifications.
I have no idea how to test `GetProxyUser()`.

pkg/appmgmt/sparkoperator/spark.go
```diff
@@ -124,3 +127,21 @@ func (os *Manager) deleteApplication(obj interface{}) {
...
+func GetProxyUser(pod *v1.Pod) string {
+	if appName, ok := pod.Labels[constants.SparkOperatorLabelAppName]; ok {
+		if crdInformerFactory == nil {
+			log.Logger().Info("Spark operator AppMgmt service is not initialized, so the username from the SparkApp CRD cannot be obtained")
+			return ""
+		}
+		app, err := crdInformerFactory.Sparkoperator().V1beta2().SparkApplications().Lister().SparkApplications(pod.Namespace).Get(appName)
```
craigcondit, line 137: This should not make an API server call but use the informers instead. Basically, cache all the Spark application objects in-memory (track add/update/delete events) and use those to lookup the proxy user.

craigcondit (changes requested):
See review comments.

wilfred-s:
With the changes that we have made for user retrieval via [YUNIKORN-1306](https://issues.apache.org/jira/browse/YUNIKORN-1306) I think we need to rethink what we do here. Can the change to retrieve the user from the CRD be rolled into the admission controller?

wilfred-s:
We need to rethink this approach inline with the changes that we have made for user retrieval via [YUNIKORN-1306](https://issues.apache.org/jira/browse/YUNIKORN-1306) 
Closing

PR #415, opened 2022-05-03
[POC][YUNIKORN-1194] Task objects are not created when applications are restored in recoverApps()

pbacsko:
### What is this PR for?
The code inside `AppManagementService.recoverApps()` does not call `Context.AddTask()`. Which means that the generated `Application` object will be invalid since it won't include any `Task` object in the `taskMap`.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1194

### Questions:

### DIFF
--- a/pkg/appmgmt/appmgmt.go
+++ b/pkg/appmgmt/appmgmt.go
@@ -37,24 +37,28 @@ type AppManagementService struct {
 	apiProvider client.APIProvider
 	amProtocol  interfaces.ApplicationManagementProtocol
 	managers    []interfaces.AppManager
+	generalMgr  *general.Manager
 }
 
 func NewAMService(amProtocol interfaces.ApplicationManagementProtocol,
 	apiProvider client.APIProvider) *AppManagementService {
+	generalMgr := general.NewManager(amProtocol, apiProvider)
 	appManager := &AppManagementService{
 		amProtocol:  amProtocol,
 		apiProvider: apiProvider,
 		managers:    make([]interfaces.AppManager, 0),
+		generalMgr:  generalMgr,
 	}
 
 	log.Logger().Info("Initializing new AppMgmt service")
 
 	if !apiProvider.IsTestingMode() {
 		log.Logger().Info("Registering Spark operator with the AppMgmt service")
+
 		appManager.register(
 			// registered app plugins
 			// for general apps
-			general.NewManager(amProtocol, apiProvider),
+			generalMgr,
 			// for spark operator - SparkApplication
 			sparkoperator.NewManager(amProtocol, apiProvider),
 			// for application crds
--- a/pkg/appmgmt/general/general.go
+++ b/pkg/appmgmt/general/general.go
@@ -90,7 +90,7 @@ func (os *Manager) Stop() {
 	// noop
 }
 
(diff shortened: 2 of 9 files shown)


pkg/appmgmt/appmgmt.go
```diff
@@ -37,24 +37,28 @@ type AppManagementService struct {
...
 }
 
 func NewAMService(amProtocol interfaces.ApplicationManagementProtocol,
 	apiProvider client.APIProvider) *AppManagementService {
+	generalMgr := general.NewManager(amProtocol, apiProvider)
 	appManager := &AppManagementService{
 		amProtocol:  amProtocol,
 		apiProvider: apiProvider,
 		managers:    make([]interfaces.AppManager, 0),
+		generalMgr:  generalMgr,
```
pbacsko, line 50: Needed for `GetTaskMetaData()`.

pkg/appmgmt/general/general.go
```diff
@@ -113,7 +113,7 @@ func (os *Manager) getTaskMetadata(pod *v1.Pod) (interfaces.TaskMetadata, bool)
 	}, true
 }
 
-func (os *Manager) getAppMetadata(pod *v1.Pod, recovery bool) (interfaces.ApplicationMetadata, bool) {
+func (os *Manager) GetAppMetadata(pod *v1.Pod, recovery bool) (interfaces.ApplicationMetadata, bool) {
```
pbacsko, line 116: Changed the visibility to be consistent with `GetTaskMetaData()`, but not strictly necessary.

pkg/cache/context.go
```diff
@@ -595,7 +595,9 @@ func (ctx *Context) getNamespaceObject(namespace string) *v1.Namespace {
 
 func (ctx *Context) AddApplication(request *interfaces.AddApplicationRequest) interfaces.ManagedApp {
 	log.Logger().Debug("AddApplication", zap.Any("Request", request))
-	if app := ctx.GetApplication(request.Metadata.ApplicationID); app != nil {
+	appID := request.Metadata.ApplicationID
+	if app := ctx.GetApplication(request.Metadata.ApplicationID); app != nil && !request.ForceUpdate {
+		log.Logger().Debug("Application already in the context", zap.String("appID", appID))
```
pbacsko, line 600: This part will likely change after YUNIKORN-1186. With proper coordination, the replacement won't be needed. But in this case, the patch stands on its own.

pkg/appmgmt/general/general.go
```diff
@@ -90,7 +90,7 @@ func (os *Manager) Stop() {
 	// noop
 }
 
-func (os *Manager) getTaskMetadata(pod *v1.Pod) (interfaces.TaskMetadata, bool) {
+func (os *Manager) GetTaskMetadata(pod *v1.Pod) (interfaces.TaskMetadata, bool) {
```
pbacsko, line 93: I tried moving both methods to `utils`, but ended up with circular deps. So leaving them here.

pbacsko:
Update: I'm working on YUNIKORN-1186 and as far as I can see, fixing that also fixes this. So no need to review it right now, I'll post further updates whether it's necessary or not.

PR #421, opened 2022-05-12
[YUNIKORN-1197] Placeholders are immediately replaced during recovery

pbacsko:
### What is this PR for?
In certain cases, running placeholders are immediately replaced if Yunikorn is restarted. It's because the `Application` in on the shim side does not stay in `Reserving` state, but transitions to `Running`. This causes resource asks to be added and the scheduler replaces the placeholder pods.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1197

### Questions:

### DIFF
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -336,6 +336,19 @@ func (app *Application) GetAllocatedTasks() []*Task {
 	return app.getTasks(events.States().Task.Allocated)
 }
 
+func (app *Application) GetPlaceHolderTasks() []*Task {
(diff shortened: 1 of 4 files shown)


wilfred-s (approved):
LGTM, test will be added as part of YUNIKORN-1198

PR #428, opened 2022-05-31
[YUNIKORN-1207] [Shim] preempt pods based on the priorities

manirajv06:
### What is this PR for?

Mark the task as "originator" after identify the same as first pod and passes the same info to core via SI message.

### What type of PR is it?
* [ ] - Feature

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1207

### Questions:

### DIFF
--- a/pkg/cache/amprotocol_mock.go
+++ b/pkg/cache/amprotocol_mock.go
@@ -86,7 +86,6 @@ func (m *MockedAMProtocol) AddTask(request *interfaces.AddTaskRequest) interface
 			// If yes, then make it as "first pod/owner/driver" of the application and set the task as originator
 			for _, ownerReference := range app.getPlaceholderOwnerReferences() {
 				referenceID := string(ownerReference.UID)
-				log.Logger().Info("task added", zap.String("dd", request.Metadata.TaskID), zap.String("dd", referenceID))
 				if request.Metadata.TaskID == referenceID {
 					originator = true
 					ownerReferenceUID = referenceID
@@ -94,14 +93,12 @@ func (m *MockedAMProtocol) AddTask(request *interfaces.AddTaskRequest) interface
 				}
 			}
 			task := NewFromTaskMeta(request.Metadata.TaskID, app, nil, request.Metadata, originator)
-
 			app.addTask(task)
 			log.Logger().Info("task added",
 				zap.String("appID", app.applicationID),
 				zap.String("taskID", task.taskID),
 				zap.String("taskState", task.GetTaskState()))
-			if app.getOriginatingTask() == nil {
-				log.Logger().Info("task added", zap.String("dd", ownerReferenceUID))
+			if app.GetOriginatingTask() == nil {
 				if task, taskErr := app.GetTask(ownerReferenceUID); task != nil && taskErr == nil {
 					log.Logger().Info("app request originating pod added",
 						zap.String("appID", app.applicationID),
--- a/pkg/cache/amprotocol_mock.go
+++ b/pkg/cache/amprotocol_mock.go
@@ -81,15 +81,16 @@ func (m *MockedAMProtocol) AddTask(request *interfaces.AddTaskRequest) interface
 		existingTask, err := app.GetTask(request.Metadata.TaskID)
 		if err != nil {
 			var originator bool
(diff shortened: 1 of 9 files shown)


pkg/cache/amprotocol_mock.go
```diff
@@ -71,16 +74,39 @@ func (m *MockedAMProtocol) RemoveApplication(appID string) error {
...
+					break
+				}
+			}
+			task := NewFromTaskMeta(request.Metadata.TaskID, app, nil, request.Metadata, originator)
 			app.addTask(task)
+			log.Logger().Info("task added",
+				zap.String("appID", app.applicationID),
+				zap.String("taskID", task.taskID),
+				zap.String("taskState", task.GetTaskState()))
+			if app.GetOriginatingTask() == nil {
```
pbacsko, line 97: Just curious... If we've already called `setOriginatingTask()` then this no longer returns `nil`. Also, we set this if `originator == true` and we have only one originator, right? So to me, it's cleaner to read sth like:

```
if originator {
  app.setOriginatingTask(task)
  log.Logger().Info("app request originating pod added", zap.String("original task", task.GetTaskID()))
}
```
If `originator` is true and `app.GetTask(ownerReferenceUID)` returns an existing task, that means an inconsistent state, isn't it? 

If I'm correct, I'd rather simplify this and handle the `task != nil` condition with an error, because it's something that's not supposed to happen:

```
if originator {
  if app.GetOriginatingTask() != nil {
    log.Logger.Error("Inconsistent state - found another originator task for an application", zap.String("taskId", task.GetTaskID())
  }
  app.setOriginatingTask(task)
  log.Logger().Info("app request originating pod added", zap.String("original task", task.GetTaskID()))
}
```

pkg/cache/context.go
```diff
@@ -691,21 +691,30 @@ func (ctx *Context) AddTask(request *interfaces.AddTaskRequest) interfaces.Manag
...
-				if app.getOriginatingTask() == nil {
-					for _, ownerReference := range app.getPlaceholderOwnerReferences() {
-						if task, taskErr := app.GetTask(string(ownerReference.UID)); task != nil && taskErr == nil {
-							log.Logger().Info("app request originating pod added",
-								zap.String("appID", app.applicationID),
-								zap.String("original task", task.GetTaskID()))
-							app.setOriginatingTask(task)
-							break
-						}
+				if app.GetOriginatingTask() == nil {
```
pbacsko, line 712: Just curious... If we've already called `setOriginatingTask()` then this no longer returns `nil`. Also, we set this if `originator == true` and we have only one originator, right? So to me, it's cleaner to read sth like:

```
if originator {
  app.setOriginatingTask(task)
  log.Logger().Info("app request originating pod added", zap.String("original task", task.GetTaskID()))
}
```
If `originator` is true and `app.GetTask(ownerReferenceUID)` returns an existing task, that means an inconsistent state, isn't it? 

If I'm correct, I'd rather simplify this and handle the `task != nil` condition with an error, because it's something that's not supposed to happen:

```
if originator {
  if app.GetOriginatingTask() != nil {
    log.Logger.Error("Inconsistent state - found another originator task for an application", zap.String("taskId", task.GetTaskID())
  }
  app.setOriginatingTask(task)
  log.Logger().Info("app request originating pod added", zap.String("original task", task.GetTaskID()))
}
```

pkg/cache/context.go
```diff
@@ -692,22 +692,34 @@ func (ctx *Context) AddTask(request *interfaces.AddTaskRequest) interfaces.Manag
...
+				var originator bool
+
+				// Is this task the originator of the application?
+				// If yes, then make it as "first pod/owner/driver" of the application and set the task as originator
+				if app.GetOriginatingTask() == nil {
+					for _, ownerReference := range app.getPlaceholderOwnerReferences() {
+						referenceID := string(ownerReference.UID)
+						if request.Metadata.TaskID == referenceID {
+							originator = true
+							break
```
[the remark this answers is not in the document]
manirajv06: Test cases has been already added in general_test.go. I think it is good enough.

pbacsko:
+1

PR #430, opened 2022-06-01
[YUNIKORN-1227] Snapshot scheduler cache when running predicates

craigcondit:
### What is this PR for?
Fixes a race condition caused by K8s predicates expecting a non-changing set of nodes / pods. This PR snapshots the state of the scheduler cache so that this can drive the predicate logic. The snapshots are only created when needed, and are kept around as long as the cache has not changed.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1227

New unit tests added, e2e tests passing successfully.

### Questions:

### DIFF
--- a/pkg/cache/external/scheduler_cache_test.go
+++ b/pkg/cache/external/scheduler_cache_test.go
@@ -322,7 +322,28 @@ func TestSnapshot(t *testing.T) {
 	snapshot2 := cache.Snapshot()
 	assert.Check(t, snapshot == snapshot2, "snapshots are not the same")
 
-	pod1 := &v1.Pod{
+	// add node
+	resourceList := make(map[v1.ResourceName]resource.Quantity)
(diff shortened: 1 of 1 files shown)


manirajv06 (approved):
LGTM.

By any chance, can we get the data race issue in plugin mode?  I don't think so as I don't see using snapshot in scheduler_plugin.go. Just trying to confirm..

pbacsko:
A qq from me: in some code paths, `cache.snapshot = nil` is not covered, isn't that important?

craigcondit:
> A qq from me: in some code paths, `cache.snapshot = nil` is not covered, isn't that important?

It is, but that's a side effect of many of those code paths not being covered in unit tests at all. That's a bigger issue than this PR is attempting to address.

PR #431, opened 2022-06-02
[YUNIKORN-1217] Ensure that Spark driver pod is processed before executor pods during recovery

pbacsko:
### What is this PR for?
We have to make sure that we always call `addPod()` to the Spark driver pod before the executors (per application) when doing recovery, because that contains annotations which executors don't.
In general, adding the very first pod of the existing application makes sense.

This PR contains the simplest solution: after retrieving the current list of pods from K8s, just sort them based on CreationTime.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1217

### Questions:

### DIFF
--- a/pkg/appmgmt/appmgmt_recovery.go
+++ b/pkg/appmgmt/appmgmt_recovery.go
@@ -20,6 +20,7 @@ package appmgmt
 
 import (
 	"fmt"
+	"sort"
 	"time"
 
 	"go.uber.org/zap"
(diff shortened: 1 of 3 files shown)


wilfred-s (approved):
LGTM

PR #432, opened 2022-06-06
[YUNIKORN-1228] Fix race condition when serializing K8s objects

craigcondit:
### What is this PR for?
Fixes a race condition that occurs when K8s objects are serialized for transmission. If the `version` component of the `GroupVersionKind` element (which is part of every K8s object) is not set, the serializer sets it, causing race conditions with concurrent reads. This patch ensures that we defensively copy objects that are sent over the wire.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1228

No functional changes, so all existing unit and e2e tests apply.

### Questions:

### DIFF
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -398,7 +398,7 @@ func (app *Application) scheduleTasks(taskScheduleCondition func(t *Task) bool) 
 					log.Logger().Warn("init task failed", zap.Error(err))
 				}
 			} else {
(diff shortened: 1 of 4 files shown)


craigcondit:
Implementation note: I realize there are instances in this patch where an object is cloned twice in a single method (once per I/O call). This is due to the fact that many of the I/O methods are async in nature and so this ensures that we have completely independent copies for events in flight.

pbacsko (approved):
+1 lgtm

PR #433, opened 2022-06-14
[YUNIKORN-1235] pass priority on pods from shim into the core

manirajv06:
### What is this PR for?
Extract `Priority` from Pod spec and pass onto core via SI message

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1235

### Questions:

### DIFF
--- a/pkg/common/test/recoverable_apps_mock.go
+++ b/pkg/common/test/recoverable_apps_mock.go
@@ -19,7 +19,7 @@
 package test
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 
 	"github.com/apache/yunikorn-k8shim/pkg/common/constants"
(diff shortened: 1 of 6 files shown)


wilfred-s (changes requested):
The lint failure must be fixed.
Instead of using `assert.Assert(t, <pointer> != nil)` we should use an if structure to make it clear that you test against nil:
```
if <pointer> == nil {
    t.Fail("unexpected nil pointer")
}
```

pkg/common/test/recoverable_apps_mock.go
```diff
@@ -19,7 +19,7 @@
 package test
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
pbacsko, line 22: Nit: why this change?

PR #435, opened 2022-06-21
[YUNIKORN-1244] Fix performance regression  after YUNIKORN-1227

craigcondit:
### What is this PR for?
Reverts the cache snapshotting functionality introduced in YUNIKORN-1227 in
favor of locking the cache.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1244

E2e tests still pass, and race condition is not observed. Performance impact of YUNIKORN-1227 is irrelevant as that change is reverted as part of this fix.

### Questions:


yangwwei:
hi @craigcondit  thanks for the fix, this looks good.

Just curious about this:

> With only 1 node / 1 pod allocated, the cost of a new snapshot is 27us on a test machine. With 10k nodes and 10k pods this time grows to 41ms, and with 1k nodes and 100k pods the snapshot takes 332ms.

you got these numbers by constructing some pod/node objects, correct? When I run into this issue, I feel it was much worse than this with about ~1K nodes and ~20k pods.

yangwwei (approved):
+1

wilfred-s:
> > With only 1 node / 1 pod allocated, the cost of a new snapshot is 27us on a test machine. With 10k nodes and 10k pods this time grows to 41ms, and with 1k nodes and 100k pods the snapshot takes 332ms.
> 
> you got these numbers by constructing some pod/node objects, correct? When I run into this issue, I feel it was much worse than this with about ~1K nodes and ~20k pods.

What you would have seen is the impact of continuous snapshot generation, not just a one off. The one off times are bad by itself but even if 1k nodes and 20k pods "only" takes about 50ms the repeated snapshots cause issues. During a normal scheduling cycle to get one pod on a node we might need to make multiple snapshots as things change while the checks are performed. This would cause the slow down to be larger than what you could expect to see from taking just one snapshot.
 
It is not just the one-off time that is important it is the frequency at which we need to make them that define the impact.

wilfred-s (approved):
LGTM

Filed a follow up YUNIKORN-1245 to clean up the `GetNodesInfoMap` call.

PR #439, opened 2022-06-30
[YUNIKORN-1243] Build replacement web UI for e2e tests

wilfred-s:
### What is this PR for?
Build a simple web UI replacement image so we no longer rely on a web
image with the latest tag in the public hub.
The web image runs as a non root user and can run as a dynamic user on
RedHat Openshift.

Includes major cleanup of the run_e2e_tests.sh script
Makefile comment fixes to line up with the actions

### What type of PR is it?
* [X] - Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1243

covered during standard PR tests


craigcondit (approved):
+1 LGTM.

PR #440, opened 2022-06-30
[YUNIKORN-1248] Support aarch64 and arm64 for ARM processors

wilfred-s:
### What is this PR for?
The M1 MacBook does not identify as aarch64 but as arm64. Both aarch64
and arm64 are the same and are supported as such.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1248

Build locally on a M1 mac

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,18 +64,23 @@ endif
 # Build architecture settings:
 # EXEC_ARCH defines the architecture of the executables that gets compiled
 # DOCKER_ARCH defines the architecture of the docker image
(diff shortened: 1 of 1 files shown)


0yukali0:
LGTM.
It works on my mac containing M1 processor.
截圖 2022-06-30 下午6 12 30

0yukali0 (approved):

PR #442, opened 2022-07-14
[YUNIKORN-1257] docker build improvements

wilfred-s:
### What is this PR for?
Build optimisation with reduced data copies and image layers.
Incorporates Openshift dynamic user support.

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1257

Image build must still provide working image, tested locally


craigcondit (changes requested):
Looks good, with one minor nit: Let's move the creation of the $HOME dir to before addgroup/adduser. Adduser complains if the home dir doesn't exist (but still succeeds).

craigcondit (approved):
+1 LGTM.

PR #443, opened 2022-07-30
[YUNIKORN-1265] fix shellcheck issues in our shell scripts

lowc1012:
### What is this PR for?
Fix issues and add shellcheck into ci

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Fix issues and add shellcheck into CI for other repos

### What is the Jira issue?
[YUNIKORN-1265](https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-1265)

### Questions:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,34 +115,46 @@ $(LINTBIN):
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(LINTBASE) v1.46.2
 	stat $@ > /dev/null 2>&1
 
-.PHONY:
-SHELLCHECK_PATH := "$(BASE_DIR)shellcheck"
-SHELLCHECK_VERSION := "v0.8.0"
-SHELLCHECK_ARCHIVE := "shellcheck-$(SHELLCHECK_VERSION).$(OS).$(HOST_ARCH).tar.xz"
-install_shellcheck:
-	@if ! command -v -- "shellcheck" > /dev/null 2>&1 && [ ! -e $(SHELLCHECK_PATH) ]; then \
-		echo "shellcheck is not installed"; \
-		if [ "$(HOST_ARCH)" = "arm64" ]; then \
-			echo "arm is not supported"; \
-			exit 1; \
-		else \
-			echo "Installing shellcheck $(SHELLCHECK_VERSION) ..." \
-			&& curl -sSfL https://github.com/koalaman/shellcheck/releases/download/$(SHELLCHECK_VERSION)/$(SHELLCHECK_ARCHIVE) | tar -x -J --strip-components=1 shellcheck-$(SHELLCHECK_VERSION)/shellcheck \
-			&& echo "shellcheck $(SHELLCHECK_VERSION) has been installed at $(SHELLCHECK_PATH)"; \
-		fi \
-	else \
-		echo "shellcheck has been installed"; \
-	fi
-
-.PHONY: lint
 # Run lint against the previous commit for PR and branch build
 # In dev setup look at all changes on top of master
+.PHONY: lint
 lint: $(LINTBIN)
 	@echo "running golangci-lint"
 	git symbolic-ref -q HEAD && REV="origin/HEAD" || REV="HEAD^" ; \
 	headSHA=$$(git rev-parse --short=12 $${REV}) ; \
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -358,3 +358,37 @@ arch:
...
 	cd ./test/e2e && ginkgo -r -v -timeout=2h -- -yk-namespace "yunikorn" -kube-config $(KUBECONFIG)
+
+.PHONY: install_shellcheck
+SHELLCHECK_PATH := $(BASE_DIR)scripts/shellcheck
+SHELLCHECK_VERSION := "v0.8.0"
+SHELLCHECK_ARCHIVE := "shellcheck-$(SHELLCHECK_VERSION).$(OS).$(HOST_ARCH).tar.xz"
+install_shellcheck: $(eval SHELL:=/bin/bash)
+	@if [[ ! -f $(SHELLCHECK_PATH) && "$(shell which shellcheck)" == "" ]]; then \
+		echo "shellcheck is not installed"; \
+		if [[ $(HOST_ARCH) == "arm64" ]]; then \
```
wilfred-s, line 369: `which` could be a builtin or an external command. Behaviour differs between shells zsh always says "not found" and breaks this detection. `command -v exe-name` is portable

Makefile
```diff
@@ -358,3 +358,37 @@ arch:
...
+        	echo "Can't find shellcheck for arm64 from https://github.com/koalaman/shellcheck/releases/$(SHELLCHECK_VERSION)"; \
```
wilfred-s, line 370: Simple failure message is enough: arm is not supported

Makefile
```diff
@@ -358,3 +358,37 @@ arch:
...
+			exit 1; \
+		else \
+			echo "Installing shellcheck $(SHELLCHECK_VERSION) ..." \
+			&& set -x \
```
wilfred-s, line 374: Keep it clean no need to show details of the process

Makefile
```diff
@@ -358,3 +358,37 @@ arch:
...
+			&& wget https://github.com/koalaman/shellcheck/releases/download/$(SHELLCHECK_VERSION)/$(SHELLCHECK_ARCHIVE) \
+			&& tar -xf $(SHELLCHECK_ARCHIVE) shellcheck-$(SHELLCHECK_VERSION)/shellcheck \
+			&& rm $(SHELLCHECK_ARCHIVE) \
```
wilfred-s, line 377: wget is not installed on a mac by default we should not use it, target failed for me due to that. Other installs use curl:
```
curl -sSfL https://github.com/koalaman/shellcheck/releases/download/v0.8.0/shellcheck-v0.8.0.linux.x86_64.tar.xz| tar -x --strip-components=1 shellcheck-v0.8.0/shellcheck
```
That drops the `shellcheck` binary at the same level as from where the make target has been started and does not save the archive, it should maintain the x bit (no need for chmod), removes the need for mv and rmdir also.

Makefile
```diff
@@ -358,3 +358,37 @@ arch:
...
+	else \
+		echo "shellcheck has been installed"; \
+	fi
+
+# Check scripts
+.PHONY: check_scripts
+ALLSCRIPTS := $(shell find . -name '*.sh')
+check_scripts: install_shellcheck
+	@echo "running shellcheck"
+	@if ! $(shell which shellcheck) $(ALLSCRIPTS); then exit 1; fi
```
wilfred-s, line 393: This fails if the local machine does not have shellcheck installed and it was added via the make target and the $(BASE_DIR)scripts is not my the path.
Also see comment above about which and zsh

Makefile
```diff
@@ -358,3 +358,37 @@ arch:
...
+	@if [ -e $(SHELLCHECK_PATH) ]; then rm $(SHELLCHECK_PATH) ; fi
```
wilfred-s, line 394: Leave the binary and add it to `.gitignore`

wilfred-s (changes requested):
Makefile portability remarks only: the check must be able to be run without interaction from a clean machine.
Script changes all look good.

wilfred-s (approved):
LGTM
I'll fix the missing newline in the .gitignore when I commit this.

PR #448, opened 2022-08-11
[YUNIKORN-1276] REST API for specific application

manirajv06:
### What is this PR for?
Use specific Application REST API to fetch the application info directly.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1276

### Questions:


wilfred-s (changes requested):
We need the go.mod update to point to the latest core otherwise we do not have the API available.

https://github.com/apache/yunikorn-core/commit/f68f002d43f70cf1ab36a6ab1cb338099aea71c6

test/e2e/framework/helpers/yunikorn/rest_api_utils.go
```diff
@@ -128,19 +128,14 @@ func (c *RClient) GetApps(partition string, queueName string) ([]map[string]inte
 	return apps, err
 }
 
-func (c *RClient) GetAppInfo(partition string, queueName string, appID string) (map[string]interface{}, error) {
-	apps, err := c.GetApps(partition, queueName)
+func (c *RClient) GetAppInfo(partition string, queueName string, appID string) (*dao.ApplicationDAOInfo, error) {
```
wilfred-s, line 131: This signature change should trigger updates in all callers. Looking at the code there are at least 7 callers of this function.
We also need to become consistent in the way we call and process: either always return a `dao` type object for the thing we request or the generic `[]interface{}`.  I think `dao` is the only correct answer.
We already use `dao` for queue and partition but not for application...

wilfred-s (changes requested):

manirajv06:
@craigcondit Thanks.  Can you merge this? This is needed for 1.1.0 branch which might happen today.

craigcondit (approved):
+1 will merge shortly.

PR #450, opened 2022-08-18
[YUNIKORN-1280] Shim side changes for YUNIKORN-1275

manirajv06:
### What is this PR for?
Shim side changes to support arbitrary resources in namespace annotation.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1280

### Questions:

### DIFF
--- a/pkg/common/utils/utils.go
+++ b/pkg/common/utils/utils.go
@@ -19,6 +19,7 @@
 package utils
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


pkg/common/utils/utils.go
```diff
@@ -115,13 +116,22 @@ func GetNamespaceQuotaFromAnnotation(namespaceObj *v1.Namespace) *si.Resource {
...
+
+	// order of annotation preference
+	// 1. namespace.quota
+	// 2. namespace.max.* (Retaining for backwards compatibility. Need to be removed in next major release)
+	if namespaceQuota != "" {
+		var namespaceQuotaMap map[string]string
+		json.Unmarshal([]byte(namespaceQuota), &namespaceQuotaMap)
+		return common.GetResource(namespaceQuotaMap)
+	} else if cpuQuota != "" || memQuota != "" {
+		log.Logger().Warn("Using deprecated namespace.max.* annotations to set cpu and/or memory for namespace.",
```
wilfred-s, line 129: We need to mention the new annotations to use in the message.
We should also log the fact that both exist on the same namespace and we ignore the deprecated one.

test/e2e/queue_quota_mgmt/queue_quota_mgmt_test.go
```diff
@@ -39,8 +39,7 @@ var _ = Describe("", func() {
 	var sleepRespPod *v1.Pod
 	var maxCPU int64
 	var maxMem int64
-	var maxCPUAnnotation = "yunikorn.apache.org/namespace.max.cpu"
-	var maxMemAnnotation = "yunikorn.apache.org/namespace.max.memory"
+	var namespaceQuotaAnnotation = "yunikorn.apache.org/namespace.quota"
```
wilfred-s, line 42: we should not remove the old tests yet

wilfred-s (changes requested):
Can we split the tests for annotations into 3 tests:
- test annotations based on the new setup (new function)
- mixed new and deprecated (new function)
- deprecated annotation (existing test)

Makes checks and regressions easier to track down.

manirajv06:
> Can we split the tests for annotations into 3 tests:
> 
> * test annotations based on the new setup (new function)
> * mixed new and deprecated (new function)
> * deprecated annotation (existing test)
> 
> Makes checks and regressions easier to track down.

Taken care

wilfred-s (approved):
LGTM

PR #459, opened 2022-09-02
[YUNIKORN-1305] git sha label substitution

wilfred-s:
### What is this PR for?
The sha label on the image is not set correctly during local builds and
as part of the release process.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1305

The image should contain labels called "yunikorn-*-revision" with a sha reference and not empty or a command

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,9 +87,9 @@ DOCKER_ARCH := amd64
 endif
 
 # Image hashes
-CORE_SHA=$$(go list -m "github.com/apache/yunikorn-core" | cut -d "-" -f4)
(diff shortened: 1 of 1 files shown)


wilfred-s:
The build badge is showing no status, seems to happen sometimes for workflows.
Not sure if the link should be updated as lots of other repos have done as a workaround build badge Maybe we should change to the push to master workflow... might need to reopen YUNIKORN-1302

craigcondit (approved):
+1 LGTM.

PR #462, opened 2022-09-09
[YUNIKORN-1314] The current UserkeyLabel is showed in log when user name couldn't be found in pod's labels.

0yukali0:
### What is this PR for?
1.  The iteration is not required for searching userLabelKey.
2.  The logger adopting debug level show the current userLabelKey
3.  The userLabelKey would be replaced by the default userLabelKey when the userLabelKey returned from the configuration is empty.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1314

Pods' labels contains userLabelKey and the username rule is added to the partition.
```
metadata:
  labels:
    app: nginx
    applicationId: "username-rule-example02"
    yunikorn.apache.org/username: developer
  name: task1
spec:
  schedulerName: yunikorn
```

```
  partitions:
    -
      name: default
      placementrules:
        - name: user
          create: true
        - name: tag
          value: namespace
          create: true
      queues:
        - name: root
          submitacl: '*'
```

For example, an user, called developer, owns a pod. 
The queue doesn't exist or the queue label is not added in the pod.
The pod would be assigned to root.developer.
截圖 2022-09-09 下午11 24 30

### Questions:

### DIFF
--- a/pkg/common/utils/utils.go
+++ b/pkg/common/utils/utils.go
@@ -240,18 +240,21 @@ func MergeMaps(first, second map[string]string) map[string]string {
 // find user name from pod label
 func GetUserFromPod(pod *v1.Pod) string {
 	userLabelKey := conf.GetSchedulerConf().UserLabelKey
-	// User name to be defined in labels
-	for name, value := range pod.Labels {
-		if name == userLabelKey {
-			log.Logger().Info("Found user name from pod labels.",
-				zap.String("userLabel", userLabelKey), zap.String("user", value))
-			return value
-		}
+	// UserLabelKey should not be empty
+	if len(userLabelKey) == 0 {
+		userLabelKey = constants.DefaultUserLabel
+		log.Logger().Debug("userLabelKey is empty and replaced by default setting",
+			zap.String("userLabel", userLabelKey))
 	}
-	value := constants.DefaultUser
 
+	// User name to be defined in labels
+	if username, ok := pod.Labels[userLabelKey]; ok {
+		log.Logger().Info("Found user name from pod labels.",
+			zap.String("userLabel", userLabelKey), zap.String("user", username))
+		return username
+	}
 	log.Logger().Debug("Unable to retrieve user name from pod labels. Empty user label",
-		zap.String("userLabel", constants.DefaultUserLabel))
+		zap.String("userLabel", userLabelKey))
 
-	return value
+	return constants.DefaultUser
 }
--- a/pkg/common/utils/utils.go
+++ b/pkg/common/utils/utils.go
@@ -244,16 +244,16 @@ func GetUserFromPod(pod *v1.Pod) string {
 	if len(userLabelKey) == 0 {
 		userLabelKey = constants.DefaultUserLabel
 	}
-
-	username := constants.DefaultUser
 	// User name to be defined in labels
 	if username, ok := pod.Labels[userLabelKey]; ok {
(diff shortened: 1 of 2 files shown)


pkg/common/utils/utils.go
```diff
@@ -240,18 +240,21 @@ func MergeMaps(first, second map[string]string) map[string]string {
...
-		if name == userLabelKey {
-			log.Logger().Info("Found user name from pod labels.",
-				zap.String("userLabel", userLabelKey), zap.String("user", value))
-			return value
-		}
+	// UserLabelKey should not be empty
+	if len(userLabelKey) == 0 {
+		userLabelKey = constants.DefaultUserLabel
+		log.Logger().Debug("userLabelKey is empty and replaced by default setting",
+			zap.String("userLabel", userLabelKey))
```
wilfred-s, line 247: The default is documented and we should not need to print it every time we try to retrieve a user, simple like this: "userLabelKey not set using default value"

Need to also handle this as part of a unit test, codecov complains now

pkg/common/utils/utils.go
```diff
@@ -240,18 +240,21 @@ func MergeMaps(first, second map[string]string) map[string]string {
...
 	}
-	value := constants.DefaultUser
 
+	// User name to be defined in labels
+	if username, ok := pod.Labels[userLabelKey]; ok {
```
wilfred-s, line 251: Can the label be found and still return an empty username?
We need to handle that case and still set the default user

wilfred-s (changes requested):

pkg/common/utils/utils.go
0yukali0: Yes, the default user name will be return if yunikorn can't get the username from the pod.
I added the modification in the current commit.

pkg/common/utils/utils.go
0yukali0: Hi @wilfred-s, I had added the relevant thing to the latest commit.
The empty userLabelKey case was added to the unit test.
The log info was removed

0yukali0:
Hi @wilfred-s, Is the time-out event for waiting yunikorn pod related to GetUserFromPod in e2e test when the Kubernetes version is 1.25?
Should I re-run the pre-commit check?

wilfred-s:
re-running the tests see if it is consistently breaking or not.

pkg/common/utils/utils.go
wilfred-s: A label value can be empty. That would result in `ok` to be set to true but username to be "" (or empty string) in that case we should assign the default user also.
That is not covered by the change.

wilfred-s:
Test failed, not sure if it is the same one. Plugin with K8s 1.25 has been the problem, maybe @craigcondit knows more.
@0yukali0 can you have a look at my comment around the empty value that point is not resolved yet.

pkg/common/utils/utils.go
0yukali0: Ok, I am going to add relevant modifications for this case.

0yukali0:
Hi @wilfred-s 
I will check this case and update the commit tonight.

wilfred-s (approved):
LGTM

PR #465, opened 2022-09-26
[YUNIKORN-1322] [admission] add new config value handling and allow/deny label if defined

pbacsko:
### What is this PR for?
If an userinfo annotation is present on the pod, validate which user/group added it.
If the user/groups are not on a particular list, then the pod submission is denied.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1322

### Questions:


wilfred-s:
Still not sure why the admission controller is stored under the plugins. It does not belong there and should really be moved out of the plugin area. Can you please file a follow up jira to move the admission controller to its own (top level) directory.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -184,6 +239,13 @@ func (c *admissionController) mutate(req *admissionv1.AdmissionRequest) *admissi
 		return admissionResponseBuilder(uid, false, err.Error(), nil)
 	}
 
+	if annotation, ok := pod.Annotations[userInfoAnnotation]; ok && !c.bypassAuth {
+		if err := c.validateExternalUserInfo(req, annotation); err != nil {
+			log.Logger().Error("external authentication failed", zap.Error(err))
```
wilfred-s, line 244: The message is not correct. The external authentication did not fail. The `real` user is not allowed to set the annotations. Message should be more like: "User XX is not allowed to set user annotation".

pkg/plugin/admissioncontrollers/webhook/usergroup.go
```diff
@@ -0,0 +1,77 @@
...
+	"strings"
+
+	"go.uber.org/zap"
+	admissionv1 "k8s.io/api/admission/v1"
+
+	"github.com/apache/yunikorn-k8shim/pkg/log"
+	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
+)
+
+func (c *admissionController) validateExternalUserInfo(req *admissionv1.AdmissionRequest, userInfo string) error {
```
wilfred-s, line 33: Is the annotation only used in the error message?
We should not need to include the annotation details in the error message, they are not really relevant for the rejection of the pod.

pkg/plugin/admissioncontrollers/webhook/webhook.go
```diff
@@ -80,6 +92,44 @@ func main() {
...
+	if !ok {
+		systemUsers = defaultSystemUsers
+	}
+	externalUsers, ok := os.LookupEnv(admissionControllerExternalUsers)
+	if !ok {
+		externalUsers = ""
+	}
+	externalGroups, ok := os.LookupEnv(admissionControllerExternalGroups)
+	if !ok {
+		externalUsers = ""
```
wilfred-s, line 117: This should be `externalGroups` if I am correct.
Can we add some unit tests so we can catch things like this?

wilfred-s (changes requested):

pkg/plugin/admissioncontrollers/webhook/webhook.go
pbacsko: Code was extracted to a function & unit tests were added

pkg/plugin/admissioncontrollers/webhook/usergroup.go
pbacsko: Used in a Debug message & to validate that the json fragment itself is parseable.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
pbacsko: Corrected in both places

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -51,6 +51,7 @@ const (
 	configHotFreshResponse       = "ConfigHotRefresh is disabled. Please use the REST API to update the configuration, or enable configHotRefresh. "
 	admissionReviewAPIVersion    = "admission.k8s.io/v1"
 	admissionReviewKind          = "AdmissionReview"
+	userInfoAnnotation           = "yunikorn.apache.org/user.info"
```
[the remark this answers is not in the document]
pbacsko: Done

pkg/plugin/admissioncontrollers/webhook/usergroup.go
wilfred-s: Split the two actions: one for the json check and one for the allowed auth user check.
Makes it easier to understand what is going on

pbacsko: Done

wilfred-s: With the split into two calls we need to update the signature of the function and remove the userInfo parameter .

wilfred-s (changes requested):
two minor changes

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -184,6 +240,18 @@ func (c *admissionController) mutate(req *admissionv1.AdmissionRequest) *admissi
 		return admissionResponseBuilder(uid, false, err.Error(), nil)
 	}
 
+	if annotation, ok := pod.Annotations[userInfoAnnotation]; ok && !c.bypassAuth {
+		if err := c.validateExternalUserInfo(req, annotation); err != nil {
+			log.Logger().Error("user info validation failed", zap.Error(err))
```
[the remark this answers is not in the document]
pbacsko: Did a little refactor in this method. We just need a yes/no answer, so returning a "bool" from there instead of an "error".

pkg/plugin/admissioncontrollers/webhook/usergroup.go
pbacsko: Done

pbacsko:
Test failure is unrelated.

wilfred-s (approved):
LGTM

PR #466, opened 2022-09-27
[YUNIKORN-1196] Build against Kubernetes 1.23.12

craigcondit:
### What is this PR for?
Update build dependencies to Kubernetes 1.23.12. This provides better support for newer Kubernetes releases. However, this will drop support for running on Kubernetes releases earlier than 1.21.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1196

E2e tests are provided and test matrix has been updated.

### Questions:
* [x] - There is breaking changes for older versions.


VladZ97:
Hi, wanted to mention that building against 1.23.12 fixed [YUNIKORN-1253](https://issues.apache.org/jira/browse/YUNIKORN-1253)
Tested on openshift 4.10.15 (Which is using v1.23.5+9ce5071), with yunikorn Helm chard deployed as standalone scheduler w/o the admission controller.
Guessing the problem was that Generic ephemeral volumes were added in 1.23

craigcondit:
> Hi, wanted to mention that building against 1.23.12 fixed [YUNIKORN-1253](https://issues.apache.org/jira/browse/YUNIKORN-1253) Tested on openshift 4.10.15 (Which is using v1.23.5+9ce5071), with yunikorn Helm chart deployed as standalone scheduler w/o the admission controller. Guessing the problem was that Generic ephemeral volumes were added in 1.23

Unfortunately, there's still some issues here, especially when running in plugin mode. It seems the default scheduler logic in 1.23.x puts the pods into an internal unschedulable queue, where they remain for quite some time (sometimes over a minute). Until that's solved, this will be blocked.

wilfred-s:
> Unfortunately, there's still some issues here, especially when running in plugin mode. It seems the default scheduler logic in 1.23.x puts the pods into an internal unschedulable queue when they are not yet ready to be scheduled by YuniKorn, where they remain for quite some time (sometimes over a minute). Until that's solved, this will be blocked.

That should hopefully change when we have KEP-3521 fixed on the K8s side. With that change in place a pod can be marked not ready for scheduling and YuniKorn can release it.

craigcondit:
Code coverage decrease is expected, as there are now several no-op stub functions necessary in `framework_handle.go`. These are never called but are needed to satisfy the `FrameworkHandle` interface in K8s 1.23.

deployments/image/plugin/start-yunikorn-plugin.sh
```diff
@@ -17,13 +17,15 @@
 # limitations under the License.
 
 cd "${HOME}" || exit
+
+# Embed scheduler name into config file
+sed 's/schedulerName:.*/schedulerName: '"${SCHEDULER_NAME}"'/' < "${SCHEDULER_CONFIG}" > /tmp/scheduler-config.yaml || exit
```
wilfred-s, line 22: We need to move this out of this shell script and into pre-processing as this requires `sed` to be installed in the image which might not be the case for stripped down setups.

wilfred-s (changes requested):
I am not sure that changing the schedulerName will work as intended in all cases as the admission controller sets the schedullerName to yunikorn. This could have side effects which we need to asses.

deployments/image/plugin/start-yunikorn-plugin.sh
craigcondit: The point was to allow customizing the scheduler name, which as of 1.23 has to be in the scheduler-config file. However, as part of the configv2 redesign, we're removing the ability to change the scheduler name (which never really worked anyway). So I'm removing this code.

test/e2e/recovery_and_restart/recovery_and_restart_test.go
```diff
@@ -171,9 +171,15 @@ var _ = ginkgo.Describe("", func() {
...
+		err = kClient.WaitForJobPodsRunning(dev, job2.Name, parallelism, 60*time.Second)
+		Ω(err).NotTo(gomega.HaveOccurred())
+
+		ginkgo.By("Deleting sleep jobs")
+		err = kClient.DeleteJob(job1.Name, dev)
+		Ω(err).NotTo(gomega.HaveOccurred())
+		err = kClient.DeleteJob(job2.Name, dev)
 		Ω(err).NotTo(gomega.HaveOccurred())
 
 		ginkgo.By("Deleting sleep pods")
```
pbacsko, line 185: I think this part can be removed. This is exactly why I filed YUNIKORN-1404. No need to individually delete pods (optionally, we can wait for them to disappear).

pbacsko (approved):
+1 LGTM

You can remove the pod deletion part so YUNIKORN-1404 becomes unecessary.

0yukali0 (approved):
LGTM.

test/e2e/recovery_and_restart/recovery_and_restart_test.go
craigcondit: I think we can resolve this with a later PR. For now, let's just get this merged.

craigcondit:
Rebased on master since we have conflicts.

PR #469, opened 2022-10-13
[YUNIKORN-1334] Admission controller: look up user & group and add annotation if necessary

pbacsko:
### What is this PR for?
If the userinfo annotation is missing, it has be added to the incoming workload.
The existing mutation webhook will be extended with the necessary code which handles all types of workloads.
Attempts to change the userinfo annotation should be rejected.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [x] - Rebase on YUNIKORN-1338 once that PR is committed to master

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1334

### Questions:

### DIFF
--- a/pkg/plugin/admissioncontrollers/webhook/admission_controller.go
+++ b/pkg/plugin/admissioncontrollers/webhook/admission_controller.go
@@ -40,9 +40,9 @@ import (
 	"github.com/apache/yunikorn-k8shim/pkg/common/constants"
 	schedulerconf "github.com/apache/yunikorn-k8shim/pkg/conf"
 	"github.com/apache/yunikorn-k8shim/pkg/log"
-	"github.com/apache/yunikorn-k8shim/pkg/plugin/admissioncontrollers/webhook/annotation"
 	"github.com/apache/yunikorn-k8shim/pkg/plugin/admissioncontrollers/webhook/common"
 	"github.com/apache/yunikorn-k8shim/pkg/plugin/admissioncontrollers/webhook/conf"
+	"github.com/apache/yunikorn-k8shim/pkg/plugin/admissioncontrollers/webhook/metadata"
 )
 
 const (
@@ -66,7 +66,8 @@ var (
 
 type admissionController struct {
 	conf              *conf.AdmissionControllerConf
-	annotationHandler *annotation.UserGroupAnnotationHandler
+	annotationHandler *metadata.UserGroupAnnotationHandler
+	labelExtractor    metadata.LabelExtractor
 }
 
 type ValidateConfResponse struct {
@@ -77,7 +78,7 @@ type ValidateConfResponse struct {
 func initAdmissionController(conf *conf.AdmissionControllerConf) *admissionController {
 	hook := &admissionController{
 		conf:              conf,
-		annotationHandler: annotation.NewUserGroupAnnotationHandler(conf),
+		annotationHandler: metadata.NewUserGroupAnnotationHandler(conf),
 	}
 
 	log.Logger().Info("Initialized YuniKorn Admission Controller")
@@ -172,7 +173,7 @@ func (c *admissionController) processPod(req *admissionv1.AdmissionRequest, name
 	}
 
 	if !userInfoSet && !c.conf.GetBypassAuth() {
-		log.Logger().Info("setting user info annotation on pod")
+		log.Logger().Info("setting user info metadata on pod")
 		patchOp, err := c.annotationHandler.GetPatchForPod(pod.Annotations, userName, groups)
 		if err != nil {
 			return admissionResponseBuilder(uid, false, err.Error(), nil)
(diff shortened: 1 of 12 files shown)


wilfred-s:
committed #467, I think we can proceed with this one now

wilfred-s (approved):
LGTM
@craig can you please review this one too?

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -199,11 +218,10 @@ func (c *admissionController) processPod(req *admissionv1.AdmissionRequest) *adm
 	return admissionResponseBuilder(uid, true, "", patchBytes)
 }
 
-func (c *admissionController) processWorkload(req *admissionv1.AdmissionRequest) *admissionv1.AdmissionResponse {
+func (c *admissionController) processWorkload(req *admissionv1.AdmissionRequest, namespace string) *admissionv1.AdmissionResponse {
```
craigcondit, line 221: The `namespace` parameter is unused here.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -212,39 +230,99 @@ func (c *admissionController) processWorkload(req *admissionv1.AdmissionRequest)
...
+	}
+
+	if labelAppValue, ok := newPod.Labels[constants.LabelApp]; ok {
+		if labelAppValue == yunikornPod {
+			log.Logger().Info("pod update - ignore yunikorn pod")
+			return admissionResponseBuilder(uid, true, "", nil)
+		}
+	}
+
+	if !c.shouldProcessNamespace(namespace) {
```
craigcondit, line 279: I think this should be: 
  `if !c.shouldProcessNamespace(namespace) || !c.shouldLabelNamespace(namespace) {`
Only labeled pods get scheduled via YuniKorn queueing, so we should exclude based on both process and label status.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
craigcondit: Also, I think we need to exclude workloads the same way we do pods, by checking the namespace against (`process/bypass)Namespaces` and `(label/noLabel)Namespaces`. Otherwise we're affecting workloads that will not be scheduled by YuniKorn.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
```diff
@@ -212,39 +230,99 @@ func (c *admissionController) processWorkload(req *admissionv1.AdmissionRequest)
...
+		log.Logger().Info("pod update - bypassing namespace", zap.String("namespace", namespace))
+		return admissionResponseBuilder(uid, true, "", nil)
+	}
+
+	pod, err := c.clientSet.CoreV1().Pods(newPod.Namespace).Get(ctx.Background(), newPod.Name, metav1.GetOptions{})
```
craigcondit, line 284: This is ugly. We're forcing a round-trip to the API server during a webhook. It seems as though you're only checking this to validate that the user info hasn't changed? If so, I'm pretty sure `AdmissionRequest` has an `OldObject` member that should be populated for `Update` requests.

craigcondit (changes requested):
Mostly looks good, but there's a few things that need addressing.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
craigcondit: To expand on this... especially in plugin mode, yunikorn will get basically all traffic, but use label/noLabel to determine which pods to send to the embedded default scheduler.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
pbacsko: Yeah, that's what I was looking for, but didn't find it. I thought it's the only option.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
pbacsko: Could you double check this?

If something is in `noLabelNamespaces`, we still add the `scheduler` label, we skip `disableStateAware`, `queue` and `applicationId`. But to me it looks like Yunikorn will process it.

craigcondit: Yunikorn processes it, however because the applicationId is missing the scheduler plug-in goes into no-op mode and passes it through to the underlying default scheduler. 

In standard mode, Yunikorn just bails out and won't schedule the pod at all. ☹️ 

Technically even this check isn't quite good enough, because we don't account for user-supplied applicationId being present. We should probably check for existing appId or spark appId and opt back in to the security checks. 

In that case our logic becomes:

shouldProcess && (appId/sparkAppId present || shouldLabel)

That's ugly, but I think correct. Probably build a helper function for this and we can call it from wherever needed.

craigcondit (changes requested):
Updated logic for old pod looks better. We do still need a more comprehensive check for what objects (pods or otherwise) should be processed here:

1. if shouldProcessNamespace() is false, do nothing
2. if shouldLabelNamespace() is true, do processing
3. If shouldLabelNamespace() is false, check for `spark-app-selector` or `applicationId` on the pod (or pod template). If either are present, do processing.
4. Otherwise, do nothing.

pkg/plugin/admissioncontrollers/webhook/admission_controller.go
pbacsko: Done.

I had to add extra code to extract the labels from deployments, replicasets, etc. This required some refactoring and new tests.

craigcondit (approved):
Looks good to me. @wilfred-s can you review the last few changes as well?

wilfred-s:
looking good now, merging

PR #470, opened 2022-10-28
[YUNIKORN-1375] [shim] pass group info from the shim to the core

pbacsko:
### What is this PR for?
The list of groups in `si.AddApplicationRequest` is not passed from the shim to core.
We need to make sure that we store the group inside the `Application` object and pass them when needed.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1375

### Questions:


craigcondit (approved):
+1 LGTM.

PR #471, opened 2022-11-02
[YUNIKORN-1378] Add predicate with equivalence cache to improve the scheduler performance

zhuqi-lucas:
### What is this PR for?
Now we call Filter every time when we predicate, we can cache some operations for predicate so that we can improve the scheduler performance.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1378

### Questions:


zhuqi-lucas:
> We need unit tests for the new code. The change also causes existing unit tests to fail with a panic.
> 
> The change introduces a huge memory leak: there is no clean up of the cache which means it will grow unbound. I can understand that the cache might speed things up but we need to have cache management. Without proper cache management we should not do this.

Thank you @wilfred-s for review, great suggestions,  i will change Unit test, and try to add cache management!

craigcondit:
The latest commit still leaks resources badly, as the cache will only be cleaned up when nodes are removed. This is very infrequent, and in some clusters may never be called. The result is that the cache will grow without bound, containing an entry for every node/pod combination. In a cluster with 1000 nodes, you'll get 1 million entries with only 1000 pods, and a billion entries with 1 million pods (cumulative over time). A simple nested map structure isn't going to be sufficient.

I also have doubts as to how safe caching the predicate results is. The key for the cache seems to only be {node name, pod uid}. The state of either the pod or node is not considered. So for example, nodeunschedulable may return different results at different times based on the node status, but the first time we see a result for a given node that value will be cached indefinitely for the pod. This is clearly incorrect behavior, and may result in YuniKorn picking a node which is unsuitable for a pod, or never picking a node even if it's the only one that pod is able to schedule to. This could have disastrous results especially when auto-scaling. The other two predicates that are cached here have similar issues - the results that are cached are subject to change.

Finally, I really question how much performance can be gained here. The three predicates that are cached are all relatively inexpensive, as they are doing simple in-memory operations such as map lookups (not unlike what the new cache itself needs to do). I'd like to see some solid benchmarking proving that this is indeed solving a real scalability bottleneck.

craigcondit (changes requested):
Given my concerns above, I'm -1 on this until there's some architectural changes and  benchmarking to ensure that this change is correct, performant, and necessary.

zhuqi-lucas:
Thank you, @craigcondit for your review, you are right, we also need to add more work to make this without leaking resources.
And the performance may not be gained very well, and i saw volcano add cache for this, and i don't know the performance gain for vovano, but they add a flag to enable cache, they set the flag disable cache default.
https://github.com/volcano-sh/volcano/pull/1165

wilfred-s:
We do things completely different than Volcano. Even if they see an improvement it does not mean we will see one too. In other words: even if they say they see an impact that will not change the requirement from our side.

Closing this PR and moving the discussion to the jira to work through some of the real basic points made earlier.

PR #472, opened 2022-11-03
[YUNIKORN-1323] Add e2e tests for group handling

pbacsko:
### What is this PR for?
End-to-end tests which validate that the user-info annotation is properly added by the admission controller.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1323

### Questions:

### DIFF
--- a/test/e2e/admission_controller/admission_controller_suite_test.go
+++ b/test/e2e/admission_controller/admission_controller_suite_test.go
@@ -24,6 +24,7 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -42,6 +43,7 @@ var ns = "admission-controller-test"
 var blackNs = "kube-system"
 var restClient yunikorn.RClient
 var oldConfigMap *v1.ConfigMap
+var replicas = int32(1)
 var testPod = v1.Pod{
 	ObjectMeta: metav1.ObjectMeta{
 		Name:   "sleepjob",
@@ -56,6 +58,20 @@ var testPod = v1.Pod{
 			},
 		},
 	},
+}
+
+var testDeployment = appsv1.Deployment{
+	Spec: appsv1.DeploymentSpec{
+		Replicas: &replicas,
+		Selector: &metav1.LabelSelector{
+			MatchLabels: testPod.Labels,
+		},
+		Template: v1.PodTemplateSpec{
+			ObjectMeta: testPod.ObjectMeta,
+			Spec:       testPod.Spec,
+		},
+	},
+	ObjectMeta: testPod.ObjectMeta,
 }
 
 func TestAdmissionController(t *testing.T) {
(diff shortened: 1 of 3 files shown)


test/e2e/admission_controller/admission_controller_suite_test.go
```diff
@@ -58,6 +60,39 @@ var testPod = v1.Pod{
...
+			Spec: v1.PodSpec{
+				Containers: []v1.Container{
+					{
+						Name:    "sleepjob",
+						Image:   "alpine:latest",
+						Command: []string{"sleep", "30"},
+					},
+				},
+			},
+		},
```
wilfred-s, line 86: This template will come back in a number of places we might want to factor this out so we can re-use.

wilfred-s (commented):
Is it possible to check the "do not trust controllers" case inside the e2e tests?
Would be nice to have that one added if possible.
Checking user allowed users is I think not possible with the test setup we use which is OK to skip.

test/e2e/admission_controller/admission_controller_suite_test.go
pbacsko: Done

pbacsko:
> Is it possible to check the "do not trust controllers" case inside the e2e tests? Would be nice to have that one added if possible. Checking user allowed users is I think not possible with the test setup we use which is OK to skip.

Actually both can be tested, because the configmap can be updated and changes are applied on-the-fly.

wilfred-s:
Kicked off the two failed tests again. K8s 1.25 failed for both default and plugin but at different points.
Does not look like a code issue.

wilfred-s (approved):
LGTM, would like to see the K8s 1.25 tests pass before we commit

pbacsko:
> LGTM, would like to see the K8s 1.25 tests pass before we commit

We have a green build, but I think the tests are still flaky. We can give it a go and if they're very unstable then we'll figure out something.

test/e2e/admission_controller/admission_controller_test.go
```diff
@@ -107,6 +110,139 @@ var _ = ginkgo.Describe("AdmissionController", func() {
...
+		gomega.Ω(deployment.Status.ReadyReplicas).To(gomega.Equal(int32(0)))
+
+		// restore setting
+		ginkgo.By("Restore trustController setting")
+		configMap, err = kubeClient.GetConfigMap(constants.ConfigMapName, configmanager.YuniKornTestConfig.YkNamespace)
+		gomega.Ω(err).ShouldNot(gomega.HaveOccurred())
+		configMap.Data[amConf.AMAccessControlTrustControllers] = "true"
+		_, err = kubeClient.UpdateConfigMap(configMap, configmanager.YuniKornTestConfig.YkNamespace)
+		gomega.Ω(err).ShouldNot(gomega.HaveOccurred())
+
```
wilfred-s, line 194: Should have a sleep here like in the other places we do the update

wilfred-s (changes requested):

test/e2e/admission_controller/admission_controller_test.go
craigcondit: That will definitely introduce a delay, but doesn't verify that YuniKorn has applied the config. We may want to follow-up later with a more comprehensive method to check that config has been applied. This is fine for now though.

wilfred-s (approved):
LGTM
the eventhandler waits for a notify that of the change, ew then give the system 1 second to process the change that should be more than enough

PR #475, opened 2022-11-04
[YUNIKORN-1320] [shim] Handle new userInfo annotation

pbacsko:
### What is this PR for?
Add code which can extract the user-group info from the new annotation.
For backwards compatbility, the original user label is still checked if the annotation is missing.  

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1320

### Questions:

### DIFF
--- a/pkg/common/utils/utils.go
+++ b/pkg/common/utils/utils.go
@@ -35,8 +35,11 @@ import (
 	"github.com/apache/yunikorn-k8shim/pkg/common/constants"
 	"github.com/apache/yunikorn-k8shim/pkg/conf"
 	"github.com/apache/yunikorn-k8shim/pkg/log"
+	siCommon "github.com/apache/yunikorn-scheduler-interface/lib/go/common"
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
 )
+
+const userInfoKey = siCommon.DomainYuniKorn + "user.info"
 
 func Convert2Pod(obj interface{}) (*v1.Pod, error) {
 	pod, ok := obj.(*v1.Pod)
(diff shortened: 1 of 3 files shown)


pkg/common/utils/utils.go
```diff
@@ -238,7 +241,27 @@ func MergeMaps(first, second map[string]string) map[string]string {
...
+	if pod.Annotations[userInfoKey] != "" {
+		userInfoJSON := pod.Annotations[userInfoKey]
+		var userGroup si.UserGroupInformation
+		err := json.Unmarshal([]byte(userInfoJSON), &userGroup)
+		if err != nil {
+			log.Logger().Error("unable to process user info annotation", zap.Error(err))
+			return constants.DefaultUser, nil
+		}
+		user := userGroup.User
+		groups := userGroup.Groups
```
kannon92, line 254: New to the project so this may be a useless suggestion. Do you care if groups are empty? Should there be logic for a default group? Or at least a warning if groups are empty?

pbacsko: Makes sense, I added a default group + logging

craigcondit: I disagree that we should have default groups, as this creates "magic" values that behave in somewhat unpredictable ways. An empty group list is allowed explicitly by the design, so we should not add additional groups here.

wilfred-s (approved):
LGTM

PR #480, opened 2022-11-14
[YUNIKORN-1369] Shim: Read configuration from configmaps and deprecate legacy configs

craigcondit:
### What is this PR for?
Read configuration from new ConfigMap entries in addition to legacy values, and deprecate old configuration sources.

[WIP] Based on and depends on from YUNIKORN-1365.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1369

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -38,6 +38,7 @@ import (
 	"github.com/apache/yunikorn-k8shim/pkg/common/constants"
 	"github.com/apache/yunikorn-k8shim/pkg/common/events"
 	"github.com/apache/yunikorn-k8shim/pkg/common/utils"
+	schedulerconf "github.com/apache/yunikorn-k8shim/pkg/conf"
 	"github.com/apache/yunikorn-k8shim/pkg/dispatcher"
 	"github.com/apache/yunikorn-k8shim/pkg/log"
 	"github.com/apache/yunikorn-k8shim/pkg/plugin/predicates"
@@ -55,6 +56,7 @@ type Context struct {
 	predManager    predicates.PredicateManager    // K8s predicates
 	pluginMode     bool                           // true if we are configured as a scheduler plugin
 	namespace      string                         // yunikorn namespace
+	configMaps     []*v1.ConfigMap                // cached yunikorn configmaps
 	lock           *sync.RWMutex                  // lock
 }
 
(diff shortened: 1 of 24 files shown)


pkg/client/kubeclient.go
```diff
@@ -73,6 +68,29 @@ func newSchedulerKubeClient(kc string) SchedulerKubeClient {
 	}
 }
 
+func createRestConfig(kc string) *rest.Config {
+	// attempt to use in-cluster config
+	config, err := rest.InClusterConfig()
+	if err != nil && err != rest.ErrNotInCluster {
+		// this is called during early initialization, so abort
+		panic(err)
```
pbacsko, line 76: Nit: how about `log.Logger.Panic()` or `log.Logger.Fatal()`?

pkg/common/constants/constants.go
```diff
@@ -56,7 +57,7 @@ const DaemonSetType = "DaemonSet"
 const AppManagerHandlerName = "yunikorn-app"
 
 // Gang scheduling
-const PlaceholderContainerImage = "k8s.gcr.io/pause"
+const PlaceholderContainerImage = "registry.k8s.io/pause:3.7"
```
pbacsko, line 60: This image is hard-coded at various places, like in the `Dockerfile` or `values.yaml`, do those need to be changed?

pkg/client/kubeclient.go
craigcondit: Circular dependencies. This is called before logging is safe to call. Panic will log the error and abort, so we still get it. It's also pretty unlikely to fail in real life.

pkg/common/constants/constants.go
craigcondit: The other places will be going away once this PR is complete, as we are transitioning to having defaults be in code only.

pkg/cache/context.go
```diff
@@ -290,33 +295,79 @@ func (ctx *Context) filterConfigMaps(obj interface{}) bool {
...
-	ctx.triggerReloadConfig(nil)
-}
+	var configmap *v1.ConfigMap = nil
+	switch t := obj.(type) {
+	case *v1.ConfigMap:
+		configmap = t
+	case cache.DeletedFinalStateUnknown:
+		configmap = utils.Convert2ConfigMap(obj)
+	default:
+		log.Logger().Debug("unable to convert to configmap")
```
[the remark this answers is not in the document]
craigcondit: Fair point. I'll update in the next version.

pbacsko (approved):
+1 LGTM (pending small change related to logging)

craigcondit:
Merged to master, updated Debug() to Warn() as requested.

pkg/client/kubeclient.go
craigcondit: Hmm, noticed after commit that we can in fact do this. Will update in the next PR in this series.

PR #482, opened 2022-11-16
[YUNIKORN-1370] Read admission controller config from configmaps

craigcondit:
### What is this PR for?
Updates the admission controller to read its configuration from configmaps instead of environment variables. Additionally, config reloading is supported.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1370

New unit tests added, updated e2e test to allow empty configmap to validate.

### Questions:

### DIFF
--- a/pkg/plugin/admissioncontrollers/webhook/conf/am_conf.go
+++ b/pkg/plugin/admissioncontrollers/webhook/conf/am_conf.go
@@ -1,0 +1,477 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package conf
+
+import (
+	"fmt"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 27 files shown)


pkg/plugin/admissioncontrollers/webhook/conf/am_conf.go
```diff
@@ -0,0 +1,477 @@
...
+
+func parseConfigString(config map[string]string, key string, defaultValue string, env string) string {
+	result := defaultValue
+	var envValue, confValue string
+	var envOk, confOk bool
+
+	if env != "" {
+		envValue, envOk = os.LookupEnv(env)
+		if envOk {
+			result = envValue
```
pbacsko, line 448: Just an idea - we might want to display a message with `Warn()` that configmaps are preferred over env variables and suggest migration. If the plan is to completely move away from it, then we can be loud about it.

craigcondit: This is temporary code. Once this merges, Helm will be updated to only set the configmap entries and the env card will go away completely. That wasn't done here as it would break intermediate builds.

test/e2e/framework/helpers/yunikorn/yk_utils.go
```diff
@@ -21,6 +21,13 @@ package yunikorn
 import (
 	"fmt"
 
+	"github.com/apache/yunikorn-core/pkg/common/configs"
+	"github.com/apache/yunikorn-k8shim/pkg/common/constants"
+
+	//	"github.com/apache/yunikorn-k8shim/pkg/common/constants"
```
pbacsko, line 27: Nit: delete comment

craigcondit: Good catch. Will remove on commit if that's the only required change.

pbacsko (changes requested):
I found two minor things.

pkg/plugin/admissioncontrollers/webhook/conf/am_conf.go
pbacsko: I see, then I'm ok with this.

pbacsko (approved):
Ok, fix the comment on commit.
+1 LGTM

craigcondit:
> I found two minor things.

Can you give a conditional +1 since the comment removal is trivial? That will save a full re-review cycle.

craigcondit:
Merged to master with extra comment removed.

PR #486, opened 2022-11-21
[YUNIKORN-1423] Update README.md in placement examples that how to update configuration

0yukali0:
### What is this PR for?
[YUNIKORN-1371](https://github.com/apache/yunikorn-release/pull/116) updated the deployment, so YUNIKORN-1423 update the README.md to describe how to deploy the configuration.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [x] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1423?

### Questions:


deployments/examples/placements/README.md
```diff
@@ -20,13 +20,22 @@
...
 The sleep pod is described in the example file and the partition is provided in the config yaml file.
-Before deploying the pods, the configuration field in yunikorn-release/helm/yunikorn/value.yaml should be replaced by the configuration field in the config.yaml.
 
 * [Provided Rule](./provided)
 * [User Name Rule](./username)
 * [Fixed Rule](./fixed)
 * [Tag Rule](./tag)
 
+## How to adopt the queues in `config.yaml`
+Before deploying the pods, threre are some modifications required in [release/value.yaml](yunikorn-release/helm/yunikorn/value.yaml).
```
craigcondit, line 30: Typo here - this should be `values.yaml`

deployments/examples/placements/README.md
```diff
@@ -20,13 +20,22 @@
...
+1. Replace the configuration field by the configuration in the `config.yaml`.
+2. Add `queue.yaml: ""` in `yunikornDefaults` field.
+```
+yunikornDefaults: {
+  queue.yaml: ""
```
craigcondit, line 35: This should be `queues.yaml`, not `queue.yaml`. Also, why are we setting this to an empty string and not simply to the new configuration?

craigcondit (changes requested):

deployments/examples/placements/README.md
0yukali0: I trace the code in the https://github.com/apache/yunikorn-release/blob/master/helm-charts/yunikorn/templates/yunikorn-defaults.yaml#L60-L68。
It seems to get the key but it will not get the value.
So I use empty string here.
Is there a suggested value of the queues.yaml?

deployments/examples/placements/README.md
0yukali0: Oops,  I typed wrong words here.
I am going to update this words including `values.yaml` and `queues.yaml`.

deployments/examples/placements/README.md
```diff
@@ -20,13 +20,22 @@
...
 Every placement example includes a example yaml file and a config yaml file.
 The sleep pod is described in the example file and the partition is provided in the config yaml file.
-Before deploying the pods, the configuration field in yunikorn-release/helm/yunikorn/value.yaml should be replaced by the configuration field in the config.yaml.
 
 * [Provided Rule](./provided)
 * [User Name Rule](./username)
 * [Fixed Rule](./fixed)
 * [Tag Rule](./tag)
 
+## How to adopt the queues in `config.yaml`
```
craigcondit, line 29: We shouldn't just wipe out the default here. Instead, show the full configuration using the placement rules in `yunikornDefaults`, but also point out that the configuration can also be uploaded to the `yunikorn-configs` ConfigMap where it will take precedence. Editing `values.yaml` should be discouraged.

craigcondit (changes requested):
See comments.

deployments/examples/placements/README.md
craigcondit: This is the effective default, edit as necessary for your example here:
```
yunikornDefaults:
  queues.yaml: |
    partitions:
      - name: default
        placementrules:
          - name: tag
            value: namespace
            create: true
        queues:
          - name: root
            submitacl: '*'    
```

0yukali0: Ok, will update this week.

wilfred-s (approved):
LGTM

wilfred-s:
With the work that has just started to allow queue type changes the processing of placement rules will probably change.
Any follow up that would be needed should be hooked into that major work.

PR #457, opened 2022-09-01
[YUNIKORN-1281] Add e2e tests

manirajv06:
### What is this PR for?
Added e2e tests to cover arbitrary resources support.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1281

### Questions:
* [ ] - The licenses files need update.
* [ ] - There is breaking changes for older versions.
* [ ] - It needs documentation.

PR #491, opened 2022-11-28
[YUNIKORN-1430] change go version check in Makefile

wusamzong:
### What is this PR for?
Point `MOD_VERSION` to `.go_version`, which will cause the make to fail if go is too old
I am going to update to all repos
[yunikorn-core #457](https://github.com/apache/yunikorn-core/pull/457)
[yunikorn-k8shim #491](https://github.com/apache/yunikorn-k8shim/pull/491)
[yunikorn-scheduler-interface #78](https://github.com/apache/yunikorn-scheduler-interface/pull/78)

### What type of PR is it?
* [x] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-1430

`make run`

Here is the result I use go1.16
image

Here is the result I use go1.18
image

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -13,12 +13,12 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 2
-      - name: Check license
-        run: make license-check
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
           go-version-file: .go_version
+      - name: Check license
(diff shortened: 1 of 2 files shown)


craigcondit (changes requested):
Looks like we need 'Setup Go' to execute before 'License Check'

wilfred-s (approved):
LGTM

PR #493, opened 2022-11-30
[YUNIKORN-1435] Expose RM diagnostics in full state dump

craigcondit:
### What is this PR for?
Register the RM as a `StateDumpPlugin` so that it can participate in state dumps. During a full state dump, the contents of the scheduler cache and internal data structures will be output.

NOTE: This PR will be in draft status until https://github.com/apache/yunikorn-core/pull/459 is merged, as the Core / SI must be updated first.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1435

Unit tests have been added to verify basic functionality.

### Questions:


craigcondit:
Note for reviewers: The massive go.mod / go.sum changes were needed as something broke the build upstream (I suspect the prometheus update).

manirajv06 (changes requested):
changes looks good. overall test coverage needs to be improved lot specifically for each object dao inside dump.

pkg/cache/external/scheduler_cache_dao.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+type ContainerDao struct {
+	Name      string                  `json:"name"`
+	Resources v1.ResourceRequirements `json:"resources"`
+}
+
+type PodSchedulingInfoDao struct {
+	Namespace       string    `json:"-"`
+	Name            string    `json:"-"`
```
manirajv06, line 87: Is "-" intentional?

manirajv06 (changes requested):

pkg/cache/external/scheduler_cache_dao.go
craigcondit: Yes, we don't want these to emit via JS.

craigcondit:
> changes looks good. overall test coverage needs to be improved lot specifically for each object dao inside dump.

The objects themselves are simply copies of K8s objects. The new test cases verify that they are created correctly.

manirajv06:
> The objects themselves are simply copies of K8s objects.

Not all, I guess. For ex, assignedPods, assumedPods, pendingAllocations etc. Even I am not going by code coverage tool results. Yes, objects creation are asserted. I meant the assertion to verify the object properties/details. May be, we can file a newbie jira to cover these properties in detail.

manirajv06 (approved):

PR #501, opened 2022-12-10
[YUNIKORN-1463] Add PriorityClass informer

craigcondit:
### What is this PR for?

Updates the shim and admission controller to listen for changes to Kubernetes PriorityClass objects. These are then used to populate the yunikorn.apache.org/allow-preemption annotation on new pods.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1463

Unit tests have been added. E2e tests have also been added to ensure the admission controller properly sets the value of the allow-preemption annotation.

### Questions:


pkg/admission/priority_class_cache.go
```diff
@@ -0,0 +1,117 @@
...
+
+func (h *priorityClassUpdateHandler) OnDelete(obj interface{}) {
+	var pc *schedulingv1.PriorityClass
+	switch t := obj.(type) {
+	case *schedulingv1.PriorityClass:
+		pc = t
+	case cache.DeletedFinalStateUnknown:
+		pc = utils.Convert2PriorityClass(obj)
+	default:
+		log.Logger().Warn("unable to convert to configmap")
```
pbacsko, line 107: "configmap"

pkg/admission/priority_class_cache.go
```diff
@@ -0,0 +1,117 @@
...
+
+func (pcc *PriorityClassCache) RegisterHandlers(priorityClasses informersv1.PriorityClassInformer) {
+	priorityClasses.Informer().AddEventHandler(&priorityClassUpdateHandler{cache: pcc})
+}
+
+func (pcc *PriorityClassCache) IsPreemptSelfAllowed(priorityClassName string) bool {
+	priorityClass := pcc.getPriorityClass(priorityClassName)
+
+	// default to true
+	if priorityClass == nil {
```
pbacsko, line 52: Can this happen? I'd log something here.

pkg/common/utils/utils.go
```diff
@@ -428,3 +436,13 @@ func UpdatePodLabelForAdmissionController(pod *v1.Pod, namespace string) map[str
 
 	return result
 }
+
+func UpdatePodAnnotationForAdmissionController(pod *v1.Pod, key string, value string) map[string]string {
```
pbacsko, line 440: If this code is only used by the A.C., then "pkg/admission" is a better place for this function.

pkg/admission/priority_class_cache.go
craigcondit: Yes, it can and probably won't be that unusual. In fact, we allow passing an empty string into here.
This can also happen if the priority class referenced is not found.
I'd prefer not to log as I think this may actually be somewhat common.

pkg/admission/admission_controller.go
```diff
@@ -197,6 +199,7 @@ func (c *AdmissionController) processPod(req *admissionv1.AdmissionRequest, name
 
 	if c.shouldLabelNamespace(namespace) {
 		patch = updateLabels(namespace, &pod, patch)
+		patch = c.updatePreemptionInfo(&pod, patch)
```
pbacsko, line 202: I think some refactor or clean-up is becoming unavoidable in this file. It's getting larger and these functions are inconsistent (there are more). Not in scope, just saying, maybe it's worth a follow-up JIRA.

craigcondit: Agreed.

pkg/admission/priority_class_cache.go
craigcondit: Good catch, will update.

pkg/cache/context.go
```diff
@@ -693,6 +745,19 @@ func (ctx *Context) AddApplication(request *interfaces.AddApplicationRequest) in
 	return app
 }
 
+func (ctx *Context) IsPreemptSelfAllowed(priorityClassName string) bool {
+	priorityClass := ctx.schedulerCache.GetPriorityClass(priorityClassName)
+	if priorityClass == nil {
+		return true
+	}
+	if value, ok := priorityClass.Annotations[constants.AnnotationAllowPreemption]; ok {
+		if value == "false" {
```
pbacsko, line 754: `constants.False` ?

pkg/common/utils/utils.go
craigcondit: Agreed.

pkg/cache/context.go
craigcondit: Good catch.

pbacsko (changes requested):
I found some small things, overall looks good.

pkg/cache/context.go
craigcondit: Updated in latest PR.

pkg/common/utils/utils.go
craigcondit: Updated in latest PR.

pkg/admission/priority_class_cache.go
craigcondit: Updated in latest PR.

PR #518, opened 2023-01-30
[YUNIKORN-1549] Introduce a config to specify what attribute to use as instance type

zhuqi-lucas:
### What is this PR for?
We need instance type for https://issues.apache.org/jira/browse/YUNIKORN-1385, and we need to make the node label instance type configuration.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1549

### Questions:

### DIFF
--- a/pkg/common/constants/constants.go
+++ b/pkg/common/constants/constants.go
@@ -27,6 +27,7 @@ const DefaultNodeAttributeHostNameKey = "si.io/hostname"
 const DefaultNodeAttributeRackNameKey = "si.io/rackname"
 const DefaultNodeAttributeNodeLabelsKey = "si.io/nodelabels"
 const DefaultRackName = "/rack-default"
+const DefaultNodeInstanceTypeNodeLabelKey = "node.kubernetes.io/instance-type"
 
 // Application
 const LabelApp = "app"
--- a/pkg/conf/schedulerconf.go
+++ b/pkg/conf/schedulerconf.go
@@ -53,16 +53,17 @@ const (
 	PrefixKubernetes = "kubernetes."
 
 	// service
-	CMSvcClusterID              = PrefixService + "clusterId"
-	CMSvcPolicyGroup            = PrefixService + "policyGroup"
-	CMSvcSchedulingInterval     = PrefixService + "schedulingInterval"
-	CMSvcVolumeBindTimeout      = PrefixService + "volumeBindTimeout"
-	CMSvcEventChannelCapacity   = PrefixService + "eventChannelCapacity"
-	CMSvcDispatchTimeout        = PrefixService + "dispatchTimeout"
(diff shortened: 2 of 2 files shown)


craigcondit:
We already have all the node labels passed in for each node. Rather than specifying this configuration on every node creation, wouldn't it make more sense to just use the config on the core side? Then the correct key could be extracted there.

zhuqi-lucas:
Thank you @craigcondit , good suggestion, let me refactor to core and parse it in the core side.

zhuqi-lucas:
Hi @craigcondit , If i make sense right, we need to make it configuration in shim, the core will read the change if we add the config in shim side?

wilfred-s:
-1 on the proposal and implementation

The k8shim should pass all labels that are defined on a node, no filtering, as tags to the core.
That is what is implemented in YUNIKORN-1389. I do think that the way that jira is implemented is broken to begin with.

zhuqi-lucas:
Thank you @wilfred-s for review, if you mean we shouldn't pass all labels to core, we just need to pass which we needed to core, for example, here we pass the instanceType attribute.

zhuqi-lucas:
And the corresponding change in the core side is:
https://github.com/apache/yunikorn-core/pull/498
cc @wilfred-s @craigcondit

craigcondit:
There is NO change required on the shim side, as all configuration is passed along to the core as of v1.2.0, including unknown / extra variables. Define a constant in the scheduler interface for the value and default, and reference that value in the core.

-1 on the PR.

wilfred-s:
Let me clarify why I am a -1 on the change, this includes the change in the core:
The design of YuniKorn allows multiple shims to be registered to one core. Even if we do not use it currently we should not go and willingly break that. If you make this a configuration value that is processed on the core side every shim must use the same setting. That is wrong.
If you want to allow multiple possible labels to be used you need to hide that in the shim. Setup a well known attribute name to pass on to the core. Only use that well known attribute on the core side. The key for that value would be defined in the `scheduler interface` as a constant. That would at least not break the multiple shims with different settings. How we would fit that into the config is a different question.

Before we even go there you need to fix YUNIKORN-1550, the way the attributes are used to pass a json string is broken.

zhuqi-lucas:
Thank you @wilfred-s @craigcondit , it makes sense to me now, let me do the YUNIKORN-1550 fix first, so that we don't need to pass the json string!

zhuqi-lucas:
Close it now.
We don't need the change in shim side, update the code change in core side.

PR #519, opened 2023-02-01
[YUNIKORN-1550] pass k8s node labels as attributes not json

zhuqi-lucas:
### What is this PR for?
The k8shim in [YUNIKORN-1389](https://issues.apache.org/jira/browse/YUNIKORN-1389) was extended to pass on the K8s node labels to the core via the attributes. The K8s node labels are marshalled into a single json string. The string is added using a new key.

There are multiple issues with that. The overhead of un-marshalling it each time a label needs to be checked for existence is large. It also needs a two step check to find a label. First find the label in the attributes and then un-marshal the string and do another check for the real key & value needed.

We should add the labels from the K8s node as entries in the attributes map. Each label becomes and entry not json marshalling.

### What type of PR is it?
* [x] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1550

### Questions:

### DIFF
--- a/pkg/cache/nodes.go
+++ b/pkg/cache/nodes.go
@@ -19,7 +19,6 @@
 package cache
 
 import (
-	"encoding/json"
 	"fmt"
 	"sync"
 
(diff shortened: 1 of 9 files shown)


pbacsko (approved):
+1 LGTM

zhuqi-lucas:
Thank you @pbacsko for review.

pkg/cache/nodes.go
```diff
@@ -101,20 +100,13 @@ func (nc *schedulerNodes) addAndReportNode(node *v1.Node, reportNode bool) {
...
-		nodeLabels, err := json.Marshal(node.Labels) // A nil pointer encodes as the "null" JSON value.
-		if err != nil {
-			log.Logger().Error("failed to marshall node labels to json", zap.Error(err))
-			nodeLabels = make([]byte, 0)
-		}
-
 		log.Logger().Info("adding node to context",
 			zap.String("nodeName", node.Name),
-			zap.String("nodeLabels", string(nodeLabels)),
+			zap.String("nodeLabels", fmt.Sprintf("%v", node.Labels)),
```
wilfred-s, line 105: Leverage zap.Any() it can handle the map with less overhead.
There will be no string allocated etc if the level is not set to INFO.

wilfred-s (changes requested):
one nit on the logging

pkg/cache/nodes.go
zhuqi-lucas: Thank you @wilfred-s for review, good suggestion, addressed in latest PR.

zhuqi-lucas:
Hi @wilfred-s , @pbacsko , @craigcondit  if any other suggestions about this PR? Thanks!

wilfred-s (approved):
LGTM

PR #520, opened 2023-02-03
[YUNIKORN-1526] support K8s pod overhead

zhuqi-lucas:
### What is this PR for?
With K8s 1.24 a pod can now provide an overhead in the pod spec: pod.Spec.Overhead. The pod overhead allows specifying an overhead based on the runtime set on the pod. For certain runtimes this overhead can be large.  See this [KEP for details](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/688-pod-overhead).

The scheduler should take into account this overhead if set on a pod. We currently calculate the size of the pod based on the containers but do not take the overhead into account. That needs to change.

We need to take the overhead into account as part of scheduling and quota calculations: include the pod.Spec.Overhead resources in the size of the pod before sending it to the core. Overhead only support cpu and memory and is added to the requests. The plugin framework (node related checks) calculates the pod size each time it is called (overhead!) and includes the overhead. The callback for the predicates must take that into account and not be broken by that implementation.

Adding clearly how we have calculated the overall pod size in the logging is a requirement.

Note from @wilfred-s :
The field is optional and will not be set in any pod, unless the cluster has been setup with it, even if you run 1.24 or later.
It is not a user managed field it is set by an admission controller. You cannot set the field on a pod when you create the pod. The field has been part of the offical pod spec since 1.16, and marked as a beta field since 1.18. It was only moved to GA in 1.24.
We need to handle it as an optional field. Even for a release that has the feature as GA the field can be a nil 

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1526

### Questions:

### DIFF
--- a/pkg/common/resource.go
+++ b/pkg/common/resource.go
@@ -79,6 +79,21 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
 		checkInitContainerRequest(pod, podResource)
 	}
 
+	// K8s pod EnableOverHead from:
+	// alpha: v1.16
+	// beta: v1.18
+	// Enables PodOverhead, for accounting pod overheads which are specific to a given RuntimeClass
+
+	// If Overhead is being utilized, add to the total requests for the pod
+	if pod.Spec.Overhead != nil {
+		podOverHeadResource := getResource(pod.Spec.Overhead)
+		podResource = Add(podResource, podOverHeadResource)
+		// Logging the overall pod size and pod overhead
+		log.Logger().Debug("We have calculated the overall pod size which includes the pod overhead",
+			zap.String("Pod overall size", podResource.String()),
+			zap.String("Pod overhead size", podOverHeadResource.String()))
+	}
+
 	return podResource
 }
 
--- a/pkg/common/resource.go
+++ b/pkg/common/resource.go
@@ -89,7 +89,8 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
 		podOverHeadResource := getResource(pod.Spec.Overhead)
 		podResource = Add(podResource, podOverHeadResource)
 		// Logging the overall pod size and pod overhead
-		log.Logger().Debug("We have calculated the overall pod size which includes the pod overhead",
+		log.Logger().Debug("Pod overhead specified, overall pod size adjusted",
+			zap.String("taskID", string(pod.UID)),
 			zap.String("Pod overall size", podResource.String()),
 			zap.String("Pod overhead size", podOverHeadResource.String()))
 	}
--- a/pkg/common/resource.go
+++ b/pkg/common/resource.go
@@ -91,8 +91,8 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
 		// Logging the overall pod size and pod overhead
 		log.Logger().Debug("Pod overhead specified, overall pod size adjusted",
 			zap.String("taskID", string(pod.UID)),
-			zap.String("Pod overall size", podResource.String()),
-			zap.String("Pod overhead size", podOverHeadResource.String()))
+			zap.Stringer("overallSize", podResource),
+			zap.Stringer("overheadSize", podOverHeadResource))
 	}
 
 	return podResource
--- a/pkg/common/resource_test.go
+++ b/pkg/common/resource_test.go
@@ -191,6 +191,19 @@ func TestParsePodResource(t *testing.T) {
 	assert.Equal(t, res.Resources[siCommon.CPU].GetValue(), int64(3000))
 	assert.Equal(t, res.Resources["nvidia.com/gpu"].GetValue(), int64(5))
 
+	// Add pod OverHead, only support CPU and Memory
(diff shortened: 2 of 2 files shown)


pkg/common/resource_test.go
```diff
@@ -191,6 +191,19 @@ func TestParsePodResource(t *testing.T) {
 	assert.Equal(t, res.Resources[siCommon.CPU].GetValue(), int64(3000))
 	assert.Equal(t, res.Resources["nvidia.com/gpu"].GetValue(), int64(5))
 
+	// Add pod OverHead, only support CPU and Memory
```
0yukali0, line 194: Hi @zhuqi-lucas , the getResource function would parse the resource which includes cpu, memory, gpu and other else.
Should we add filtering function to specific resource or a case which overhead contains different type of resources?

pkg/common/resource.go
```diff
@@ -79,6 +79,21 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
 		checkInitContainerRequest(pod, podResource)
 	}
 
+	// K8s pod EnableOverHead from:
+	// alpha: v1.16
+	// beta: v1.18
+	// Enables PodOverhead, for accounting pod overheads which are specific to a given RuntimeClass
+
+	// If Overhead is being utilized, add to the total requests for the pod
```
wilfred-s, line 87: you are skipping BestEffortQOS pods as they have already returned earlier in the function. A BestEffortQOS pod can have init containers filed a follow up jira for QOS BestEffort pod improvements

pkg/common/resource.go
```diff
@@ -79,6 +79,21 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
...
+	if pod.Spec.Overhead != nil {
+		podOverHeadResource := getResource(pod.Spec.Overhead)
+		podResource = Add(podResource, podOverHeadResource)
+		// Logging the overall pod size and pod overhead
+		log.Logger().Debug("We have calculated the overall pod size which includes the pod overhead",
```
wilfred-s, line 92: Text in the message with "we": who is "we"?
Make it simple and factual, include a pod reference in the message as this is not traceable to anything:
```
  log.Logger().Debug("Pod overhead specified, overall pod size adjusted",
  zap.String("taskID", pod.UID),
  zap.String("Pod overall size", podResource.String()),
  zap.String("Pod overhead size", podOverHeadResource.String()))
```

pkg/common/resource_test.go
wilfred-s: Nothing in the KEP says that there is a limit on the type of resource that can be added. That means we need to follow the second option and extend the UT

pkg/common/resource.go
zhuqi-lucas: Sure @wilfred-s , i see the follow-up now, i will add the corresponding things in
https://issues.apache.org/jira/browse/YUNIKORN-1559

pkg/common/resource.go
zhuqi-lucas: Good suggestion!
Addressed in latest PR.

pkg/common/resource_test.go
zhuqi-lucas: Thank you @wilfred-s and @0yukali0 for review, sure, addressed in latest PR.

pkg/common/resource.go
```diff
@@ -79,6 +79,22 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
...
+	// alpha: v1.16
+	// beta: v1.18
+	// Enables PodOverhead, for accounting pod overheads which are specific to a given RuntimeClass
+
+	// If Overhead is being utilized, add to the total requests for the pod
+	if pod.Spec.Overhead != nil {
+		podOverHeadResource := getResource(pod.Spec.Overhead)
+		podResource = Add(podResource, podOverHeadResource)
+		// Logging the overall pod size and pod overhead
+		log.Logger().Debug("Pod overhead specified, overall pod size adjusted",
```
craigcondit, line 92: Can we use zap.Stringer(...) instead of zap.String(...) here? Then .String() won't be called unless debug logging is enabled. Also, everywhere else we log using simple strings for zap keys (i.e. "overallSize" vs. "Pod overall size", "overheadSize" vs. "Pod overhead size". This helps consistency and makes it easier for tooling to parse.

craigcondit (changes requested):
Minor nit.

pkg/common/resource.go
zhuqi-lucas: Great suggestion @craigcondit, just noticed that the Stringer's String method is called lazily, this is very useful.

zhuqi-lucas: Addressed in latest PR now.

craigcondit (approved):
+1 LGTM.

PR #522, opened 2023-02-07
[YUNIKORN-1562] plugin: remove pod status update

wilfred-s:
### What is this PR for?
The k8shim in plugin mode updates the pod status to move the pod from the unschedulablePods structure to the activeQ. The state change was added to the plugin code based on an incorrect assumption that it would trigger a move between the unschedulablePods and activeQ. However, the scheduler explicitly excluded pod status when checking for updates.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1562

e2e tests for the plugin must pass


craigcondit (approved):
+1 will merge shortly.

PR #523, opened 2023-02-07
YUNIKORN-1564 Constants for node attributes are not consistent in scheduler interface and shim

zhuqi-lucas:
### What is this PR for?
Before we do more changes to remove nodename noderack in node attribute, and then add noderack to nodeinfo, also change in Scheduler Interface and webapp in https://issues.apache.org/jira/browse/YUNIKORN-1560 

We first fix the consistence with interface and shim, then i will do above things.

And don't change the nodelabels field, because we will address new nodelabels pass instead of the json now in another jira:
https://issues.apache.org/jira/browse/YUNIKORN-1550

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1564

### Questions:

### DIFF
--- a/pkg/common/constants/constants.go
+++ b/pkg/common/constants/constants.go
@@ -23,8 +23,6 @@ const True = "true"
 const False = "false"
 
 // Cluster
-const DefaultNodeAttributeHostNameKey = "si.io/hostname"
-const DefaultNodeAttributeRackNameKey = "si.io/rackname"
 const DefaultNodeAttributeNodeLabelsKey = "si.io/nodelabels"
 const DefaultRackName = "/rack-default"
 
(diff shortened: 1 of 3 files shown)


wilfred-s:
See the comment in [YUNIKORN-1560](https://issues.apache.org/jira/browse/YUNIKORN-1560?focusedCommentId=17687660&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17687660) I don't think this is needed as it will be  overwritten by that change.

wilfred-s:
This change is not needed, [YUNIKORN-1560](https://issues.apache.org/jira/browse/YUNIKORN-1560) will handle this

PR #527, opened 2023-02-09
[YUNIKORN-1569] Admission controller cannot update CronJobs

pbacsko:
### What is this PR for?
Admission controller cannot update the annotations on a CronJob, the API server throws an error after the patch is generated.
There is a type in the JSON path, we need to use "jobTemplate" and not "jobtemplate".

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1569

### Questions:

### DIFF
--- a/pkg/admission/metadata/usergroup.go
+++ b/pkg/admission/metadata/usergroup.go
@@ -44,7 +44,7 @@ func NewUserGroupAnnotationHandler(conf *conf.AdmissionControllerConf) *UserGrou
 
 const (
 	defaultPodAnnotationsPath = "/spec/template/metadata/annotations"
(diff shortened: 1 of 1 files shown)


wilfred-s (changes requested):
The change itself looks good, simple typo fix.
However this should have been caught in the e2e tests we have for the admission controller and they should be added.

wilfred-s (approved):
LGTM
@manirajv06 please look at this from an overall e2e fit

PR #530, opened 2023-02-14
[YUNIKORN-1571] cleanup old plugin deployment examples

0yukali0:
### What is this PR for?
Removing unused config in plugin.yaml example

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring
* [x] - Tasks 

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1571

image

image

### Questions:

### DIFF
--- a/deployments/scheduler/plugin.yaml
+++ b/deployments/scheduler/plugin.yaml
@@ -36,27 +36,6 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
 
 ---
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: yunikorn-configs
-  labels:
-    app: yunikorn
-data:
-  queues.yaml: |
-    partitions:
-      - name: default
-        nodesortpolicy:
-          type: fair
-        placementrules:
-          - name: tag
-            value: namespace
-            create: true
-        queues:
-          - name: root
-            submitacl: '*'
-
----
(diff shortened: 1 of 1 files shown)


deployments/scheduler/plugin.yaml
```diff
@@ -96,8 +97,6 @@ spec:
               cpu: 4
               memory: 2Gi
           volumeMounts:
-            - mountPath: /etc/kubernetes
-              name: scheduler-config
             - mountPath: /etc/yunikorn
               name: quotamanager-config
```
wilfred-s, line 101: With the new configuration v2 do we need these?
I could not find any mounting beside the secrets for the admission controller in the current helm charts.

@craigcondit

0yukali0: Hi @wilfred-s , you are right and I will remove this in next commit.
After I removed the quatamanager-config, latest yunikorn will use the default configuration here.
image

wilfred-s: That looks correct, let's see if we have got this all correct and get a review from Craig.

wilfred-s (approved):
LGTM

PR #532, opened 2023-02-15
[YUNIKORN-1555] Filter out terminated pods during recovery

bgrams:
### What is this PR for?
Excludes pods in a terminal state (Succeeded, Failed) during scheduler recovery. Applications may otherwise be recovered as "New" and remain there in scenarios where the application has completed and pods are awaiting garbage collection. 

This should not affect recovery behavior for applications with pods in non-terminal states as the application metadata will be recovered according to existing logic.   

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
[YUNIKORN-1555](https://issues.apache.org/jira/browse/YUNIKORN-1555)

e2e

### Questions:

### DIFF
--- a/pkg/appmgmt/appmgmt_recovery_test.go
+++ b/pkg/appmgmt/appmgmt_recovery_test.go
@@ -277,10 +277,10 @@ func (ma *mockedAppManager) Stop() {
 func (ma *mockedAppManager) ListPods() ([]*v1.Pod, error) {
 	pods := make([]*v1.Pod, 8)
 	pods[0] = newPodHelper("pod1", "task01", "app01", time.Unix(100, 0), v1.PodRunning)
-	pods[1] = newPodHelper("pod2", "task02", "app01", time.Unix(500, 0), v1.PodRunning)
+	pods[1] = newPodHelper("pod2", "task02", "app01", time.Unix(500, 0), v1.PodPending)
 	pods[2] = newPodHelper("pod3", "task03", "app01", time.Unix(200, 0), v1.PodSucceeded)
 	pods[3] = newPodHelper("pod4", "task04", "app02", time.Unix(400, 0), v1.PodRunning)
-	pods[4] = newPodHelper("pod5", "task05", "app02", time.Unix(300, 0), v1.PodRunning)
+	pods[4] = newPodHelper("pod5", "task05", "app02", time.Unix(300, 0), v1.PodPending)
 	pods[5] = newPodHelper("pod6", "task06", "app02", time.Unix(600, 0), v1.PodFailed)
 
 	// these pods and apps should never be recovered
--- a/pkg/appmgmt/appmgmt_recovery_test.go
+++ b/pkg/appmgmt/appmgmt_recovery_test.go
@@ -197,6 +197,40 @@ func TestAppStatesDuringRecovery(t *testing.T) {
 	assert.Equal(t, app02.GetApplicationState(), cache.ApplicationStates().Accepted)
 }
 
+func TestPodRecovery(t *testing.T) {
+	conf.GetSchedulerConf().OperatorPlugins = "mocked-app-manager"
+	amProtocol := cache.NewMockedAMProtocol()
+	apiProvider := client.NewMockedAPIProvider(false)
+	taskRequests := make([]*interfaces.AddTaskRequest, 0)
+	amProtocol.UseAddTaskFn(func(request *interfaces.AddTaskRequest) {
+		taskRequests = append(taskRequests, request)
+	})
+	amService := NewAMService(amProtocol, apiProvider)
+	amService.register(&mockedAppManager{})
+
+	apps, err := amService.recoverApps()
+	assert.NilError(t, err)
+	assert.Equal(t, 4, len(taskRequests))
+	assert.Equal(t, 2, len(apps))
+
+	expected := map[string]map[string]bool{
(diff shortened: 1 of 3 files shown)


pkg/appmgmt/general/general.go
```diff
@@ -236,7 +236,7 @@ func (os *Manager) ListPods() ([]*v1.Pod, error) {
 		log.Logger().Debug("Looking at pod for recovery candidates", zap.String("podNamespace", pod.Namespace), zap.String("podName", pod.Name))
 		// general filter passes, and pod is assigned
 		// this means the pod is already scheduled by scheduler for an existing app
-		if utils.GeneralPodFilter(pod) && utils.IsAssignedPod(pod) {
+		if utils.GeneralPodFilter(pod) && utils.IsAssignedPod(pod) && !utils.IsPodTerminated(pod) {
```
manirajv06, line 239: Can we use utils#NeedRecovery()?

manirajv06 (changes requested):
Thanks @bgrams. Yes, we need to ensure this has been covered as part of unit tests. FYI, already there are multiple unit tests available to verify different app recovery scenarios. Also, can we change draft status?

bgrams:
> Thanks @bgrams. Yes, we need to ensure this has been covered as part of unit tests. FYI, already there are multiple unit tests available to verify different app recovery scenarios. Also, can we change draft status?

Thanks @manirajv06. I'll update tests to make sure we have coverage. 

I marked this as draft after noticing that we also require changes somewhere in `PodEventHandler#handleEventFromInformers` since there's overlapping recovery logic. Thanks for bearing with me while I figure this one out 😄

pkg/appmgmt/general/general.go
bgrams: Ah yes, perfect

wilfred-s:
The pod event handler will only be called if there is a new, updated or deleted pod while we are recovering the pods on the  list. The pods events that get cached during the recovery are not processed until after the recovery is done (see `PodEventHandlerRecoveryDone()`). At that point filtering out pod that have finished could mean that you miss a valid event. A pod received as part of the list might be running. While the pod recovery is processing the pod exits and thus goes into a finished state. You need that event to correctly delete the allocation.

wilfred-s (changes requested):
We need a test for this change too

pbacsko:
@bgrams do you have any progress with this? We got contacted by another user upstream on Slack who've seen the same issue.

bgrams:
Hey @pbacsko thanks for the nudge here. I completely forgot about this. Let me take a stab at it this weekend.

pkg/appmgmt/general/general.go
bgrams: Addressed in latest commits

wilfred-s (changes requested):
Change looks good. One test remark: pod state == Pending will also be recovered if it is assigned and scheduled by YuniKorn. That case is not covered in the test.
Pending could happen if the pod is assigned to the node but has not started all containers yet. The test has 2 running containers on each app. At least 1 container should have a state of pending.
@pbacsko and @craigcondit please correct me if I am wromg

pkg/appmgmt/appmgmt_recovery_test.go
```diff
@@ -197,6 +197,40 @@ func TestAppStatesDuringRecovery(t *testing.T) {
...
+	assert.Equal(t, 2, len(apps))
+
+	expected := map[string]map[string]bool{
+		"app01": {
+			"task01": true,
+			"task02": true,
+		},
+		"app02": {
+			"task04": true,
+			"task05": true,
```
pbacsko, line 223: I haven't deeply analized what happens here, but don't we want some variety? In all cases, we expect true, so why do we have this here?

pkg/appmgmt/appmgmt_recovery_test.go
```diff
@@ -241,11 +275,17 @@ func (ma *mockedAppManager) Stop() {
...
-	pods[1] = newPodHelper("pod2", "task02", "app01", time.Unix(500, 0))
-	pods[2] = newPodHelper("pod3", "task03", "app02", time.Unix(200, 0))
-	pods[3] = newPodHelper("pod4", "task04", "app02", time.Unix(300, 0))
+	pods := make([]*v1.Pod, 8)
+	pods[0] = newPodHelper("pod1", "task01", "app01", time.Unix(100, 0), v1.PodRunning)
+	pods[1] = newPodHelper("pod2", "task02", "app01", time.Unix(500, 0), v1.PodPending)
+	pods[2] = newPodHelper("pod3", "task03", "app01", time.Unix(200, 0), v1.PodSucceeded)
+	pods[3] = newPodHelper("pod4", "task04", "app02", time.Unix(400, 0), v1.PodRunning)
+	pods[4] = newPodHelper("pod5", "task05", "app02", time.Unix(300, 0), v1.PodPending)
+	pods[5] = newPodHelper("pod6", "task06", "app02", time.Unix(600, 0), v1.PodFailed)
```
pbacsko, line 284: What if we have app03, and task1-4 is PodSucceeded, but task5 (the last one) is PodRunning?
What's the result of that?

PR #539, opened 2023-02-21
[YUNIKORN-1595] postTaskAllocated uses same event details for different things

zhuqi-lucas:
### What is this PR for?
The volume bind and pod to node bind failures use the same event content:

https://github.com/apache/yunikorn-k8shim/blob/master/pkg/cache/task.go#L383
and 
https://github.com/apache/yunikorn-k8shim/blob/master/pkg/cache/task.go#L396

That is confusing, we should change the description text.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1595

### Questions:


pbacsko:
+1

PR #541, opened 2023-02-23
[YUNIKORN-1597] Gang scheduling: application might not transition to Running after recovery

pbacsko:
### What is this PR for?
If all remaining placeholders are in Running state after recovery, Yunikorn fails to advance the status of the application to Running.

No tests in this PR. It's very difficult to trigger this state in e2e tests and MockScheduler ignores the recovery entirely. Unit tests seem to be feasible as a follow-up.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1597

Yunikorn needs code change.
1. a call to `panic()` in `Application.onReservationStateChange()` before calling the dispatcher.
2. However, there is at least one placeholder task in `Application.onReserving()` (new `if` block), set a boolean flag and don't call `panic()`.

### Questions:

### DIFF
--- a/pkg/cache/task.go
+++ b/pkg/cache/task.go
@@ -213,12 +213,12 @@ func (task *Task) initialize() {
 
 	// task needs recovery means the task has already been
 	// scheduled by us with an allocation, instead of starting
-	// from New, directly set the task to Allocated.
+	// from New, directly set the task to Bound.
 	if utils.NeedRecovery(task.pod) {
 		task.allocationUUID = string(task.pod.UID)
 		task.nodeName = task.pod.Spec.NodeName
-		task.sm.SetState(TaskStates().Allocated)
-		log.Logger().Info("set task as Allocated",
+		task.sm.SetState(TaskStates().Bound)
+		log.Logger().Info("set task as Bound",
 			zap.String("appID", task.applicationID),
 			zap.String("taskID", task.taskID),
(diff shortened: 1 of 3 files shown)


pkg/cache/task.go
```diff
@@ -213,12 +213,12 @@ func (task *Task) initialize() {
...
 	// task needs recovery means the task has already been
 	// scheduled by us with an allocation, instead of starting
-	// from New, directly set the task to Allocated.
+	// from New, directly set the task to Bound.
 	if utils.NeedRecovery(task.pod) {
 		task.allocationUUID = string(task.pod.UID)
 		task.nodeName = task.pod.Spec.NodeName
-		task.sm.SetState(TaskStates().Allocated)
-		log.Logger().Info("set task as Allocated",
+		task.sm.SetState(TaskStates().Bound)
```
pbacsko, line 220: After analyzing the existing code, this seems to be a safe change. It is more sound to use Bound.

wilfred-s: See below

wilfred-s (approved):
LGTM

PR #542, opened 2023-02-24
[YUNIKORN-1601] [WIP] Shim: Implement preemption predicate handling

craigcondit:
### What is this PR for?
Add PreemptionPredicates() RM callback implementation to support preemption in core.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [x] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1601

Manually verified that algorithm functions when combined with https://github.com/apache/yunikorn-core/pull/511.
Unit tests (TODO) and eventually e2e tests once core is updated.

### Questions:


pkg/cache/context.go
```diff
@@ -454,6 +454,41 @@ func (ctx *Context) IsPodFitNode(name, node string, allocate bool) error {
...
+		// if pod exists in cache, try to run predicates
+		if targetNode := ctx.schedulerCache.GetNode(node); targetNode != nil {
+			// need to lock cache here as predicates need a stable view into the cache
+			ctx.schedulerCache.LockForReads()
+			defer ctx.schedulerCache.UnlockForReads()
+
+			// look up each victim in the scheduler cache
+			victims := make([]*v1.Pod, 0)
+			for _, uid := range allocations {
+				if victim, ok := ctx.schedulerCache.GetPod(uid); ok {
```
pbacsko, line 475: `GetPod()` also acquires a read lock, this call blocks here.

pkg/cache/context.go
```diff
@@ -454,6 +454,41 @@ func (ctx *Context) IsPodFitNode(name, node string, allocate bool) error {
 	return fmt.Errorf("predicates were not running because pod or node was not found in cache")
 }
 
+func (ctx *Context) IsPodFitNodeViaPreemption(name, node string, allocations []string, startIndex int) (index int, ok bool) {
```
pbacsko, line 457: Please add unit test coverage in context_test.go.

pbacsko (changes requested):
Need unit test for `IsPodFitNodeViaPreemption()`.

pkg/cache/context.go
craigcondit: No, it doesn't. It takes the cache lock (which is a different lock). This mirrors the existing code in `IsPodFitNode()` which has been working reliably forever.

pbacsko: Thanks, looks like I missed that.

pkg/cache/context.go
craigcondit: This is going to be virtually impossible for the same reasons we also don't have coverage for `IsPodFitNode()` as doing so would require mocking essentially the entire stack underneath. This is why we will do this via e2e tests once the core is complete.

pkg/cache/context.go
```diff
@@ -454,6 +454,41 @@ func (ctx *Context) IsPodFitNode(name, node string, allocate bool) error {
...
+	// assume minimal pods need killing if running in testing mode
+	if ctx.apiProvider.IsTestingMode() {
+		return startIndex, ok
+	}
+
+	ctx.lock.RLock()
+	defer ctx.lock.RUnlock()
+	if pod, ok := ctx.schedulerCache.GetPod(name); ok {
+		// if pod exists in cache, try to run predicates
+		if targetNode := ctx.schedulerCache.GetNode(node); targetNode != nil {
```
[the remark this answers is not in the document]
craigcondit: I disagree, the code as written is idiomatic go. Adding an early return and extra variable to avoid one level of nesting is not a good tradeoff IMO.

pkg/cache/context.go
```diff
@@ -454,6 +454,41 @@ func (ctx *Context) IsPodFitNode(name, node string, allocate bool) error {
...
 
+func (ctx *Context) IsPodFitNodeViaPreemption(name, node string, allocations []string, startIndex int) (index int, ok bool) {
+	// assume minimal pods need killing if running in testing mode
+	if ctx.apiProvider.IsTestingMode() {
+		return startIndex, ok
+	}
+
+	ctx.lock.RLock()
+	defer ctx.lock.RUnlock()
+	if pod, ok := ctx.schedulerCache.GetPod(name); ok {
```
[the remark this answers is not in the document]
craigcondit: Same as above, this is idiomatic go. Adding early return and extra variable in outer scope to avoid one level of nesting is not a good tradeoff.

craigcondit:
> Need unit test for `IsPodFitNodeViaPreemption()`.

Again, same as above, this would require mocking out a substantial amount of the underlying code, which doesn't really verify that it works properly. As with `IsPodFitNode()` (which this mirrors) we will test this primarily with e2e tests.

pbacsko (approved):
Questioms, concerns have been addressed.

+1 LGTM

PR #543, opened 2023-02-27
[YUNIKORN-1609] Make api client timeouts configurable

elihschiff:
### What is this PR for?

Right now there are 2 places in the scheduler with a 30 second timeout hard coded. It would be nice if those timeouts were user configurable. I was running into some issues in my cluster where 30 seconds was not enough.

Feel free to provide feedback, this is my first real change to yunikorn and I am more than happy to make improvements to my code.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
~~I have an update to the yunikorn-site coming as well. Just need to open a PR for it after I finish this one.~~ Docs pr here https://github.com/apache/yunikorn-site/pull/271

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1609

I updated some unit tests to include my change. I also was able to test this change in my k8s cluster and it solved the issue for me.

### Questions:

### DIFF
--- a/pkg/conf/schedulerconf.go
+++ b/pkg/conf/schedulerconf.go
@@ -63,6 +63,7 @@ const (
 	CMSvcDisableGangScheduling  = PrefixService + "disableGangScheduling"
 	CMSvcEnableConfigHotRefresh = PrefixService + "enableConfigHotRefresh"
 	CMSvcPlaceholderImage       = PrefixService + "placeholderImage"
+	CMSvcApiClientTimeout       = PrefixService + "apiClientTimeout"
 
 	// log
 	CMLogLevel = PrefixLog + "level"
@@ -86,6 +87,7 @@ const (
 	DefaultLogEncoding            = "console"
 	DefaultKubeQPS                = 1000
 	DefaultKubeBurst              = 1000
+	DefaultApiClientTimeout       = 30 * time.Second
 )
 
 var (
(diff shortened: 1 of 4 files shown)


pkg/conf/schedulerconf.go
```diff
@@ -63,6 +63,7 @@ const (
 	CMSvcDisableGangScheduling  = PrefixService + "disableGangScheduling"
 	CMSvcEnableConfigHotRefresh = PrefixService + "enableConfigHotRefresh"
 	CMSvcPlaceholderImage       = PrefixService + "placeholderImage"
+	CMSvcApiClientTimeout       = PrefixService + "apiClientTimeout"
```
craigcondit, line 66: Can we use "kubernetes.apiClientTimeout" instead (i.e. PrefixKubernetes). This aligns better with the existing settings.

craigcondit (changes requested):
Looks good, except we should put this under the kubernetes section.

pkg/conf/schedulerconf.go
elihschiff: That makes sense. This should now be resolved. I also updated my `yunikorn-site` repo PR to reflect this change in the prefix value.

craigcondit (approved):
+1 pending pre-commit checks. Will commit once those have passed.

PR #544, opened 2023-03-01
[YUNIKORN-1597] e2e tests: Recover assigned pods in Bound state

pbacsko:
### What is this PR for?
The PR contains an e2e test which validates that YK can recover properly with Running/Pending placeholders.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1597

### Questions:

### DIFF
--- a/scripts/run-e2e-tests.sh
+++ b/scripts/run-e2e-tests.sh
@@ -208,7 +208,13 @@ function install_cluster() {
     --set admissionController.image.pullPolicy=IfNotPresent \
     --set web.image.repository=local/yunikorn \
     --set web.image.tag="${WEBTEST_IMAGE}" \
-    --set web.image.pullPolicy=IfNotPresent
+    --set web.image.pullPolicy=IfNotPresent \
+    --set tolerations[0].operator=Equal,tolerations[0].effect=NoSchedule,tolerations[0].key="node-role.kubernetes.io/control-plane" \
+    --set tolerations[1].operator=Equal,tolerations[1].effect=NoSchedule,tolerations[1].key="node-role.kubernetes.io/master" \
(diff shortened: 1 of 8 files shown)


craigcondit:
I don't think we want to add the taints/tolerations to helm or refer to kind directly, as these tests are meant to be runnable on all types of clusters (not just kind). Can we eliminate this?

craigcondit:
Alternate approach: Query the cluster for nodes with available capacity, and construct nodes with nodeAffinity specified so that they are only schedulable on those nodes.

pbacsko:
> I'm -1 on this approach as it doesn't allow for running on anything but a Kind cluster. Our tests should be cluster-agnostic, and adapt to the environment they are run in so that they can be used on real clusters as well.
Yeah, most tests are agnostic, but there are some (maybe just the one which I modified) which already depends on having 3 nodes (1 control + 2 worker). Perhaps the existing tests should be fixed, too in a follow-up JIRA.

> Alternate approach: Query the cluster for nodes with available capacity, and construct pods with nodeAffinity specified so that they are only schedulable on those nodes.
Hm, interesting. I'll try this.

pbacsko:
Test failure seems to be unrelated.

manirajv06:
@pbacsko [PR](https://github.com/apache/yunikorn-k8shim/commit/817042e2e4b3e79b222f34132795f80ecc1e22ef) has introduced new methods to set/remove node labels etc. There was a need to add taint/untaint node methods as well to remove the dependency on k8s test framework. Hence used some methods from this PR to above said PR. So, Please rebase your wc and take it forward.

craigcondit (approved):
+1 LGTM.

PR #546, opened 2023-03-06
[YUNIKORN-1622] K8shim: Race in allocate vs pod delete

wilfred-s:
### What is this PR for?
When a pod is allocated immediately set the allocation UUID and the node as part of the transition.
Make sure that the rejected to failed transition removes lingering asks. Set the node info in the task when it is returned by the core. Dedupe log messages on task transition.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1622

No regressions in any test suite.
Fix itself is difficult to test as the issue is caused by a race in two go routines

### DIFF
--- a/pkg/cache/application_test.go
+++ b/pkg/cache/application_test.go
@@ -848,6 +848,10 @@ func TestTryReservePostRestart(t *testing.T) {
 			Containers: containers,
 		},
 	})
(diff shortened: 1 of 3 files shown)


craigcondit (approved):
+1 LGTM.

PR #548, opened 2023-03-13
[YUNIKORN-1627] Admission controller triggers infinite recreation of ReplicaSet objects

zhuqi-lucas:
### What is this PR for?
[YUNIKORN-1338](https://issues.apache.org/jira/browse/YUNIKORN-1338) added support in the admission controller for mutating workload objects other than Pods for the purposes of user tracking (and eventual enforcement). This functionality works as designed.

However, in the case where a ReplicaSet is managed by a Deployment, the changes to the ReplicaSet pod template trigger the associated Deployment to recreate the ReplicaSet without the changes, which the admission controller then updates, triggering another ReplicaSet, etc. ad infinitum. 

The workaround for now is to set the following in the yunikorn-configs ConfigMap, which disables the functionality:

admissionController.accessControl.bypassAuth: "true"
The real fix needs to take into account any ownerReference in a workload which points to another workload type that we manage. This definitely includes ReplicaSet -> Deployment, but could include other workload relationships (more investigation needed on this).

For ReplicaSets, if the ownerReference is set to a Deployment, it should be sufficient to check for that and assume that the Deployment itself has passed in the appropriate user information due to the admission controller having mutated the Deployment.

### What type of PR is it?
* [x] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1627

### Questions:

### DIFF
--- a/pkg/admission/admission_controller_test.go
+++ b/pkg/admission/admission_controller_test.go
@@ -580,6 +580,53 @@ func TestMutate(t *testing.T) {
 	resp = ac.mutate(req)
 	assert.Check(t, resp.Allowed, "response not allowed for unknown object type")
 	assert.Equal(t, len(resp.Patch), 0, "non-empty patch for deployment")
+
+	// replicaSet - annotation is set
+	replicaSet := &appsv1.ReplicaSet{
+		Spec: appsv1.ReplicaSetSpec{
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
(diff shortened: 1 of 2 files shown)


craigcondit:
@zhuqi-lucas we need to take a different approach due to the security considerations of this. I've asked @pbacsko to handle it.

PR #551, opened 2023-03-15
[YUNIKORN-1633] Add unit tests for pkg/cache/amprotocol_mock.go

wusamzong:
### What is this PR for?
Add unit test for pkg/cache/amprotocol_mock.go

### What type of PR is it?
* [x] - Improvement

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1633/

### Questions:

### DIFF
--- a/pkg/cache/amprotocol_mock_test.go
+++ b/pkg/cache/amprotocol_mock_test.go
@@ -1,0 +1,227 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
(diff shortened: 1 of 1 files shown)


craigcondit (changes requested):
amprotocol_mock *is* test code. Adding tests for it is redundant. Instead, simply rename the amprotocol_mock.go file to amprotocol_mock_test.go, as it is only referenced in test code.

wusamzong:
> amprotocol_mock _is_ test code. Adding tests for it is redundant. Instead, simply rename the amprotocol_mock.go file to amprotocol_mock_test.go, as it is only referenced in test code.

@wilfred-s thanks for your review
After renaming the file to amprotocol_mock_test.go, other packages will not be able to use MockedAMProtocol (e.g. appmgmt_test.go). I wonder if there is a way to import *_test.go into other files? If not, I can keep the original method and close this issue.

craigcondit:
Given the mock is used in other packages, I'd recommend just closing this issue.

PR #553, opened 2023-03-20
[YUNIKORN-1637] Skip the spark installation if spark is installed

0yukali0:
### What is this PR for?
If the spark directory exists, the installation of spark will use the spark directory directly.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1637

`./scripts/run-e2e-tests.sh -a test -n yk8s -v kindest/node:v1.26.0`

image

### Questions:

### DIFF
--- a/scripts/run-e2e-tests.sh
+++ b/scripts/run-e2e-tests.sh
@@ -105,11 +105,11 @@ function install_spark() {
   SPARK_VERSION="3.3.1"
   DOWNLOAD="${CURRENT}/_spark"
   SPARK_SUBMIT_CMD="${DOWNLOAD}/spark/bin/spark-submit"
-  if [[ ! -d $DOWNLOAD || ! -f $SPARK_SUBMIT_CMD ]]; then
+  if [[ ! -d "${DOWNLOAD}" || ! -f "${SPARK_SUBMIT_CMD}" ]]; then
     echo  "Installing spark in yunikorn rep."
-    rm -rf $DOWNLOAD
-    mkdir $DOWNLOAD \
-      && cd $DOWNLOAD \
+    rm -rf "${DOWNLOAD}"
+    mkdir "${DOWNLOAD}" \
+      && cd "${DOWNLOAD}" \
       && curl "https://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop3.tgz" | tar xzvf - \
       && mv "spark-${SPARK_VERSION}-bin-hadoop3" spark \
       && chmod +x spark
@@ -189,7 +189,6 @@ function install_cluster() {
 
   # create K8s cluster
   echo "step 8/10: installing K8s cluster using kind"
-  echo $(pwd)
   "${KIND}" create cluster --name "${CLUSTER_NAME}" --image "${CLUSTER_VERSION}" --config="${KIND_CONFIG}"
   exit_on_error "install K8s cluster failed"
   kubectl cluster-info --context kind-"${CLUSTER_NAME}"
--- a/scripts/run-e2e-tests.sh
+++ b/scripts/run-e2e-tests.sh
@@ -101,7 +101,7 @@ function install_kind() {
 }
 
 function install_spark() {
-  if [ ! -d "/opt/spark-3.3.1-bin-hadoop3" ] && [ $(find "/opt/spark-3.3.1-bin-hadoop3" -type f | wc -l) > 0 ]; then
+  if [ ! -d "/opt/spark-3.3.1-bin-hadoop3" ] || [ ! $(find "/opt/spark-3.3.1-bin-hadoop3" -type f | wc -l) > 0 ]; then
     echo  "/opt doesn't contains spark."
     check_cmd "wget"
     wget -qO-  https://archive.apache.org/dist/spark/spark-3.3.1/spark-3.3.1-bin-hadoop3.tgz | tar xzvf - && chmod +x spark-3.3.1-bin-hadoop3 && sudo mv spark-3.3.1-bin-hadoop3 /opt/.
--- a/scripts/run-e2e-tests.sh
+++ b/scripts/run-e2e-tests.sh
@@ -101,18 +101,18 @@ function install_kind() {
 }
 
 function install_spark() {
-  if [ ! -d "/opt/spark-3.3.1-bin-hadoop3" ]; then
-    echo  "/opt doesn't contains spark."
+  if [ ! -d ./_spark ]; then
+    echo  "Installing spark in yunikorn rep."
     check_cmd "wget"
-    wget -qO-  https://archive.apache.org/dist/spark/spark-3.3.1/spark-3.3.1-bin-hadoop3.tgz | tar xzvf - && chmod +x spark-3.3.1-bin-hadoop3 && sudo mv spark-3.3.1-bin-hadoop3 /opt/.
+    wget -qO- https://archive.apache.org/dist/spark/spark-3.3.1/spark-3.3.1-bin-hadoop3.tgz | tar xzvf - && chmod +x spark-3.3.1-bin-hadoop3 && sudo mv spark-3.3.1-bin-hadoop3 _spark
     exit_on_error "install spark failed."
   fi
-  if [ -h /opt/spark ]; then
-    sudo rm /opt/spark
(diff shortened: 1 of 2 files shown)


craigcondit (changes requested):
As long as we're fixing this, let's go one step further and not install to a system directory with sudo. We should instead install into a temporary work directory within the source repository itself (obviously added to .gitignore).

0yukali0:
Hi @craigcondit, thanks for the review.
I will update this commit.

craigcondit (changes requested):
Still have sudo here, also we should use curl instead of wget as wget is not installed by default on macOS.

scripts/run-e2e-tests.sh
```diff
@@ -101,12 +101,20 @@ function install_kind() {
...
+    echo  "Installing spark in yunikorn rep."
+    check_cmd "wget"
+    mkdir "${PWD}/_spark"
+    wget -qO- https://archive.apache.org/dist/spark/spark-3.3.1/spark-3.3.1-bin-hadoop3.tgz | tar xzvf - && chmod +x spark-3.3.1-bin-hadoop3 && sudo mv spark-3.3.1-bin-hadoop3 "${PWD}/_spark/."
+    exit_on_error "install spark failed."
+  fi
+  if [ -h "${PWD}/_spark/spark" ]; then
+    rm "${PWD}/_spark/spark"
+  fi
+  sudo ln -s "${PWD}/_spark/spark-3.3.1-bin-hadoop3" "${PWD}/_spark/spark"
```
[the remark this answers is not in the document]
manirajv06: Yes, that way we can close https://issues.apache.org/jira/browse/YUNIKORN-1663 and https://issues.apache.org/jira/browse/YUNIKORN-1662

scripts/run-e2e-tests.sh
```diff
@@ -101,12 +101,15 @@ function install_kind() {
...
-  check_cmd "wget"
-  wget -qO-  https://archive.apache.org/dist/spark/spark-3.3.1/spark-3.3.1-bin-hadoop3.tgz | tar xzvf - && chmod +x spark-3.3.1-bin-hadoop3 && sudo mv spark-3.3.1-bin-hadoop3 /opt/.
-  exit_on_error "install spark failed."
-  sudo ln -s /opt/spark-3.3.1-bin-hadoop3 /opt/spark
-  exit_on_error "install spark failed. unable to create symlink"
-  export SPARK_HOME=/opt/spark/
+  PWD=$(pwd)
+  if [ ! -d "${PWD}/_spark" ]; then
+    echo  "Installing spark in yunikorn rep."
+    check_cmd "wget"
```
craigcondit, line 107: Remove this as we no longer need wget.

scripts/run-e2e-tests.sh
```diff
@@ -101,12 +101,15 @@ function install_kind() {
...
+    mkdir "${PWD}/_spark"
+    curl https://archive.apache.org/dist/spark/spark-3.3.1/spark-3.3.1-bin-hadoop3.tgz | tar xzvf - && chmod +x spark-3.3.1-bin-hadoop3 && mv spark-3.3.1-bin-hadoop3 "${PWD}/_spark/."
```
craigcondit, line 109: Instead of moving, change into the _spark directory first -- this way the archive will go into the correct place and we don't litter up the top level directory if the script aborts.

scripts/run-e2e-tests.sh
```diff
@@ -101,12 +101,15 @@ function install_kind() {
...
 
 function install_spark() {
-  check_cmd "wget"
-  wget -qO-  https://archive.apache.org/dist/spark/spark-3.3.1/spark-3.3.1-bin-hadoop3.tgz | tar xzvf - && chmod +x spark-3.3.1-bin-hadoop3 && sudo mv spark-3.3.1-bin-hadoop3 /opt/.
-  exit_on_error "install spark failed."
-  sudo ln -s /opt/spark-3.3.1-bin-hadoop3 /opt/spark
-  exit_on_error "install spark failed. unable to create symlink"
-  export SPARK_HOME=/opt/spark/
+  PWD=$(pwd)
+  if [ ! -d "${PWD}/_spark" ]; then
```
craigcondit, line 105: This should probably check for the whole directory we want to use (i.e. _spark/spark-3.3.1-bin-hadoop3), and if not, start with `rm -rf "${PWD}_spark"`. This way we can transparently upgrade in the future.

craigcondit (changes requested):
See comments.

scripts/run-e2e-tests.sh
```diff
@@ -101,14 +101,24 @@ function install_kind() {
...
+  SPARK_VERSION="3.3.1"
+  DOWNLOAD="${CURRENT}/_spark"
+  SPARK_SUBMIT_CMD="${DOWNLOAD}/spark/bin/spark-submit"
+  if [[ ! -d "${DOWNLOAD}" || ! -f "${SPARK_SUBMIT_CMD}" ]]; then
+    echo  "Installing spark in yunikorn rep."
+    rm -rf "${DOWNLOAD}"
+    mkdir "${DOWNLOAD}" \
+      && cd "${DOWNLOAD}" \
+      && curl "https://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop3.tgz" | tar xzvf - \
+      && mv "spark-${SPARK_VERSION}-bin-hadoop3" spark \
```
craigcondit, line 114: Don't move the downloaded dir, just use it as-is.

scripts/run-e2e-tests.sh
```diff
@@ -101,14 +101,24 @@ function install_kind() {
...
-  wget -qO-  https://archive.apache.org/dist/spark/spark-3.3.1/spark-3.3.1-bin-hadoop3.tgz | tar xzvf - && chmod +x spark-3.3.1-bin-hadoop3 && sudo mv spark-3.3.1-bin-hadoop3 /opt/.
-  exit_on_error "install spark failed."
-  sudo ln -s /opt/spark-3.3.1-bin-hadoop3 /opt/spark
-  exit_on_error "install spark failed. unable to create symlink"
-  export SPARK_HOME=/opt/spark/
-  export SPARK_PYTHON_IMAGE=docker.io/apache/spark-py:v3.3.1
+  CURRENT=$(pwd)
+  SPARK_VERSION="3.3.1"
+  DOWNLOAD="${CURRENT}/_spark"
+  SPARK_SUBMIT_CMD="${DOWNLOAD}/spark/bin/spark-submit"
```
craigcondit, line 107: Set this to "${DOWNLOAD}/spark-${SPARK_VERSION}-bin-hadoop3/bin/spark-submit", then you don't have to move the archive, and you can properly detect if this needs a new version installed (for example to move to 3.3.2). In that case, the test will succeed, even though the wrong version of spark is downloaded. If you use the fully-qualitifed path, the expected version will not exist, and we will re-download as expected.

scripts/run-e2e-tests.sh
```diff
@@ -101,14 +101,24 @@ function install_kind() {
...
+    echo  "Installing spark in yunikorn rep."
+    rm -rf "${DOWNLOAD}"
+    mkdir "${DOWNLOAD}" \
+      && cd "${DOWNLOAD}" \
+      && curl "https://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop3.tgz" | tar xzvf - \
+      && mv "spark-${SPARK_VERSION}-bin-hadoop3" spark \
+      && chmod +x spark
+    exit_on_error "install spark failed."
+  fi
+  export SPARK_HOME="${DOWNLOAD}/spark"
```
craigcondit, line 118: Set this to "${DOWNLOAD}/spark-${SPARK_VERSION}-bin-hadoop3"

craigcondit (changes requested):
See comments.

scripts/run-e2e-tests.sh
0yukali0: Ok

craigcondit (approved):
+1 LGTM.

PR #555, opened 2023-03-20
[YUNIKORN-1642] Scheduler recovery failed due to listing operation timeout

yangwwei:
### What is this PR for?
The listing operation in the recovery phase: https://github.com/apache/yunikorn-k8shim/blob/c25ac60ffbc175c4966f917da21d184f34dea7b4/pkg/client/apifactory.go#L225. This could sometimes fail on some large clusters, the response time from API server is not guaranteed.  Since it is a WARN, it continues but the informers did not return anything. This confuses the scheduler that nothing needs to be recovered, and it goes ahead doing the scheduling. This causes subsequential scheduler failures. And eventually, nothing can be scheduled anymore.

This should be a FATAL error. So the scheduler can be restarted to retry recovering. 

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-1642

### Questions:

### DIFF
--- a/pkg/client/clients.go
+++ b/pkg/client/clients.go
@@ -19,6 +19,7 @@
 package client
 
 import (
+	"go.uber.org/zap"
 	"time"
 
 	"github.com/apache/yunikorn-k8shim/pkg/client/informers/externalversions/yunikorn.apache.org/v1alpha1"
@@ -31,6 +32,7 @@ import (
 
 	appclient "github.com/apache/yunikorn-k8shim/pkg/client/clientset/versioned"
 	"github.com/apache/yunikorn-k8shim/pkg/conf"
+	"github.com/apache/yunikorn-k8shim/pkg/log"
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/api"
 )
 
@@ -69,6 +71,8 @@ func (c *Clients) GetConf() *conf.SchedulerConf {
 }
 
 func (c *Clients) WaitForSync() {
+	timeNow := time.Now()
+	counter := 0
 	for {
 		if c.NodeInformer.Informer().HasSynced() &&
 			c.PodInformer.Informer().HasSynced() &&
@@ -82,6 +86,16 @@ func (c *Clients) WaitForSync() {
 			return
 		}
 		time.Sleep(time.Second)
+		counter++
+		if counter%10 == 0 {
+			if counter > 300 {
+				log.Logger().Warn("Still waiting for informers to sync, delay seems longer than usual...",
+					zap.Duration("timeElapsed", time.Since(timeNow).Round(time.Second)))
+			} else {
+				log.Logger().Info("Waiting for informers to sync",
+					zap.Duration("timeElapsed", time.Since(timeNow).Round(time.Second)))
+			}
+		}
 	}
 }
 
--- a/pkg/client/clients.go
+++ b/pkg/client/clients.go
@@ -30,7 +30,6 @@ import (
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/volumebinding"
 
 	appclient "github.com/apache/yunikorn-k8shim/pkg/client/clientset/versioned"
-	"github.com/apache/yunikorn-k8shim/pkg/common/utils"
 	"github.com/apache/yunikorn-k8shim/pkg/conf"
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/api"
 )
@@ -69,10 +68,9 @@ func (c *Clients) GetConf() *conf.SchedulerConf {
 	return c.conf
 }
 
-func (c *Clients) WaitForSync(interval time.Duration, timeout time.Duration) error {
-	return utils.WaitForCondition(func() bool {
-		// cache is re-sync'd when all informers are sync'd
-		return c.NodeInformer.Informer().HasSynced() &&
+func (c *Clients) WaitForSync() {
+	for {
+		if c.NodeInformer.Informer().HasSynced() &&
 			c.PodInformer.Informer().HasSynced() &&
 			c.PVCInformer.Informer().HasSynced() &&
 			c.PVInformer.Informer().HasSynced() &&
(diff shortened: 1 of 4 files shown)


wilfred-s:
> This should be a FATAL error. So the scheduler can be restarted to retry recovering.

I think that is a bad idea. What we do now is not right but just crashing the scheduler is not good. Just hoping that the next time we run recovery it does come back within the timeout is not right. Logging a WARN and proceeding is bad also.
We need to come up with a better solution for this issue.

-1 on changing to FATAL

yangwwei:
hi @wilfred-s 

This is not a WARN that can be skipped. Once this happens, the scheduler already runs into a bad state, and it will not work any longer. Keeping it running will cause the scheduling completely stuck. This is a FATAL error for sure. 

Solving this issue is another matter. As now this is configurable, we can possibly increase the timeout. But even with a longer timeout, it still can fail, when that happens, it should be a fatal error.

craigcondit:
This shouldn't have a timeout at all (I realize the API call requires one, but we should it as close to infinity as possible. Why? Because this is simply waiting for the K8s API server to synchronize. If the API server is temporarily unavailable, or slow, or just takes 20 minutes to sync up, then that's what it takes. What we should probably do is wait for 10 seconds, log a message, wait for ten more, etc. in a loop with no end until it succeeds.

craigcondit:
> hi @craigcondit
> 
> I agree we should increase the timeout value. It is good now this is configurable. But I think we still need a timeout. If it hangs for a very long time (configurable value), e.g 10 or 15 minutes, if it is still stuck, we should let it fail. A failure can trigger alerts to notify us something is wrong; if it is stuck forever, it will take a long time for people to detect something is wrong.

No, this is completely WRONG. If it's taking 10-15 minutes, that probably indicates the API server is throttling, or we have a very large cluster. The worst thing to do would be die uncleanly and restart and go through the whole process again. In the case of API server overload, you're just adding more pain. The underlying informers handle failures gracefully, and retry during sync at a lower level. In fact, they can even tolerate the API server being gone for a very extended time. Imagine a reboot of a server that takes 15 minutes where the API server is running - this (while not common) does happen, and the informers will gracefully pick up where they left off when the API server returns. This will manifest as a longer than normal startup, but will otherwise succeed (if we let it). Nothing will ever be "stuck forever" as you say; any such behavior would indicate a severely broken cluster that cannot be dealt with in a sane fashion. 

Whatever timeout you choose, there will exist a cluster that just needs a few seconds longer, and now you've created pain where there shouldn't be. If startup time is a concern, add external checks on how long it takes before the scheduler reports healthy via APIs.

yangwwei:
hi @craigcondit  what you said is achievable, we can set `kubernetes.apiClientTimeout`, introduced via [YUNIKORN-1609](https://issues.apache.org/jira/browse/YUNIKORN-1609) to a big value, and keep it waiting. What I am saying is if eventually, it times out, we should let it fail, instead of giving a WARN and continuing to do the scheduling.

Or, are you suggesting we should remove the timeout here?

craigcondit:
> Or, are you suggesting we should remove the timeout here?

Yes, that's exactly what I'm suggesting. Describe the failure scenario where a timeout would be hit legitimately (i.e. not simply because we didn't set it high enough). I submit there isn't one, as the underlying K8s client should be doing reliable transport recovery. So, the best thing we can do is get out of the way. I *am* in favor of polling and logging that we're still waiting (with a timer for how long it has been) as this is useful diagnostic information. But we should get rid of the arbitrary timeout.

craigcondit (changes requested):
Looks good for the most part. Can we add some logging (say every 10 seconds or so) warning that we are still waiting on the informers to sync? Something like:

INFO Waiting for informers to sync (3m10s elapsed)

We could move up to WARN past a certain point as well (5m or so)?

yangwwei:
@craigcondit 

Got it. Please see the updates. This now prints logs like this:

it prints a message every 10s
the log prompts to WARN after 5m

```
2023-03-22T18:56:27.602-0700    INFO    cache/placeholder_manager.go:153        starting the PlaceholderManager
2023-03-22T18:56:37.611-0700    INFO    client/clients.go:96    Waiting for informers to sync   {"timeElapsed": "10s"}
...
2023-03-22T18:57:27.654-0700    INFO    client/clients.go:96    Waiting for informers to sync   {"timeElapsed": "1m0s"}
2023-03-22T18:57:37.662-0700    INFO    client/clients.go:96    Waiting for informers to sync   {"timeElapsed": "1m10s"}
2023-03-22T18:57:47.673-0700    INFO    client/clients.go:96    Waiting for informers to sync   {"timeElapsed": "1m20s"}
...
2023-03-22T18:59:40.310-0700    INFO    client/clients.go:93    Still waiting for informers to sync, delay seems longer than usual...   {"timeElapsed": "5m"}
2023-03-22T18:59:40.310-0700    INFO    client/clients.go:93    Still waiting for informers to sync, delay seems longer than usual...   {"timeElapsed": "5m10s"}
```

pkg/client/clients.go
```diff
@@ -69,19 +71,33 @@ func (c *Clients) GetConf() *conf.SchedulerConf {
...
-	}, interval, timeout)
+			(c.AppInformer == nil || c.AppInformer.Informer().HasSynced()) {
+			return
+		}
+		time.Sleep(time.Second)
+		counter++
+		if counter%10 == 0 {
+			if counter > 300 {
+				log.Logger().Warn("Still waiting for informers to sync, delay seems longer than usual...",
+					zap.Duration("timeElapsed", time.Since(timeNow).Round(time.Second)))
```
pbacsko, line 94: Nit: could we rename `timeNow`? It's not changing and it feels weird to constantly referenced as "now". Like "syncStartTime" or "startTime" sound much better.

craigcondit: I agree, that makes sense.

pkg/client/clients.go
```diff
@@ -69,19 +71,33 @@ func (c *Clients) GetConf() *conf.SchedulerConf {
...
-			(c.AppInformer == nil || c.AppInformer.Informer().HasSynced())
-	}, interval, timeout)
+			(c.AppInformer == nil || c.AppInformer.Informer().HasSynced()) {
+			return
+		}
+		time.Sleep(time.Second)
+		counter++
+		if counter%10 == 0 {
+			if counter > 300 {
+				log.Logger().Warn("Still waiting for informers to sync, delay seems longer than usual...",
```
craigcondit, line 93: I know I mentioned earlier that we could switch to a warning after a certain period of time, but after some more thought, I'd rather keep this consistent with the same message and INFO status. "Longer than usual" is both potentially incorrect and very opinionated. What's normal for one cluster may not be for another. The simple repeating of the INFO message every 10s should be sufficient. This can be monitored externally by checking the elapsed time field. To help with automated tools, we could also express it as a time in seconds rather than a Duration (i.e. `{"secondsElapsed": 300}`  instead of `{"timeElapsed": "5m0s"}`.

craigcondit (changes requested):
A few minor points.

PR #556, opened 2023-03-22
[YUNIKORN-1478] Move to go v1.20

wilfred-s:
### What is this PR for?
Update to latest go release

### What type of PR is it?
 - Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1478

standard build will run with new go and linter


Makefile
```diff
@@ -390,7 +390,7 @@ fsm_graph: clean
 .PHONY: clean
 clean:
 	@echo "cleaning up caches and output"
-	go clean -cache -testcache -r -x ./... 2>&1 >/dev/null
+	go clean -cache -testcache -r
```
craigcondit, line 393: I'm with Peter on this one (referencing his comments from the core version of this patch). I think dev/null handling should stay.

craigcondit: Upon further investigation, we don't need the redirect as -x is what was generating the noise. +1, will merge shortly.

PR #557, opened 2023-03-22
[YUNIKORN-1647] Dynamic namespace support in admission controller

wilfred-s:
### What is this PR for?
Support annotations on a namespace to define processing behaviour in the admission controller. The annotations have preference over the regexp. If no annotations are found fall back on the regexp. If an annotation is found it defines the behaviour and overrides what is defined in the regexp.

Annotations to be supported on a namespace:
* yunikorn.apache.org/namespace.generateAppId
* yunikorn.apache.org/namespace.enableYuniKorn

Supported values for the annotation: true and false.

### What type of PR is it?
* [X] - Feature

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-1647

Unit tests are added.

### DIFF
--- a/pkg/admission/admission_controller.go
+++ b/pkg/admission/admission_controller.go
@@ -67,6 +67,7 @@ var (
 type AdmissionController struct {
 	conf              *conf.AdmissionControllerConf
 	pcCache           *PriorityClassCache
+	nsCache           *NamespaceCache
 	annotationHandler *metadata.UserGroupAnnotationHandler
 	labelExtractor    metadata.LabelExtractor
 }
@@ -76,10 +77,11 @@ type ValidateConfResponse struct {
 	Reason  string `json:"reason"`
 }
 
-func InitAdmissionController(conf *conf.AdmissionControllerConf, pcCache *PriorityClassCache) *AdmissionController {
+func InitAdmissionController(conf *conf.AdmissionControllerConf, pcCache *PriorityClassCache, nsCache *NamespaceCache) *AdmissionController {
 	hook := &AdmissionController{
 		conf:              conf,
 		pcCache:           pcCache,
+		nsCache:           nsCache,
 		annotationHandler: metadata.NewUserGroupAnnotationHandler(conf),
 	}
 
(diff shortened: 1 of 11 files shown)


craigcondit (approved):
Not entirely thrilled with the "int as tri-state bool" pattern... It just feels ugly. An enum with constants might be cleaner, and easier to follow in the logic. True/False/Maybe. We get type safety that way too.

pkg/admission/admission_controller.go
```diff
@@ -524,11 +526,28 @@ func (c *AdmissionController) namespaceMatchesNoLabelList(namespace string) bool
...
 
+// shouldProcessNamespace returns true if the pod in the namespace must be redirected to the
+// YuniKorn scheduler by setting the schedulerName
+// First check is the namespace annotation (tri-state)
+// - if present (0, 1) return the value as boolean
+// - if not present (-1) fallback to matching names based on regexp
 func (c *AdmissionController) shouldProcessNamespace(namespace string) bool {
+	process := c.nsCache.enableYuniKorn(namespace)
+	if process != -1 {
+		return process == 1
```
pbacsko, line 537: 1 / -1 are special values, could you name them? Difficult to reason about them without names.

pkg/admission/admission_controller.go
```diff
@@ -524,11 +526,28 @@ func (c *AdmissionController) namespaceMatchesNoLabelList(namespace string) bool
...
 
+// shouldLabelNamespace returns true if the pod in the namespace must be labeled
+// First check is the namespace annotation (tri-state)
+// - if present (0, 1) return the value as boolean
+// - if not present (-1) fallback to matching names based on regexp
 func (c *AdmissionController) shouldLabelNamespace(namespace string) bool {
+	label := c.nsCache.generateAppID(namespace)
+	if label != -1 {
+		return label == 1
+	}
```
pbacsko, line 550: Same thing here

pbacsko:
> Not entirely thrilled with the "int as tri-state bool" pattern... It just feels ugly. An enum with constants might be cleaner, and easier to follow in the logic. True/False/Maybe. We get type safety that way too.

+1 on this, naming the values increases readability.

pkg/admission/admission_controller.go
wilfred-s: Follow up jira to move to an iota: [YUNIKORN-1652](https://issues.apache.org/jira/browse/YUNIKORN-1653)

PR #564, opened 2023-03-29
[YUNIKORN-1535] Use scratch image as base in Dockerfiles

craigcondit:
### What is this PR for?
In the interests of security, use a scratch image for Docker images where we deploy a single Go binary.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1535

Existing e2e tests cover this.

### Questions:

### DIFF
--- a/deployments/image/admission/Dockerfile
+++ b/deployments/image/admission/Dockerfile
@@ -14,7 +14,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM scratch
+ARG TARGETPLATFORM
+FROM --platform=$TARGETPLATFORM scratch
 COPY --chown=0:0 scheduler-admission-controller /
 USER 4444:4444
 ENTRYPOINT [ "/scheduler-admission-controller" ]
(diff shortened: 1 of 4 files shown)


deployments/image/admission/Dockerfile
```diff
@@ -14,22 +14,7 @@
...
-
-COPY --chown=4444:0 scheduler-admission-controller $HOME/bin/
-
-WORKDIR $HOME
-USER yunikorn
-ENTRYPOINT $HOME/bin/scheduler-admission-controller
+FROM scratch
+COPY --chown=0:0 scheduler-admission-controller /
+USER 4444:4444
+ENTRYPOINT [ "/scheduler-admission-controller" ]
```
craigcondit, line 20: Implementation details: We use `COPY --chown=0:0` to ensure that the installed binaries are owned by root (as we should not have write access to them). `USER 4444:4444` is used to ensure we run (by default at least) as a non-privileged user. Finally, the entrypoint is given in long form so that no shell is needed in the container. The resulting container image will contain *only* the resources listed in the `COPY`.

wilfred-s:
Change looks good adding extra reviewers so that more people are aware of the move to scratch images.
Let's keep this one open for a couple of days to get feedback.

wilfred-s (approved):
LGTM

craigcondit:
@wilfred-s can you re-review this? I found some minor changes required for multi-arch builds. The updated patch should work on Mac and Linux for all 3 architectures.

PR #565, opened 2023-03-31
[YUNIKORN-1670] Make application recovery more robust

craigcondit:
### What is this PR for?
Remove timeout from application recovery, and ensure that applications are removed from the recovery list as soon as they have transitioned out of New or Recovering status (rather than into Accepted). This prevents failure on large clusters due to timeouts, and also prevents failure in the case of a rejected application.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1670

Existing unit tests cover this.

### Questions:


pkg/appmgmt/appmgmt_recovery.go
```diff
@@ -28,18 +28,18 @@ import (
 	"github.com/apache/yunikorn-k8shim/pkg/appmgmt/general"
 	"github.com/apache/yunikorn-k8shim/pkg/appmgmt/interfaces"
 	"github.com/apache/yunikorn-k8shim/pkg/cache"
-	"github.com/apache/yunikorn-k8shim/pkg/common/utils"
 	"github.com/apache/yunikorn-k8shim/pkg/log"
 )
 
-func (svc *AppManagementService) WaitForRecovery(maxTimeout time.Duration) error {
+func (svc *AppManagementService) WaitForRecovery() error {
```
wilfred-s, line 34: Please document the expected behaviour for the function under normal and test case setup.
It looks like the abort could happen in production but the change explicitly says no timeout for app recovery

pkg/appmgmt/appmgmt_recovery.go
```diff
@@ -71,33 +71,47 @@ func (svc *AppManagementService) recoverApps() (map[string]interfaces.ManagedApp
...
-
+func (svc *AppManagementService) waitForAppRecovery(recoveringApps map[string]interfaces.ManagedApp) bool {
+	svc.cancelRecovery.Store(false) // reset cancellation token
+	recoveryStartTime := time.Now()
+	counter := 0
+	for {
+		// check for cancellation token
+		if svc.cancelRecovery.Load() {
+			log.Logger().Info("Waiting for recovery canceled.")
+			svc.cancelRecovery.Store(false)
```
wilfred-s, line 82: why reset? if recovery is cancelled we should never re-run it and a new recovery sets it to false on line 75

pkg/appmgmt/appmgmt_recovery.go
```diff
@@ -71,33 +71,47 @@ func (svc *AppManagementService) recoverApps() (map[string]interfaces.ManagedApp
...
+			log.Logger().Info("Waiting for application recovery",
+				zap.Duration("timeElapsed", time.Since(recoveryStartTime).Round(time.Second)),
+				zap.Int("appsRemaining", len(recoveringApps)))
+		}
+		time.Sleep(1 * time.Second)
 	}
+}
 
-	return nil
+// cancelWaitForAppRecovery is used by testing code to ensure that waitForAppRecovery() does not block forever
```
wilfred-s, line 101: nit: no () in the comment after the function name

pkg/appmgmt/appmgmt_recovery.go
```diff
@@ -71,33 +71,47 @@ func (svc *AppManagementService) recoverApps() (map[string]interfaces.ManagedApp
...
+func (svc *AppManagementService) cancelWaitForAppRecovery() {
+	svc.cancelRecovery.Store(true)
+}
+
+// removeRecoveredApps() is used to walk the currently recovering apps list and remove those that have finished recovering
```
wilfred-s, line 106: nit: no () in the comment after the function name

wilfred-s (changes requested):
comment updates only, code looks good.

pkg/appmgmt/appmgmt_recovery.go
craigcondit: This is for test code that runs this multiple times to check negative constraints.

craigcondit: Specifically, TestAppStatesDuringRecovery.

pkg/appmgmt/appmgmt_recovery.go
craigcondit: Updated.

pkg/appmgmt/appmgmt_recovery.go
craigcondit: Updated.

pkg/appmgmt/appmgmt_recovery.go
craigcondit: Updated comments. The abort is only used by testing code.

wilfred-s (approved):
LGTM

PR #569, opened 2023-04-05
[YUNIKORN-1676] E2E: Spark test times out

zhuqi-lucas:
### What is this PR for?
Change the spark kill way to the  stable kill all pod way.

### What type of PR is it?
* [x] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:

### DIFF
--- a/test/e2e/spark_jobs_scheduling/spark_jobs_scheduling_test.go
+++ b/test/e2e/spark_jobs_scheduling/spark_jobs_scheduling_test.go
@@ -157,16 +157,9 @@ var _ = Describe("", func() {
 		}
 
 		By("Killing all spark jobs")
-		sparkPods, err := kClient.GetPodNamesFromNS(sparkNS)
+		// delete the Spark pods one by one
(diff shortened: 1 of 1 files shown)


craigcondit (changes requested):
-1. This isn't really a timeout issue, it's a flaky test. The problem seems to be that the "kill spark jobs" step is unreliable. This should take seconds, not minutes. Need to determine why that is failing and fix the underlying issue.

zhuqi-lucas:
Thank you @craigcondit for review,
I also think increse timeout is not a good idea, but it seems hard for us to find the root flaky case.

zhuqi-lucas:
@craigcondit 
Change spark jobs kill to a stable way to delete pod one by one in latest PR.

craigcondit (approved):
+1 LGTM.

PR #570, opened 2023-04-11
[YUNIKORN-1684] Ensure yunikorn-defaults and yunikorn-configs are always honored

craigcondit:
### What is this PR for?
Fixes a race condition in scheduler initialization where one of the ConfigMaps (yunikorn-defaults or yunikorn-configs) may be ignored.

The root cause is a race condition on startup where the initial configmaps are used to configure YuniKorn, but are not cached in the context. Then, depending on which configmap gets added first, the other may be forgotten and ignored during config refresh. Additionally locking was added to the reload code path to ensure that if both configmaps are updated in quick succession, we serialize the updates.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1684

This is difficult to reproduce, but e2e tests should be more reliable. 

### Questions:


pbacsko (approved):
+1

PR #572, opened 2023-04-19
[YUNIKORN-1701] Make simple preemptor cluster agnostic

manirajv06:
### What is this PR for?
Ensure simple preemptor runs seamlessly in all cluster environments.

### What type of PR is it?
* [ ] - Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1701

### Questions:


test/e2e/simple_preemptor/simple_preemptor_test.go
```diff
@@ -79,13 +81,29 @@ var _ = ginkgo.BeforeSuite(func() {
 
 	// Extract node allocatable resources
 	for _, node := range nodes.Items {
-		if node.Name == Worker1 {
+		// skip master if it's marked as such
+		node := node
+		if k8s.IsMasterNode(&node) || !k8s.IsComputeNode(&node) {
```
craigcondit, line 86: Isn't any node which is not a master node by definition a compute node? Why the additional check?

test/e2e/simple_preemptor/simple_preemptor_test.go
```diff
@@ -79,13 +81,29 @@ var _ = ginkgo.BeforeSuite(func() {
...
+			continue
+		}
+		if Worker1 == "" {
```
craigcondit, line 89: This code seems brittle. Can we just iterate the nodes and take the first two matching candidates without all the special-casing?

craigcondit (changes requested):
See comments.

test/e2e/simple_preemptor/simple_preemptor_test.go
manirajv06: I have seen 3 types - 1. Master 2. Workers 2a). Infra 2b). Compute. We have to consider only compute nodes, not even 2a for running pods. Once Compute nodes has been chosen, we have to limit the no. of compute nodes (mostly two) for running pods and taint other compute nodes as un schedulable based on the use cases.

test/e2e/framework/helpers/k8s/k8s_utils.go
```diff
@@ -1292,10 +1292,22 @@ func IsMasterNode(node *v1.Node) bool {
...
+	roleNodeLabelExists := false
+	for labelKey, labelValue := range node.Labels {
+		if labelKey == common.RoleNodeLabel {
+			roleNodeLabelExists = true
+			if _, ok := common.ComputeNodeLabels[labelValue]; ok {
+				return true
+			}
+		}
+	}
+	return !roleNodeLabelExists
```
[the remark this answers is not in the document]
manirajv06: There is no "role" node label in kind clusters setup. "role" node label is available in aks/eks clsusters. IsMasterNode method  is good enough for Kind clusters to separate out the master from worker nodes. IsComputeNode is mainly for separate out the compute worker nodes from infra worker nodes in aks/eks clusters. Just to satisfy kind clusters or any other cluster setup (where there is no "role" node label), we need to return true here (!roleNodeLabelExists).

test/e2e/simple_preemptor/simple_preemptor_test.go
manirajv06: We need to exclude 1. Master 2. Infra worker nodes and choose only "compute" worker nodes. Even no. of "compute" worker nodes (mostly two) would be chosen based on use cases and other remaining "compute" worker nodes would be tainted as "unschedulable" till we complete the test execution.

test/e2e/simple_preemptor/simple_preemptor_suite_test.go
```diff
@@ -36,17 +35,16 @@ func init() {
 
 func TestSimplePreemptor(t *testing.T) {
 	ginkgo.ReportAfterSuite("TestSimplePreemptor", func(report ginkgo.Report) {
-		err := common.CreateJUnitReportDir()
-		Ω(err).NotTo(gomega.HaveOccurred())
-		err = reporters.GenerateJUnitReportWithConfig(
+		err := reporters.GenerateJUnitReportWithConfig(
 			report,
-			filepath.Join(configmanager.YuniKornTestConfig.LogDir, "TEST-simple_preemptor_junit.xml"),
+			filepath.Join(configmanager.YuniKornTestConfig.LogDir, "TEST-simple_pre_emptor_junit.xml"),
```
craigcondit, line 40: Minor nit: Preemptor is one word. WIll fix this on commit.

craigcondit (approved):
+1.

PR #577, opened 2023-04-21
[YUNIKORN-1705] Remove timeout on admission controller informer sync

craigcondit:
### What is this PR for?
The informer sync code in the admission controller has a fixed timeout before failing during initial sync. We should wait until the sync completes regardless of how long it may take and log our progress instead. This makes the AM more resilient to slow clusters, as the underlying informers should recover cleanly.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1705

E2e tests should succeed now that proper permissions have been granted to the admission controller helm charts.

### Questions:

### DIFF
--- a/pkg/admission/informers.go
+++ b/pkg/admission/informers.go
@@ -27,7 +27,6 @@ import (
 	schedulinginformersv1 "k8s.io/client-go/informers/scheduling/v1"
 
 	"github.com/apache/yunikorn-k8shim/pkg/client"
-	"github.com/apache/yunikorn-k8shim/pkg/common/utils"
 	"github.com/apache/yunikorn-k8shim/pkg/log"
 )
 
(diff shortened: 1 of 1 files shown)


zhuqi-lucas (commented):
LGTM, And one question, if we can use an asynchronous way to wait, when it is ready then return.

manirajv06 (approved):
looks good.

Won't there be any performance impact because of this never ending loop (in worst case)?

craigcondit:
> one question, if we can use an asynchronous way to wait, when it is ready then return.
The API does not support async wait, so we need to poll.

PR #578, opened 2023-04-24
[YUNIKORN-1653] use iota instead of defined values for tri-state

wusamzong:
### What is this PR for?
Use iota to define value for tri-state

### What type of PR is it?
* [x] - Improvement

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1653/

### Questions:


craigcondit (approved):
+1 LGTM.

PR #579, opened 2023-04-25
[YUNIKORN-1707] Ignore pods in non-labeled namespaces

craigcondit:
### What is this PR for?
The admission controller allows pods in configured namespaces to be either
bypassed entirely, or passed through to YuniKorn but not labeled for
processing (useful in the plugin deployment to pass through to the embedded
default scheduler). However, if YuniKorn detects an applicationId or Spark
spark-app-selector, it will attempt to schedule that Pod via YuniKorn
regardless. This may not succeed if there is not a valid queue available
which matches placement rules.

To prevent this, the admission controller will now annotate pods which are
in non-labeled namespaces to mark them as ignored by YuniKorn, and the
scheduler will honor this annotation. This ensures that a namespace is
either processed completely by YuniKorn, or not at all

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1707

Unit tests updated.

### Questions:

### DIFF
--- a/pkg/common/utils/utils.go
+++ b/pkg/common/utils/utils.go
@@ -109,6 +109,16 @@ func GetQueueNameFromPod(pod *v1.Pod) string {
 }
 
 func GetApplicationIDFromPod(pod *v1.Pod) (string, error) {
+	// if pod was tagged with ignore-application, return
+	if value := GetPodAnnotationValue(pod, constants.AnnotationIgnoreApplication); value != "" {
+		ignore, err := strconv.ParseBool(value)
+		if err != nil {
+			log.Logger().Warn("Failed to parse annotation "+constants.AnnotationIgnoreApplication, zap.Error(err))
+		} else if ignore {
+			return "", nil
+		}
+	}
+
 	// application ID can be defined in annotations
 	if value := GetPodAnnotationValue(pod, constants.AnnotationApplicationID); value != "" {
 		return value, nil
--- a/pkg/common/utils/utils.go
+++ b/pkg/common/utils/utils.go
@@ -74,7 +74,7 @@ func NeedRecovery(pod *v1.Pod) bool {
 	// 1. Pod is scheduled by us
 	// 2. pod is already assigned to a node
 	// 3. pod is not in terminated state
-	if GeneralPodFilter(pod) && IsAssignedPod(pod) && !IsPodTerminated(pod) {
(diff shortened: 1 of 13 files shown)


zhuqi-lucas (approved):
LGTM Thanks @craigcondit .

pkg/common/utils/utils.go
```diff
@@ -109,6 +109,16 @@ func GetQueueNameFromPod(pod *v1.Pod) string {
...
+	// if pod was tagged with ignore-application, return
+	if value := GetPodAnnotationValue(pod, constants.AnnotationIgnoreApplication); value != "" {
+		ignore, err := strconv.ParseBool(value)
+		if err != nil {
+			log.Logger().Warn("Failed to parse annotation "+constants.AnnotationIgnoreApplication, zap.Error(err))
+		} else if ignore {
+			return "", nil
+		}
+	}
+
```
wilfred-s, line 121: We should add this in the `utils.GeneralPodFilter(obj)`.
Any pod flows through that filter first. It removes the pod from the internal YuniKorn handling. Pods that do not pass that filter do not end up in the cache either (see FilteringResourceEventHandler).

The plugins (Filter, PreFilter and PostBind) should also use the `GeneralPodFilter`. If the general pod filter passes an application ID must always be there.
It gives us a more consistent way of handling pods. It also removes pods from the cache that we want to bypass YuniKorn.

zhuqi-lucas: Good suggestion,  and GeneralPodFilter is ahead of GetApplicationIDFromPod.

// filter pods by scheduler name and state
func (os *Manager) filterPods(obj interface{}) bool {
	switch obj.(type) {
	case *v1.Pod:
		pod := obj.(*v1.Pod)
		if utils.GeneralPodFilter(pod) {
			// only application ID is required
			if _, err := utils.GetApplicationIDFromPod(pod); err == nil {
				return true
			}
		}
		return false
	default:
		return false
	}
}

craigcondit: Merged the two functions in the latest PR.

wilfred-s (approved):
LGTM

PR #584, opened 2023-05-03
[YUNIKORN-1708] Filtered owner references for placeholder pods.

zhuqi-lucas:
### What is this PR for?
In AWS EMR on EKS service, the driver real pod's ownerReference is configmap.
And placeholder's ownerReference is also the driver configmap.

When user cancels emr-containers job, the job-submitter is terminated,
but the placeholder still remains in pending state.
https://docs.aws.amazon.com/emr/latest/EMR-on-EKS-DevelopmentGuide/emr-eks.html

Make sure configmap owner reference is not returned for pod, it will return the real driver pod itself
because when the placeholder pod set ownerreference to configmap, the configmap will not be deleted for placeholder pod
to GC, we need set the placeholder pod's ownerreference to the real driver pod, when the real driver pod is deleted, the
placeholder will be deleted as well.

The placeholders should really only have one owner: the pod that triggered the creation nothing else.

The placeholder manager should just default to generate just that one.

### What type of PR is it?
* [x] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1708

### Questions:

### DIFF
--- a/pkg/appmgmt/general/general.go
+++ b/pkg/appmgmt/general/general.go
@@ -105,15 +105,7 @@ func isStateAwareDisabled(pod *v1.Pod) bool {
 }
 
 func getOwnerReference(pod *v1.Pod) []metav1.OwnerReference {
-	if len(pod.OwnerReferences) > 0 {
-		// Filter the pod kind owner reference
-		// keep this because we need to find the original pod
-		for _, ref := range pod.OwnerReferences {
-			if ref.Kind == reflect.TypeOf(v1.Pod{}).Name() {
-				return []metav1.OwnerReference{ref}
-			}
-		}
-	}
+	// Just return the originator pod as the owner of placeholder pods
 	controller := false
 	blockOwnerDeletion := true
 	ref := metav1.OwnerReference{
--- a/pkg/appmgmt/general/general.go
+++ b/pkg/appmgmt/general/general.go
@@ -19,7 +19,6 @@
 package general
 
 import (
-	"reflect"
 	"strconv"
 
 	v1 "k8s.io/api/core/v1"
@@ -109,8 +108,8 @@ func getOwnerReference(pod *v1.Pod) []metav1.OwnerReference {
 	controller := false
 	blockOwnerDeletion := true
 	ref := metav1.OwnerReference{
-		APIVersion:         v1.SchemeGroupVersion.String(),
-		Kind:               reflect.TypeOf(v1.Pod{}).Name(),
+		APIVersion:         "v1",
+		Kind:               "Pod",
 		Name:               pod.Name,
 		UID:                pod.UID,
 		Controller:         &controller,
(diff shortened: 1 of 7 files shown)


wilfred-s:
See comments in the jira, different approach is needed

zhuqi-lucas:
Thank you @wilfred-s , i will check it!

pkg/appmgmt/general/general.go
```diff
@@ -104,9 +104,15 @@ func isStateAwareDisabled(pod *v1.Pod) bool {
...
+func getOwnerReference(pod *v1.Pod) []metav1.OwnerReference {
 	if len(pod.OwnerReferences) > 0 {
-		return pod.OwnerReferences
+		// Filter the pod kind owner reference
+		// keep this because we need to find the original pod
+		for _, ref := range pod.OwnerReferences {
+			if ref.Kind == reflect.TypeOf(v1.Pod{}).Name() {
+				return []metav1.OwnerReference{ref}
+			}
+		}
```
wilfred-s, line 115: This is not needed, the call should only ever be made with the originator or owner pod.
Just always build a new reference without doing anything else.

NIT simplification:
`v1.SchemeGroupVersion.String()` is always "v1" unless we change the signature. 
`reflect.TypeOf(v1.Pod{}).Name()` is always "Pod" nothing else.

pkg/appmgmt/general/general_test.go
```diff
@@ -167,6 +167,7 @@ func TestOriginatorPod(t *testing.T) {
 	owner := apis.OwnerReference{
 		APIVersion: "v1",
 		UID:        "UID-POD-00002",
+		Kind:       "Pod",
 	}
 	refer := []apis.OwnerReference{
 		owner,
```
wilfred-s, line 173: use getOwnerReference to build this

wilfred-s (changes requested):
Two point left and please look at the e2e test failure

pkg/appmgmt/general/general.go
zhuqi-lucas: Thank you @wilfred-s for review, addesseed in latest PR!

pkg/appmgmt/general/general_test.go
```diff
@@ -208,10 +210,12 @@ func TestOriginatorPod(t *testing.T) {
 	}
 	am.AddPod(&pod1)
 	assert.Equal(t, len(app.GetNewTasks()), 2)
-	task, err = app.GetTask("UID-POD-00002")
+	task, err = app.GetTask("UID-POD-00001")
```
zhuqi-lucas, line 213: Hi @wilfred-s , if we always return the pod, the test here the originator will be the pod1 who first add into cluster.

wilfred-s (approved):
LGTM

PR #586, opened 2023-05-05
[YUNIKORN-1723] Remove string comparison from hasReadyCondition() function

pbacsko:
### What is this PR for?
Remove the comparison `NodeStatus.String()` against `nil`. It's unnecessary and expensive.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1723

### Questions:

### DIFF
--- a/pkg/cache/nodes.go
+++ b/pkg/cache/nodes.go
@@ -217,7 +217,7 @@ func (nc *schedulerNodes) schedulerNodeEventHandler() func(obj interface{}) {
 }
 
 func hasReadyCondition(node *v1.Node) bool {
-	if node != nil && node.Status.String() != "nil" {
+	if node != nil {
 		for _, condition := range node.Status.Conditions {
 			if condition.Type == v1.NodeReady && condition.Status == v1.ConditionTrue {
 				return true


zhuqi-lucas (approved):
Good catch, LGTM +1

pkg/cache/nodes.go
```diff
@@ -217,7 +217,7 @@ func (nc *schedulerNodes) schedulerNodeEventHandler() func(obj interface{}) {
 }
 
 func hasReadyCondition(node *v1.Node) bool {
-	if node != nil && node.Status.String() != "nil" {
+	if node != nil {
```
yangwwei, line 220: can node.Status be a nil? do we need
```
node.Status != nil?
```

wilfred-s: No it cannot be nil as it is not a pointer it is a NodeStatus struct.

wilfred-s (approved):
LGTM
@manirajv06 please check

manirajv06:
looks good.

PR #587, opened 2023-05-05
[YUNIKORN-1724] Improve the performance of shim side scheduling cycle

pbacsko:
### What is this PR for?
Improve the retrieval of `New` tasks by dynamically updating a separate map. With thousands of schedulable pods, scanning the `taskMap` with retrieving the state yields severe degradation in performance.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1724

### Questions:

### DIFF
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/google/btree"
 	"github.com/looplab/fsm"
 	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pbacsko:
A simple testcase to see the differences:

```
func TestSchedule(t *testing.T) {
	context := initContextForTest()
	dispatcher.RegisterEventHandler(dispatcher.EventTypeTask, context.TaskEventHandler())
	dispatcher.Start()

	app := NewApplication(appID, "root.a", "testuser", testGroups, map[string]string{}, newMockSchedulerAPI())
	context.addApplication(app)

	app.sm.SetState("Running")
	for i := 0; i < 20000; i++ {
		task := NewTask("task"+strconv.Itoa(i), app, context, &v1.Pod{})
		app.addTask(task)
		app.Schedule()
	}
}
```

pkg/cache/application.go
```diff
@@ -396,7 +438,7 @@ func (app *Application) Schedule() bool {
 }
 
 func (app *Application) scheduleTasks(taskScheduleCondition func(t *Task) bool) {
-	for _, task := range app.GetNewTasks() {
+	for _, task := range app.newTasks.getTasks() {
```
pbacsko, line 441: There is one thing that the patch doesn't take into account: the order of pods based on submission time. I'll update it on Monday.

pbacsko: Ok, I couldn't wait so I updated it already.

pkg/cache/application.go
```diff
@@ -62,6 +63,68 @@ type Application struct {
...
+
+	newTasks *newTasksTracker
+}
+
+// For performance reasons, we track tasks with "New" state for quick retrieval.
+// Application.GetNewTasks (and getTasks() in general) must not be used on critical paths because it
+// uses a read lock inside fsm.FSM which becomes a bottleneck with large number of pods.
+type newTasksTracker struct {
+	sortedTasks *btree.BTree
+	sync.Mutex
```
craigcondit, line 75: Shouldn't we use an RWMutex here as reads are likely to (significantly) outstrip writes?

pkg/cache/application_test.go
```diff
@@ -1343,3 +1343,64 @@ func (ctx *Context) addApplication(app *Application) {
...
+	}
+	verifyNewTaskOrding(t, app)
+	app.removeTask("task04")
+	app.removeTask("task05")
+	verifyNewTaskOrding(t, app)
+	assert.Equal(t, 3, app.newTasks.count())
+}
+
+func verifyNewTaskOrding(t *testing.T, app *Application) {
+	newTasks := app.GetNewTasks()
```
craigcondit, line 1400: I think app.GetNewTasks() should be refactored to use the new method. There's no point in leaving the old mechanism around. We can add an `app.HasNewTasks() bool` method as well for quickly determining if the list is empty.

craigcondit (changes requested):
Minor changes (mostly locking), though I think we should refactor out the old method, as it seems to only be used in tests.

pbacsko:
Had a conversation w/ @wilfred-s about this, the better solution is to simply remove the scheduling cycle. By so doing, we can get rid of some states in the Application and simplify the code. This PR is not needed, the issue is not relevant with the default scheduling cycle (1s).

PR #589, opened 2023-05-11
[YUNIKORN-1732] Fix Makefile calculation of SHA hashes

craigcondit:
### What is this PR for?
Fix the Makefile's calculation of SHA hashes for build components. The scripts used are brittle and fail if the output of `go list -m` doesn't fit the expected pattern (especially when replacing local dependencies).

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1732

Build now succeeds where it failed previously and generates proper hashes.

### Questions:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,8 +86,8 @@ DOCKER_ARCH := amd64
 endif
 
 # Image hashes
-CORE_SHA=$(shell go list -m "github.com/apache/yunikorn-core" | cut -d " " -f2 | cut -d "-" -f3)
-SI_SHA=$(shell go list -m "github.com/apache/yunikorn-scheduler-interface" | cut -d " " -f2 | cut -d "-" -f3)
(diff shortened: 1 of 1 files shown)


wilfred-s (changes requested):
Proposal is to use this, removes all the field dependencies:
```
go list -f '{{.Version}}' -m "github.com/apache/yunikorn-core" | cut -d "-" -f3`
```

craigcondit:
> Proposal is to use this, removes all the field dependencies:
> 
> ```
> go list -f '{{.Version}}' -m "github.com/apache/yunikorn-core" | cut -d "-" -f3`
> ```

Nice catch, have updated.

sunilgovind (approved):
lgtm

PR #593, opened 2023-05-19
[YUNIKORN-1750] remove sprintf from logging calls

FrankYang0529:
### What is this PR for?

Don't use `fmt.Sprintf` in zap logging.

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1750

1. Create the following ConfigMap:

```yaml
apiVersion: v1
data:
  log.level: "not int"
kind: ConfigMap
metadata:
  name: yunikorn-defaults
  namespace: yunikorn
```

2. Run admission webhook.
3. We should see the following error message:

```
Unable to parse int value, using default    {"key": "log.level", "value": "not int", "default": 0, "error": "strconv.ParseInt: parsing \"not int\": invalid syntax"}
```

### Questions:

### DIFF
--- a/pkg/admission/conf/am_conf.go
+++ b/pkg/admission/conf/am_conf.go
@@ -351,8 +351,8 @@ func parseConfigRegexps(config map[string]string, key string, defaultValue strin
 	value := parseConfigString(config, key, defaultValue)
 	result, err := parseRegexes(value)
 	if err != nil {
-		log.Logger().Error(fmt.Sprintf("Unable to parse regex values '%s' for configuration '%s', using default value '%s'",
-			value, key, defaultValue), zap.Error(err))
+		log.Logger().Error("Unable to parse regex values, using default",
+			zap.String("key", key), zap.String("value", value), zap.String("default", defaultValue), zap.Error(err))
 		result, err = parseRegexes(defaultValue)
 		if err != nil {
 			log.Logger().Fatal("BUG: can't parse default regex pattern", zap.Error(err))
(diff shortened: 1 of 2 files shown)


pkg/admission/conf/am_conf.go
```diff
@@ -351,8 +351,8 @@ func parseConfigRegexps(config map[string]string, key string, defaultValue strin
 	value := parseConfigString(config, key, defaultValue)
 	result, err := parseRegexes(value)
 	if err != nil {
-		log.Logger().Error(fmt.Sprintf("Unable to parse regex values '%s' for configuration '%s', using default value '%s'",
-			value, key, defaultValue), zap.Error(err))
+		log.Logger().Sugar().With(zap.Error(err)).Errorf("Unable to parse regex values '%s' for configuration '%s', using default value '%s'",
```
pbacsko, line 354: don't use sugar, it's avoided in the project, do this instead:
```
log.Logger.Error("Unable to parse regex values, using default", zap.String("key", key), zap.String("value", value), zap.String("default", default))
```

pkg/admission/conf/am_conf.go
```diff
@@ -365,8 +365,8 @@ func parseConfigBool(config map[string]string, key string, defaultValue bool) bo
 	value := parseConfigString(config, key, fmt.Sprintf("%t", defaultValue))
 	result, err := strconv.ParseBool(value)
 	if err != nil {
-		log.Logger().Error(fmt.Sprintf("Unable to parse bool value '%s' for configuration '%s', using default value '%t'",
-			value, key, defaultValue), zap.Error(err))
+		log.Logger().Sugar().With(zap.Error(err)).Errorf("Unable to parse bool value '%s' for configuration '%s', using default value '%t'",
```
pbacsko, line 368: Same as above

pbacsko (changes requested):
Just use normal `Error()` calls without sugared logger.

pkg/admission/conf/am_conf.go
FrankYang0529: Updated it. I still keep `zap.Error(err)`. May I know why we don't use sugar? Is it cause of performance? Thank you.

pbacsko: `zap.Error()` is fine. I remember inside Yunikorn, we avoid it for performance reasons. It probably wouldn't make a difference inside the Admission Controller, but at least be consistent with the existing code.

PR #595, opened 2023-05-20
[YUNIKORN-1746] Improve the performance of nodeInfoListerImpl.List()

pbacsko:
### What is this PR for?
The method `nodeInfoListerImpl.List()` is called very frequently and always creates a slice with the same content over and over again. It's only necessary to update if a nodes joins/leaves the cluster.
The code is also on a critical path, optimizing it has visible performance benefits.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1746

### Questions:

### DIFF
--- a/pkg/cache/external/scheduler_cache.go
+++ b/pkg/cache/external/scheduler_cache.go
@@ -56,10 +56,8 @@ type SchedulerCache struct {
 	pendingAllocations    map[string]string // map of pod to node ID, presence indicates a pending allocation for scheduler
 	inProgressAllocations map[string]string // map of pod to node ID, presence indicates an in-process allocation for scheduler
 	lock                  sync.RWMutex
-	clients               *client.Clients // client APIs
-
-	// for caching
-	nodesMapGeneration int64
+	clients               *client.Clients       // client APIs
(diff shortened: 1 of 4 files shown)


craigcondit:
Had a conversation offline about this. I'd like to see the caching logic encapsulated completely within the SchedulerCache object. We can add a GetNodeList() method that returns the list, and cache the list inside the SchedulerCache itself. If the pointer to that list is nil, then assume it is invalid. Then we compute it on demand in GetNodeList(), and set to nil any time the node map is modified. Finally, update the NodeLister to call GetNodeList() instead of the GetNodesInfoMap().

craigcondit (changes requested):
See comment above.

craigcondit (approved):
+1 LGTM. On merge, I'm going to add a short comment to GetNodesInfo to reflect the locking semantics required (same as GetNodesInfoMap).

PR #597, opened 2023-05-23
[YUNIKORN-1650] Add toggle for adding default queue in admission controller

mitdesai:
### What is this PR for?
This PR add a configuration to skip adding default queue name label to the pods.

This option is needed when complex placement rules are enabled. Especially when _provided_ rule is used in conjunction with other placement rules and provided rule is higher in the hierarchy. If default queue label is always getting added for pods that do not have a queue label, all the apps will be scheduled via _provided_ rule and the other rules after that will never be executed. 

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1650

By default, the configuration is turned off
1. Out of the box, no changes are required if you do not intent to change the current behavior.
2. For skipping to add the default queue name, add this configuration in the admission controller configs: `admissionController.label.defaultQueue=""`. The queue name label should only get added to the pod if a value for queue name was specified during the app submission
3.  For changing the default queue name to something other than `root.default`, add this configuration to the admission controller configs: `admissionController.label.defaultQueue="root.mydefault"`. Note: The value for the default queue has to be a fully qualified queue name.

### Questions:
* [X] - It needs documentation.


pkg/admission/conf/am_conf.go
```diff
@@ -60,6 +61,9 @@ const (
 	AMAccessControlSystemUsers      = AccessControlPrefix + "systemUsers"
 	AMAccessControlExternalUsers    = AccessControlPrefix + "externalUsers"
 	AMAccessControlExternalGroups   = AccessControlPrefix + "externalGroups"
+
+	// label configurations
+	AMLabelDefaultQueue = LabelPrefix + "defaultQueue"
```
craigcondit, line 66: Minor nit: Let's put these under filtering (ala `generateUniqueAppId`).

craigcondit (changes requested):
This looks pretty reasonable. I'm in favor, other than the specific key used for configuration.

pkg/admission/conf/am_conf.go
mitdesai: Thanks for the review @craigcondit.
I think that looks better. Let me update the PR.

mitdesai: @craigcondit, I've updated the PR. I'll resolve this conversation.

craigcondit (approved):
+1 LGTM.

mitdesai:
Thanks @craigcondit !

PR #598, opened 2023-05-24
[YUNIKORN-1555] Deduplicate async pod events during recovery

bgrams:
### What is this PR for?
Follow up to https://github.com/apache/yunikorn-k8shim/pull/532. Original PR failed to address duplicated events coming from the informer via `asyncEvent`s as this scenario was not covered by unit tests.

Changes involve passing state between recovery methods to deduplicate `AddPod` operations given the pod UID.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [x] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
[YUNIKORN-1555](https://issues.apache.org/jira/browse/YUNIKORN-1555)

Unit tests modified

### Questions:

### DIFF
--- a/pkg/appmgmt/appmgmt_recovery.go
+++ b/pkg/appmgmt/appmgmt_recovery.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/apache/yunikorn-k8shim/pkg/appmgmt/general"
 	"github.com/apache/yunikorn-k8shim/pkg/appmgmt/interfaces"
@@ -63,14 +64,18 @@ func (svc *AppManagementService) recoverApps() (map[string]interfaces.ManagedApp
 				return pods[i].CreationTimestamp.Unix() < pods[j].CreationTimestamp.Unix()
 			})
 
+			// Track pods that we have already seen in order to
+			// skip redundant handling of async events in RecoveryDone
+			seenEvents := make(map[types.UID]string)
 			for _, pod := range pods {
 				if utils.NeedRecovery(pod) {
 					app := svc.podEventHandler.HandleEvent(general.AddPod, general.Recovery, pod)
 					recoveringApps[app.GetApplicationID()] = app
 				}
+				seenEvents[pod.UID] = pod.GetResourceVersion()
 			}
 			log.Logger().Info("Recovery finished")
-			svc.podEventHandler.RecoveryDone()
+			svc.podEventHandler.RecoveryDone(seenEvents)
 		}
 	}
 
--- a/pkg/appmgmt/appmgmt_recovery.go
+++ b/pkg/appmgmt/appmgmt_recovery.go
@@ -23,14 +23,12 @@ import (
 	"sort"
 	"time"
 
-	"go.uber.org/zap"
-	"k8s.io/apimachinery/pkg/types"
-
 	"github.com/apache/yunikorn-k8shim/pkg/appmgmt/general"
 	"github.com/apache/yunikorn-k8shim/pkg/appmgmt/interfaces"
 	"github.com/apache/yunikorn-k8shim/pkg/cache"
 	"github.com/apache/yunikorn-k8shim/pkg/common/utils"
 	"github.com/apache/yunikorn-k8shim/pkg/log"
+	"go.uber.org/zap"
 )
 
 // WaitForRecovery initiates and waits for the app management service to finish recovery. If recovery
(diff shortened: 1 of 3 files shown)


pkg/appmgmt/general/podevent_handler_test.go
```diff
@@ -73,21 +73,48 @@ func TestHandleAsyncEventWhenNotRecovering(t *testing.T) {
...
 	task, err := app.GetTask("pod1")
 	assert.NilError(t, err)
 	assert.Equal(t, cache.TaskStates().Completed, task.GetTaskState())
+
+	_, err = app.GetTask("pod2")
+	assert.ErrorContains(t, err, "task pod2 doesn't exist in application")
+
+	task, err = app.GetTask("pod3")
+	assert.NilError(t, err)
+	assert.Equal(t, "1", task.GetTaskPod().GetResourceVersion())
```
bgrams, line 107: I'd like to hear feedback on this assertion specifically, as the `UpdatePod` event simply completes the task without actually updating the associated pod. So the tests are passing but this is not extremely intuitive.

wilfred-s: I think the assumption in the code that a UpdatePod event should transition the pod to finished is not correct. In the handling part of `Manager.updatePod` a check of the current pod and new pod is performed. Then it only transitions the pod if needed by calling the handler.

With VPA (vertical pod auto scaler) there might be more that we need to handle for updates, that is a discussion for later.

I think `PodEventHandler.updatePod` needs to get both pods and the logic of the `Manager.updatePod` needs to move to it. That needs a bit of "surgery" of the PodEventHandler, **BUT** not for as part of this change.

wilfred-s (changes requested):

pkg/appmgmt/general/podevent_handler_test.go
pbacsko: Created an umbrella (https://issues.apache.org/jira/browse/YUNIKORN-1765) and a subtask (https://issues.apache.org/jira/browse/YUNIKORN-1766) for this.

pkg/appmgmt/general/podevent_handler.go
```diff
@@ -99,6 +100,14 @@ func (p *PodEventHandler) RecoveryDone() {
 		log.Logger().Info("Processing async events that arrived during recovery",
 			zap.Int("no. of events", noOfEvents))
 		for _, event := range p.asyncEvents {
+			// Adds get special treatment because most of these events will have
+			// been processed in the initial app recovery routine
+			// If the pod resource version of the seen event matches that of the async event,
+			// this indicates that it is a duplicate object and should be skipped
+			seenVersion, ok := seenEvents[event.pod.UID]
+			if ok && seenVersion == event.pod.GetResourceVersion() && event.eventType == AddPod {
```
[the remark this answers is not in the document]
bgrams: I think the answer is no. However I chose not to implement the same check for updates/deletes given that this change is intended simply to delegate logic to the outer `recoverApps` method, which only covers adds. We can extend this to updates if need be.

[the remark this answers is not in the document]
bgrams: to @pbacsko's point it might just be simpler to ditch resource versions for that reason

pkg/appmgmt/appmgmt_recovery.go
```diff
@@ -63,14 +64,18 @@ func (svc *AppManagementService) recoverApps() (map[string]interfaces.ManagedApp
 				return pods[i].CreationTimestamp.Unix() < pods[j].CreationTimestamp.Unix()
 			})
 
+			// Track pods that we have already seen in order to
+			// skip redundant handling of async events in RecoveryDone
+			seenEvents := make(map[types.UID]string)
 			for _, pod := range pods {
 				if utils.NeedRecovery(pod) {
```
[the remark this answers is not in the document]
bgrams: I agree this makes more sense for the scope of this change. Updated.

pkg/appmgmt/appmgmt_recovery.go
```diff
@@ -63,14 +63,18 @@ func (svc *AppManagementService) recoverApps() (map[string]interfaces.ManagedApp
...
 
+			// Track pods that we have already seen in order to
+			// skip redundant handling of async events in RecoveryDone
+			seenPods := make(map[string]bool)
 			for _, pod := range pods {
 				if utils.NeedRecovery(pod) {
 					app := svc.podEventHandler.HandleEvent(general.AddPod, general.Recovery, pod)
 					recoveringApps[app.GetApplicationID()] = app
 				}
+				seenPods[string(pod.UID)] = true
```
pbacsko, line 74: This is updated outside of the "if" branch. Which might contain all kinds of pods not scheduled by Yunikorn.
We need to be inside the branch, plus, only store pods that are not running.

If a pod is running, you don't know whether we have a new event for it or not. You can't filter out events which are about an active pod.

pbacsko (changes requested):
As I mentioned, you need to store finished pods. The `NeedRecovery()` condition has to be split up further.

pbacsko (approved):
+1 LGTM (pending tests)

cc @wilfred-s to get another review

craigcondit (approved):
Looks ok to me. I'd still like @wilfred-s to give the final sign-off as I think he has the best overall view of what's required here.

wilfred-s (approved):
LGTM

PR #602, opened 2023-05-30
[YUNIKORN-1772] remove Context.SelectApplications

FrankYang0529:
### What is this PR for?

* Remove cache Context.SelectApplications, because we only use it with `nil` filter in production code.
* For testing code, we can use cache Context.GetApplication to replace it.

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1772

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -925,15 +925,12 @@ func (ctx *Context) getTask(appID string, taskID string) *Task {
 	return task
 }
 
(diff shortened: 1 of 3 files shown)


craigcondit (approved):
+1 LGTM.

PR #605, opened 2023-05-31
[YUNIKORN-1744] Enable recovery for MockScheduler

pbacsko:
### What is this PR for?
Remove `IsTestingMode()` calls to make recovery possible when using `MockScheduler`.
Add methods so we can simulate existing nodes and pods.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:

### DIFF
--- a/pkg/appmgmt/appmgmt.go
+++ b/pkg/appmgmt/appmgmt.go
@@ -56,21 +56,14 @@ func NewAMService(amProtocol interfaces.ApplicationManagementProtocol,
 	}
 
 	log.Logger().Info("Initializing new AppMgmt service")
-
-	if !apiProvider.IsTestingMode() {
-		log.Logger().Info("Registering Spark operator with the AppMgmt service")
-		appManager.register(
-			// registered app plugins
-			// for general apps
-			general.NewManager(apiProvider, podEventHandler),
-			// for spark operator - SparkApplication
-			sparkoperator.NewManager(amProtocol, apiProvider),
-			// for application crds
-			application.NewAppManager(amProtocol, apiProvider))
-	} else {
-		podEventHandler = general.NewPodEventHandler(amProtocol, false)
-		appManager.register(general.NewManager(apiProvider, podEventHandler))
-	}
+	appManager.register(
+		// registered app plugins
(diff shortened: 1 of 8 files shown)


pkg/cache/context_recovery.go
```diff
@@ -80,86 +73,77 @@ func (ctx *Context) recover(mgr []interfaces.Recoverable, due time.Duration) err
...
-	// we simply simulate to accept or reject nodes on conditions.
-	if !ctx.apiProvider.IsTestingMode() {
-		var podList *corev1.PodList
-		podList, err = ctx.apiProvider.GetAPIs().KubeClient.GetClientSet().
-			CoreV1().Pods("").
-			List(context.Background(), metav1.ListOptions{})
-		if err != nil {
-			return err
-		}
+	pods, err := ctx.apiProvider.GetAPIs().PodInformer.Lister().List(labels.Everything())
```
pbacsko, line 76: I rewrote this to `PodInformer.Lister().List()`, documentation says it's faster.
Although at this point, we already have the list of pods from a previous listing, so we can even skip this (perhaps in a separate JIRA?).

pkg/appmgmt/appmgmt.go
```diff
@@ -56,21 +56,14 @@ func NewAMService(amProtocol interfaces.ApplicationManagementProtocol,
 	}
 
 	log.Logger().Info("Initializing new AppMgmt service")
-
-	if !apiProvider.IsTestingMode() {
-		log.Logger().Info("Registering Spark operator with the AppMgmt service")
```
zhuqi-lucas, line 61: log.Logger().Info("Registering Spark operator with the AppMgmt service")
If this log don't need for new code?

pbacsko: That line does not add too much information, we always register the Spark operator.

craigcondit (approved):
+1 LGTM.

PR #610, opened 2023-06-07
[YUNIKORN-1699] Build against Kubernetes 1.27.2

craigcondit:
Now builds against Kubernetes 1.27.2. This adds runtime suport for Kubernetes 1.27 but removes support for Kubernetes 1.21-1.23. Supported versions are now 1.24.x through 1.27.x. Updated e2e test scripts to match.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1699

### Questions:


craigcondit:
Rebased on master to fix go.mod / go.sum conflicts.

pkg/cache/external/scheduler_cache.go
```diff
@@ -589,6 +596,55 @@ func (cache *SchedulerCache) nodePodCount() int {
...
+}
+
+func (cache *SchedulerCache) updatePVCRefCounts() {
+	maxGeneration := cache.maxNodeGeneration()
+	updateAll := false
+	if cache.pvcGeneration == 0 {
+		// clear cache
+		updateAll = true
+		cache.pvcRefCounts = make(map[string]map[string]int)
+	} else if cache.pvcGeneration >= maxGeneration {
```
wilfred-s, line 623: This is expensive when we have a long list of nodes. Generation for the nodes is a monotonically increasing number and never can go down. We should handle this when the node in the cache is manipulated (new, add or remove pod) they all trigger a generation increment. Use a follow up jira to make this smarter, we have a handle on the node already that has changed and can target this as far as I can tell.

scripts/generate-groups.sh
```diff
@@ -61,7 +61,7 @@ shift 4
   # To support running this script from anywhere, we have to first cd into this directory
   # so we can install the tools.
   cd "$(dirname "${0}")"
-  go install k8s.io/code-generator/cmd/{defaulter-gen,client-gen,lister-gen,informer-gen,deepcopy-gen}
+  go install k8s.io/code-generator/cmd/{defaulter-gen,client-gen,lister-gen,informer-gen,deepcopy-gen}@v0.27.1
```
wilfred-s, line 64: nit: Should this be 1.27.2 ? same as the go.mod file can be fixed on commit if easier

wilfred-s (approved):
LGTM, one follow up for performance analysis

pkg/cache/external/scheduler_cache.go
craigcondit: This was based on the implementation the default scheduler uses. I kept it the same for now to ensure compatibility. We can revisit if this proves to be very expensive.

scripts/generate-groups.sh
craigcondit: Good catch. Original patch used 1.27.1 versions, but output is identical. I've updated locally, and will do so on commit as well.

PR #611, opened 2023-06-08
[YUNIKORN-1797] K8Shim: Implement scoped logging

craigcondit:
Add API support for scoped logging to the Kubernetes shim and admission controller. This will allow for fine-grained logging control per-subsystem.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1797

All shim log messages should now go to the "k8shim" logger, and admission logs to the "admission" logger. Future PRs will move individual subsystems to even more specific loggers.

### Questions:

### DIFF
--- a/pkg/log/logger.go
+++ b/pkg/log/logger.go
@@ -45,18 +45,21 @@ func (h LoggerHandle) String() string {
 
 // Logger constants for configuration
 const (
+	nullLogger  = ""
 	defaultLog  = "log.level"
 	logPrefix   = "log."
 	levelSuffix = ".level"
 )
 
 // Predefined loggers: when adding new loggers, ids must be sequential, and all must be added to the loggers slice in the same order
+var (
+	K8Shim     = &LoggerHandle{id: 1, name: "k8shim"}
+	Kubernetes = &LoggerHandle{id: 2, name: "kubernetes"}
+	Admission  = &LoggerHandle{id: 3, name: "admission"}
+	Test       = &LoggerHandle{id: 4, name: "test"}
+)
 
-var K8Shim = &LoggerHandle{id: 1, name: "k8shim"}
-var Kubernetes = &LoggerHandle{id: 2, name: "kubernetes"}
-var Admission = &LoggerHandle{id: 3, name: "admission"}
-var Test = &LoggerHandle{id: 4, name: "test"}
-
+// this tracks all the known logger handles, used to preallocate the real logger instances when configuration changes
 var loggers = []*LoggerHandle{
 	K8Shim,
 	Kubernetes,
@@ -64,26 +67,31 @@ var loggers = []*LoggerHandle{
 	Test,
 }
 
+// structure to hold all current logger configuration state
 type loggerConfig struct {
 	loggers []*zap.Logger
 }
 
+// tracks the currently used set of loggers; replaced completely whenever configuration changes
 var currentLoggerConfig = atomic.Pointer[loggerConfig]{}
+
+// tracks the default logger handle, which is used for legacy log.Logger() calls
 var defaultLogger = atomic.Pointer[LoggerHandle]{}
 
-// Logger retrieves the global logger
+// Logger retrieves the global logger. This is for compatibility with legacy code and
+// should not be used for new log messages; use Log(loggerHandle) instead
 func Logger() *zap.Logger {
 	once.Do(initLogger)
 	return Log(defaultLogger.Load())
 }
 
-// RootLogger retrieves the root logger
+// RootLogger retrieves the root logger, used to pass the configured logger to the scheduler core on startup
 func RootLogger() *zap.Logger {
 	once.Do(initLogger)
 	return logger
 }
 
-// Log retrieves a standard logger
+// Log retrieves a named logger
 func Log(handle *LoggerHandle) *zap.Logger {
 	once.Do(initLogger)
 	if handle == nil || handle.id == 0 {
@@ -93,6 +101,8 @@ func Log(handle *LoggerHandle) *zap.Logger {
 	return conf.loggers[handle.id-1]
 }
 
+// createLogger creates a new, named logger that has log levels filtered based on the given configuration.
+// levelMap contains a mapping of fully-qualified logger names to log levels
 func createLogger(levelMap map[string]zapcore.Level, name string) *zap.Logger {
 	level := loggerLevel(levelMap, name)
 	return logger.Named(name).WithOptions(zap.WrapCore(func(inner zapcore.Core) zapcore.Core {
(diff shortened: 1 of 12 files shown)


pkg/log/logger.go
```diff
@@ -30,13 +33,99 @@ var once sync.Once
...
+const (
+	defaultLog  = "log.level"
+	logPrefix   = "log."
+	levelSuffix = ".level"
+)
+
+var K8Shim = LoggerHandle{name: "k8shim"}
+var Kubernetes = LoggerHandle{name: "kubernetes"}
+var Admission = LoggerHandle{name: "admission"}
+var Test = LoggerHandle{name: "test"}
```
pbacsko, line 55: If we manage all loggers, we can get faster logger retrieval by defining an extra "int" field in `LoggerHandle` which tells the index of the `*zap.Logger` instance inside a slice. We can still use the map for less commonly used loggers.

craigcondit: This doesn't work because the core and shim loggers are managed separately. The performance impact of these changes is already negligible.

pkg/log/logger_test.go
```diff
@@ -0,0 +1,202 @@
...
+	"go.uber.org/zap/zapcore"
+	"gotest.tools/v3/assert"
+)
+
+var logDir string
+var logFile string
+
+var iterations = 100000
+
+func TestLegacyLoggerPerformance(t *testing.T) {
```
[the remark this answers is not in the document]
craigcondit: The code also functions as a way to exercise these code paths for coverage. I could probably do some benchmarks and factor out common code, but I'd like to keep some variant of this on the default test runs. We also don't (currently) have a benchmark makefile target, so that would need to be added as well.

pbacsko:
Looks good, I'll examine this in my IDE in more detail tomorrow just to have a slightly better understanding.

pkg/log/logger.go
craigcondit: I tried this locally and it does indeed seem to be faster. Overhead is now in the ballpark of a few ns per log call. The one downside is that there's now some more maintenance required for logging to work properly. I've added a unit test to verify that all the loggers exist and are configured as expected.

pkg/log/logger.go
```diff
@@ -82,3 +171,87 @@ func GetZapConfigs() *zap.Config {
...
+// This uses config keys of the form log.{logger}.level={level}.
+// The default level is set by log.level={level}
+func UpdateLoggingConfig(config map[string]string) {
+	once.Do(initLogger)
+	initLoggingConfig(config)
+}
+
+func initLoggingConfig(config map[string]string) {
+	levelMap := make(map[string]zapcore.Level)
+	levelMap[""] = zapcore.InfoLevel
```
pbacsko, line 191: Can we have a constant for `""`, it's not entirely clear what it represents. I guess it's a default level, but a naming would help.

pkg/log/logger.go
```diff
@@ -30,12 +33,95 @@ var once sync.Once
...
+}
+
+func createLogger(levelMap map[string]zapcore.Level, name string) *zap.Logger {
+	level := loggerLevel(levelMap, name)
+	return logger.Named(name).WithOptions(zap.WrapCore(func(inner zapcore.Core) zapcore.Core {
+		return filteredCore{inner: inner, level: level}
+	}))
+}
+
+func loggerLevel(levelMap map[string]zapcore.Level, name string) zapcore.Level {
```
pbacsko, line 103: Can we have some explanation or comments about what happens here? Just by looking at this, I have to spend some time to understand how we retrieve the level.

pkg/log/logger_test.go
```diff
@@ -0,0 +1,244 @@
...
+	nsOp := benchmarkScopedLoggerInfoFiltered(iterations)
+	Log(Test).Info("log.Log(...) performance (root=DEBUG)", zap.Int64("info (ns/op)", nsOp))
+}
+
+func benchmarkScopedLoggerInfoFiltered(iterations int) int64 {
+	_ = Logger()
+	initTestLogger()
+	defer resetTestLogger()
+	UpdateLoggingConfig(map[string]string{
+		"log.test.level": "DEBUG",
```
pbacsko, line 174: What's exactly the difference between this and the non-filtered test? What are we filtering here?

pkg/log/logger_test.go
```diff
@@ -0,0 +1,244 @@
...
+	})
+	start := time.Now()
+	for i := 0; i < iterations; i++ {
+		Log(K8Shim).Info("test", zap.String("foo", "bar"))
```
pbacsko, line 178: I suppose the message "test" is unrelated to logger named "test".

pbacsko (changes requested):
Ok, I did a re-review. Some parts are not entirely clear, it's difficult to understand what happens without comments.
You can also consider using `Example*` functions which documents how the new logger approach should be used (filtering, parent loggers, etc).

pkg/log/logger_test.go
craigcondit: Because the log level is implemented as a pre-check before we get to the root (real) logger, there's two scenarios:

1) Log level matches at both levels (i.e. root is INFO, specific logger is also INFO)
2) Log level does not match (i.e. root is DEBUG, specific logger is INFO)

In the second case this is measuring the performance of logging at DEBUG level when the logger is set to INFO, so the filtering has to happen in our code vs. in zap. The overhead should be minimal, but this test just verifies that.

pkg/log/logger_test.go
craigcondit: The message is purposely identical so that there's no variation in potential processing time (i.e. make sure it's not what we log but how we log it). The logger is chosen to determine which logging path we use relative to the config given.

craigcondit:
> Ok, I did a re-review. Some parts are not entirely clear, it's difficult to understand what happens without comments. You can also consider using `Example*` functions which documents how the new logger approach should be used (filtering, parent loggers, etc).

I'll try to add some more comments but I don't see the need for example functions. The original JIRA demonstrates this, and eventually all log statements will use the new form. Essentially:

`log.Logger().Info(...)` becomes `log.Log(log.LogName).Info(...)`

Everything else remains the same. The specific loggers will be subsystem-specific, and probably added to the codebase as we segregate things out. I don't want to do a high-level "dump everything by package" translation as I'd like there to be some thought given to organizing our loggers.

The existing `log.Logger()` call remains in place for backwards compatibility. Depending on context (admission controller, shim, or core), this becomes equivalent to either `log.Log(log.Admission)`, `log.Log(log.K8Shim)` or `log.Log(log.Core)` respectively.

Finally, `log.RootLogger()` provides the old behavior, but this is only exposed for initial configuration (i.e. passing the log reference to the core from the shim).

pkg/log/logger.go
craigcondit: Will update to `nullLogger` in next PR.

pkg/log/logger.go
```diff
@@ -30,12 +33,95 @@ var once sync.Once
...
+	}
+	return zapcore.InfoLevel
+}
+
+func parentLogger(name string) string {
+	i := strings.LastIndex(name, ".")
+	if i < 0 {
+		return ""
+	}
+	return name[0:i]
```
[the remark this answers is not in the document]
craigcondit: Added test coverage for both `parentLogger()` and `parseLevel()`.

craigcondit:
Updated PR (rebased on master) and addressed review comments. Added significant comments throughout, and additional test cases.

pbacsko (approved):
+1 LGTM

PR #613, opened 2023-06-14
[YUNIKORN-1575] [shim] Ability to specify guaranteedResources via namespace annotations

zhuqi-lucas:
### What is this PR for?
The shim side change for the ability to specify guaranteedResources via namespace annotations.

The interface change has been merged. https://github.com/apache/yunikorn-scheduler-interface/pull/94

The core side changes will following!

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1575

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -706,6 +706,15 @@ func (ctx *Context) updateApplicationTags(request *interfaces.AddApplicationRequ
 			request.Metadata.Tags[siCommon.AppTagNamespaceResourceQuota] = string(quotaStr)
 		}
 	}
+
+	// add guaranteed resource info as an app tag
+	guaranteedResource := utils.GetNamespaceGuaranteedFromAnnotation(namespaceObj)
+	if guaranteedResource != nil && !common.IsZero(guaranteedResource) {
+		if guaranteedStr, err := json.Marshal(guaranteedResource); err == nil {
+			request.Metadata.Tags[siCommon.AppTagNamespaceResourceGuaranteed] = string(guaranteedStr)
+		}
+	}
+
 	// add parent queue info as an app tag
 	parentQueue := utils.GetNameSpaceAnnotationValue(namespaceObj, constants.AnnotationParentQueue)
 	if parentQueue != "" {
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -944,11 +944,11 @@ func (ctx *Context) PublishEvents(eventRecords []*si.EventRecord) {
 		for _, record := range eventRecords {
 			switch record.Type {
 			case si.EventRecord_REQUEST:
-				taskID := record.ObjectID
-				appID := record.GroupID
+				appID := record.ObjectID
+				taskID := record.ReferenceID
 				if task := ctx.getTask(appID, taskID); task != nil {
 					events.GetRecorder().Eventf(task.GetTaskPod().DeepCopy(), nil,
-						v1.EventTypeNormal, record.Reason, record.Reason, record.Message)
+						v1.EventTypeNormal, "", "", record.Message)
 				} else {
 					log.Logger().Warn("task event is not published because task is not found",
 						zap.String("appID", appID),
@@ -972,7 +972,7 @@ func (ctx *Context) PublishEvents(eventRecords []*si.EventRecord) {
 					continue
 				}
 				events.GetRecorder().Eventf(node.DeepCopy(), nil,
-					v1.EventTypeNormal, record.Reason, record.Reason, record.Message)
+					v1.EventTypeNormal, "", "", record.Message)
 			default:
 				log.Logger().Warn("Unsupported event type, currently only supports to publish request event records",
 					zap.Stringer("type", record.Type))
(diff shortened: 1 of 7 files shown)


pkg/cache/context.go
```diff
@@ -706,6 +706,15 @@ func (ctx *Context) updateApplicationTags(request *interfaces.AddApplicationRequ
 			request.Metadata.Tags[siCommon.AppTagNamespaceResourceQuota] = string(quotaStr)
 		}
 	}
+
+	// add guaranteed resource info as an app tag
+	guaranteedResource := utils.GetNamespaceGuaranteedFromAnnotation(namespaceObj)
+	if guaranteedResource != nil && !common.IsZero(guaranteedResource) {
+		if guaranteedStr, err := json.Marshal(guaranteedResource); err == nil {
```
pbacsko, line 713: Nit: `guaranteedStr` is not a string, it's a byte array. Just call it `guaranteed`.

pkg/common/utils/utils.go
```diff
@@ -142,6 +142,21 @@ func PodUnderCondition(pod *v1.Pod, condition *v1.PodCondition) bool {
...
+// get namespace guaranteed resource from namespace annotation
+func GetNamespaceGuaranteedFromAnnotation(namespaceObj *v1.Namespace) *si.Resource {
+	// retrieve guaranteed resource info from annotations
+	namespaceGuaranteed := GetNameSpaceAnnotationValue(namespaceObj, constants.NamespaceGuaranteed)
+	var namespaceGuaranteedMap map[string]string
+	err := json.Unmarshal([]byte(namespaceGuaranteed), &namespaceGuaranteedMap)
+	if err != nil {
+		log.Logger().Warn("Unable to process namespace.guaranteed annotation",
+			zap.String("namespace", namespaceObj.Name),
+			zap.String("namespace.guaranteed val is", namespaceGuaranteed))
```
pbacsko, line 154: `namespace.guaranteed val is` -> just simply `namespace.guaranteed`

pbacsko (changes requested):
Two minor changes

pkg/common/utils/utils.go
zhuqi-lucas: Addressed in latest PR, thanks @pbacsko .

zhuqi-lucas:
> Two minor changes

Thanks for review @pbacsko , addressed comments.

zhuqi-lucas:
Error: ../../../go/pkg/mod/github.com/apache/yunikorn-core@v0.0.0-20230614080512-68a82820f0d7/pkg/events/events.go:31:3: unknown field GroupID in struct literal of type si.EventRecord
Error: ../../../go/pkg/mod/github.com/apache/yunikorn-core@v0.0.0-20230614080512-68a82820f0d7/pkg/events/events.go:32:3: unknown field Reason in struct literal of type si.EventRecord


Need to update core side to update SI interface too, i will create a jira for this.

zhuqi-lucas:
The e2e test fail is not related, it's the timeout error.

pkg/cache/context.go
```diff
@@ -702,10 +702,19 @@ func (ctx *Context) updateApplicationTags(request *interfaces.AddApplicationRequ
...
 		}
 	}
+
+	// add guaranteed resource info as an app tag
+	guaranteedResource := utils.GetNamespaceGuaranteedFromAnnotation(namespaceObj)
+	if guaranteedResource != nil && !common.IsZero(guaranteedResource) {
+		if guaranteed, err := json.Marshal(guaranteedResource); err == nil {
+			request.Metadata.Tags[siCommon.AppTagNamespaceResourceGuaranteed] = string(guaranteed)
+		}
+	}
```
pbacsko, line 716: Please add test coverage to `TestAddApplicationsWithTags()`, there is a dummy lister and existing namespace object which you can re-use.

pbacsko (changes requested):
Please enhance coverage (lines 713-715).

zhuqi-lucas:
@pbacsko 
One more question, how can we add the unit test to cover the log, for example, the log here, how we can improve the unit test to cover this, using log hook? Thanks!
```
	if err != nil {
		log.Logger().Warn("Unable to process namespace.guaranteed annotation",
			zap.String("namespace", namespaceObj.Name),
			zap.String("namespace.guaranteed is", namespaceGuaranteed))
		return nil
	}
```

pbacsko:
> @pbacsko One more question, how can we add the unit test to cover the log, for example, the log here, how we can improve the unit test to cover this, using log hook? Thanks!
> 
> ```
> 	if err != nil {
> 		log.Logger().Warn("Unable to process namespace.guaranteed annotation",
> 			zap.String("namespace", namespaceObj.Name),
> 			zap.String("namespace.guaranteed is", namespaceGuaranteed))
> 		return nil
> 	}
> ```

I think it's an overkill to capture log output. If you can check some side effect (return value, mock state) then do it, otherwise it's fine to skip. There are paths in the code where it's difficult to simulate an error condition, in those cases, we don't have coverage.

pbacsko (approved):
+1 LGTM.
You can add more tests if you think it's necessary.

zhuqi-lucas:
It makes sense, thanks @pbacsko for explain!

PR #614, opened 2023-06-14
[YUNIKORN-1656] Update SI reference

pbacsko:
### What is this PR for?
Update core due to `si.EventRecord` changes, because they result in compilation errors.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1656

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -935,11 +935,11 @@ func (ctx *Context) PublishEvents(eventRecords []*si.EventRecord) {
 		for _, record := range eventRecords {
 			switch record.Type {
 			case si.EventRecord_REQUEST:
-				taskID := record.ObjectID
-				appID := record.GroupID
+				appID := record.ObjectID
+				taskID := record.ReferenceID
 				if task := ctx.getTask(appID, taskID); task != nil {
 					events.GetRecorder().Eventf(task.GetTaskPod().DeepCopy(), nil,
-						v1.EventTypeNormal, record.Reason, record.Reason, record.Message)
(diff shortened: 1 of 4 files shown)


pkg/cache/context.go
```diff
@@ -935,11 +935,11 @@ func (ctx *Context) PublishEvents(eventRecords []*si.EventRecord) {
...
 			switch record.Type {
 			case si.EventRecord_REQUEST:
-				taskID := record.ObjectID
-				appID := record.GroupID
+				appID := record.ObjectID
+				taskID := record.ReferenceID
 				if task := ctx.getTask(appID, taskID); task != nil {
 					events.GetRecorder().Eventf(task.GetTaskPod().DeepCopy(), nil,
-						v1.EventTypeNormal, record.Reason, record.Reason, record.Message)
+						v1.EventTypeNormal, "", "", record.Message)
```
zhuqi-lucas, line 942: Will these two record.Reason paras will be removed for Eventf() function?

pbacsko: I don't know, we might need a replacement. We no longer have the "reason" field in the message. I'll talk to Wilfred about this. For this PR, it's OK to leave it empty and we'll see what we do.

zhuqi-lucas: Sure, it makes sense, it's ok for this PR!

PR #617, opened 2023-06-16
[YUNIKORN-1810] Use arm64 without variant for DOCKER_ARCH

wilfred-s:
### What is this PR for?
Building multi architecture image fails to properly split the variant from the main architecture and binary does not change without variant.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1810

Local and release build should generate single and multi architecture images that work on arm


zhuqi-lucas (approved):
LGTM

PR #621, opened 2023-06-22
[YUNIKORN-1821] Fix shell script in authz use cases

wusamzong:
### What is this PR for?
When the shell script uses a two-dimensional array, if you use double quotes (required by shellcheck) when accessing variables, the access results will not be as expected. ex. `username: admin admin`
I split two-dimensional array to two array.

### What type of PR is it?
* [x] - Improvement

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1821/

### Questions:

### DIFF
--- a/deployments/examples/authz/k8s-api-access/create-user.sh
+++ b/deployments/examples/authz/k8s-api-access/create-user.sh
@@ -16,6 +16,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# USERS & USERS_GROUP should contain the same number of elements.
+# If you wants no group or no user. You can use an empty string, i.e. ""
 USERS=("admin" "sue" "bob" "kim" "yono" "anonymous")
 USERS_GROUP=("admin" "group-a" "group-a" "group-b" "group-b" "anonymous")
 AUTH_FOLDER=./auth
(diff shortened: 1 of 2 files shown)


deployments/examples/authz/k8s-api-access/create-user.sh
```diff
@@ -16,15 +16,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-USERS=("admin admin" "sue group-a" "bob group-a" "kim group-b" "yono group-b" "anonymous anonymous")
+USERS=("admin" "sue" "bob" "kim" "yono" "anonymous")
+USERS_GROUP=("admin" "group-a" "group-a" "group-b" "group-b" "anonymous")
```
wilfred-s, line 20: Add a comment that the user and group should contain the same number of elements.
If anyone wants no group or no user can we use an empty string, i.e. "" ?

wilfred-s (changes requested):

wilfred-s (approved):
LGTM, would like @pbacsko to review this as he did the user/group changes in the admission controller

pbacsko (approved):
+1 lgtm

PR #628, opened 2023-07-05
[YUNIKORN-1851] Don't allow similar app ids by different users

manirajv06:
### What is this PR for?
Detect same app ids being used by different users for app submission and don't allow when app is already running. By default, this check has been disabled (configurable).

### What type of PR is it?
* [ ] - Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1851

### Questions:

### DIFF
--- a/pkg/appmgmt/general/podevent_handler.go
+++ b/pkg/appmgmt/general/podevent_handler.go
@@ -128,7 +128,7 @@ func (p *PodEventHandler) addPod(pod *v1.Pod, eventSource EventSource) interface
 				Metadata: appMeta,
 			})
 		} else {
-			if conf.GetSchedulerConf().GetAllowSimilarAppIdsByDifferentUsers() && app.GetApplicationState() == "Running" && app.GetUser() != appMeta.User {
+			if conf.GetSchedulerConf().GetSingleUserPerApplication() && app.GetUser() != appMeta.User {
 				log.Log(log.ShimAppMgmtGeneral).Warn("application has been submitted by different user",
 					zap.String("app id ", appMeta.ApplicationID),
 					zap.String("app user", app.GetUser()),
--- a/pkg/appmgmt/general/podevent_handler.go
+++ b/pkg/appmgmt/general/podevent_handler.go
@@ -25,6 +25,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 
 	"github.com/apache/yunikorn-k8shim/pkg/appmgmt/interfaces"
+	"github.com/apache/yunikorn-k8shim/pkg/common/events"
 	"github.com/apache/yunikorn-k8shim/pkg/conf"
 	"github.com/apache/yunikorn-k8shim/pkg/log"
 )
@@ -129,10 +130,16 @@ func (p *PodEventHandler) addPod(pod *v1.Pod, eventSource EventSource) interface
 			})
 		} else {
 			if conf.GetSchedulerConf().GetSingleUserPerApplication() && app.GetUser() != appMeta.User {
-				log.Log(log.ShimAppMgmtGeneral).Warn("application has been submitted by different user",
+				log.Log(log.ShimAppMgmtGeneral).Warn("rejecting application as it has been submitted by different user",
 					zap.String("app id ", appMeta.ApplicationID),
 					zap.String("app user", app.GetUser()),
 					zap.String("submitted by", appMeta.User))
+				events.GetRecorder().Eventf(pod.DeepCopy(), nil, v1.EventTypeWarning,
+					"Rejecting the application because already application exists with different user",
+					"Please try submitting a application by same user",
+					"SingleUserPerApplication has been configured to true. So rejecting the application because it has been submitted by different user. "+
+						"Either you can disable the check by not setting SingleUserPerApplication or try submitting a application by same use. "+
+						"By default, SingleUserPerApplication is false")
 				return nil
 			}
 			managedApp = app
--- a/pkg/appmgmt/general/podevent_handler.go
+++ b/pkg/appmgmt/general/podevent_handler.go
@@ -26,6 +26,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 
 	"github.com/apache/yunikorn-k8shim/pkg/appmgmt/interfaces"
+	"github.com/apache/yunikorn-k8shim/pkg/common/constants"
 	"github.com/apache/yunikorn-k8shim/pkg/common/events"
 	"github.com/apache/yunikorn-k8shim/pkg/conf"
 	"github.com/apache/yunikorn-k8shim/pkg/log"
@@ -130,7 +131,8 @@ func (p *PodEventHandler) addPod(pod *v1.Pod, eventSource EventSource) interface
 				Metadata: appMeta,
 			})
 		} else {
-			if conf.GetSchedulerConf().GetSingleUserPerApplication() && app.GetUser() != appMeta.User {
+			if app.GetApplicationID() != constants.AutoGenAppPrefix+"-"+pod.Namespace+"-"+constants.AutoGenAppSuffix &&
+				conf.GetSchedulerConf().GetSingleUserPerApplication() && app.GetUser() != appMeta.User {
 				log.Log(log.ShimAppMgmtGeneral).Warn("rejecting application as it has been submitted by different user",
 					zap.String("app id ", appMeta.ApplicationID),
 					zap.String("app user", app.GetUser()),
--- a/pkg/appmgmt/general/podevent_handler.go
+++ b/pkg/appmgmt/general/podevent_handler.go
@@ -19,6 +19,7 @@
 package general
 
 import (
+	"fmt"
 	"sync"
 
 	"go.uber.org/zap"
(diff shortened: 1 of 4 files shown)


craigcondit:
I think this might be a case of the cure being worse than the disease. Shared environments like this are somewhat cooperative by nature, and reusing app ids can happen either accidentally or even automatically (especially for the autogenerated app id). Rejecting a pod in this case is very heavy-handed and will result in unexpected failures. Tracking users by app instead of pod leads to this (IMO small) issue, but I don't think we should disallow behavior that Kubernetes allows -- we should strive to be more compatible, not less.

I don't see this issue being much worse than the ambiguity we have in the case of group quotas and multiple possible matches. It's not perfect, but probably good enough.

I'm -1 on this approach due to the pain it will likely cause.

pkg/appmgmt/general/podevent_handler.go
```diff
@@ -127,6 +128,13 @@ func (p *PodEventHandler) addPod(pod *v1.Pod, eventSource EventSource) interface
 				Metadata: appMeta,
 			})
 		} else {
+			if conf.GetSchedulerConf().GetAllowSimilarAppIdsByDifferentUsers() && app.GetApplicationState() == "Running" && app.GetUser() != appMeta.User {
```
wilfred-s, line 131: Why does the app need to be in a running state? I would expect all states to match this. Otherwise you can still get passed if an app is waiting to be scheduled.

`GetAllowSimilarAppIdsByDifferentUsers` does not seem to cover what is specified: if it returns true you make check that the users do not match. I would expect the check for the setting to be more like: `GetSingleUserPerApplication()` if true app user and task user must match. If not matched reject, otherwise allow it.
So:
```
if conf.GetSchedulerConf().GetSingleUserPerApplication() && app.GetUser() != appMeta.User {
     reject the task
}
```

wilfred-s (changes requested):
Also need to hook this into the one application ID per namespace setting. If you have one app ID per namespace this will probably break things.

wilfred-s:
> reusing app ids can happen either accidentally or even automatically (especially for the autogenerated app id)

This does not affect re-using application IDs. You can still do that without a problem. If the application is not currently known in the shim the user info will be set without question. For the auto generated app ID: I do think that is a case we need to handle and not reject.

> Tracking users by app instead of pod leads to this (IMO small) issue, but I don't think we should disallow behavior that Kubernetes allows -- we should strive to be more compatible, not less.

K8s does not have the concept of applications and or users on a workload so we have no compatibility look at. Looking at a simple case: I could set a Spark application ID on a pod as a completely different user, in a different namespace with a different service account etc. That pod will then be run as part of the already running Spark job in a queue which I might not have access to under a different quota. That is a huge gap which will show up as a problem. I am surprised that it has not as yet.

> I don't see this issue being much worse than the ambiguity we have in the case of group quotas and multiple possible matches. It's not perfect, but probably good enough.

The group resolution is defined and far less of a problem than this is.
The reason for adding this is that we have seen abuse similar to this on YARN years ago. We have to put mitigations in place. Even if they are off by default but we need to have them.

pkg/appmgmt/general/podevent_handler.go
```diff
@@ -127,6 +128,13 @@ func (p *PodEventHandler) addPod(pod *v1.Pod, eventSource EventSource) interface
...
+				log.Log(log.ShimAppMgmtGeneral).Warn("application has been submitted by different user",
+					zap.String("app id ", appMeta.ApplicationID),
+					zap.String("app user", app.GetUser()),
+					zap.String("submitted by", appMeta.User))
+				return nil
```
pbacsko, line 136: What happens if we just return here? Does the pod remain in Pending state?

wilfred-s: We should write out an event on the pod using the `events.GetRecorder().Eventf(...)`
We do that in other places in our code that makes it traceable from the K8s API on the pod.

pbacsko: +1 we have to do something so that users can see that something happened and they can act accordingly

pkg/appmgmt/general/podevent_handler.go
```diff
@@ -127,6 +129,19 @@ func (p *PodEventHandler) addPod(pod *v1.Pod, eventSource EventSource) interface
...
+				log.Log(log.ShimAppMgmtGeneral).Warn("rejecting application as it has been submitted by different user",
+					zap.String("app id ", appMeta.ApplicationID),
+					zap.String("app user", app.GetUser()),
+					zap.String("submitted by", appMeta.User))
+				events.GetRecorder().Eventf(pod.DeepCopy(), nil, v1.EventTypeWarning,
+					"Rejecting the application because already application exists with different user",
+					"Please try submitting a application by same user",
+					"SingleUserPerApplication has been configured to true. So rejecting the application because it has been submitted by different user. "+
+						"Either you can disable the check by not setting SingleUserPerApplication or try submitting a application by same use. "+
+						"By default, SingleUserPerApplication is false")
```
pbacsko, line 142: This is a lot of text. I don't think we need to send that much info. I'd send sth like:
```
events.GetRecorder().Eventf(pod.DeepCopy(), nil, v1.EventTypeWarning, fmt.Sprintf("Rejecting pod because application ID %s belongs to a different user", appMeta.ApplicationID))
```
This is more than enough.

pkg/appmgmt/general/podevent_handler_test.go
```diff
@@ -99,7 +106,117 @@ func TestRecoveryDone(t *testing.T) {
...
+			select {
+			case event := <-recorder.Events:
+				if strings.Contains(event, reason) && strings.Contains(event, message) {
+					return true
+				}
+			default:
+				return false
+			}
+		}
+	}, 50*time.Millisecond, 20*time.Millisecond)
```
pbacsko, line 188: Timeout is smaller then the interval. Use a timeout of 1 second (just `time.Second` instead of `20*time.Millisecond`) to be safe and avoid timeouts on slower machines.

pkg/appmgmt/general/podevent_handler_test.go
```diff
@@ -99,7 +106,117 @@ func TestRecoveryDone(t *testing.T) {
...
```
pbacsko, line 211: Timeout is smaller then the interval. Use a timeout of 1 second (just `time.Second` instead of `20*time.Millisecond`) to be safe and avoid timeouts on slower machines.

pkg/appmgmt/general/podevent_handler_test.go
```diff
@@ -99,7 +106,117 @@ func TestRecoveryDone(t *testing.T) {
...
+	assert.Error(t, err, "timeout waiting for condition")
```
pbacsko, line 189: Error message should be different, you do want a timeout in this case.

pbacsko (changes requested):
Found some smaller things.

pkg/appmgmt/general/podevent_handler_test.go
manirajv06: yes, we want a timeout. utils.WaitForCondition() returns "timeout waiting for condition".

pbacsko: Ah, ok. I thought that's printed when there's no timeout.

pbacsko (approved):
+1 LGTM
I'll let Wilred take a look at it too.

pkg/appmgmt/general/podevent_handler.go
```diff
@@ -127,6 +130,15 @@ func (p *PodEventHandler) addPod(pod *v1.Pod, eventSource EventSource) interface
 				Metadata: appMeta,
 			})
 		} else {
+			if conf.GetSchedulerConf().GetSingleUserPerApplication() && app.GetUser() != appMeta.User {
```
wilfred-s, line 133: This is not good enough as it will break the autogen IDs. If we generate one ID per namespace the system comes to a grinding halt. That behaviour is managed by the setting in `admissionController.filtering.generateUniqueAppId` if that is false (default) all pods that miss an app ID will have it set to: `yunikorn-NAMESPACE-autogen` that could be for multiple users.

wilfred-s (changes requested):

pkg/appmgmt/general/podevent_handler.go
manirajv06: We would skip the check for auto gen app id's

pkg/appmgmt/general/podevent_handler.go
```diff
@@ -130,7 +131,8 @@ func (p *PodEventHandler) addPod(pod *v1.Pod, eventSource EventSource) interface
 				Metadata: appMeta,
 			})
 		} else {
-			if conf.GetSchedulerConf().GetSingleUserPerApplication() && app.GetUser() != appMeta.User {
+			if app.GetApplicationID() != constants.AutoGenAppPrefix+"-"+pod.Namespace+"-"+constants.AutoGenAppSuffix &&
+				conf.GetSchedulerConf().GetSingleUserPerApplication() && app.GetUser() != appMeta.User {
```
wilfred-s, line 135: Almost there: the generated ID check is not right yet.
If `admissionController.filtering.generateUniqueAppId` is **false** we should not check the user at all if the appID is the same as the generated one. However if the setting is **true** we should check that the user is the same as it is a user set appID. Otherwise we still leave an option to bypass the user check.
Wrapping the appID check in a simple function (in: appID, out: boolean) that does the above is recommended.

I also recommend an order swap of the checks to improve performance:
1.  conf
2. appID (conf + content)
3. name

wilfred-s (changes requested):

pkg/appmgmt/general/podevent_handler.go
pbacsko: +1 for the separate function, this condition is getting complicated

craigcondit:
We need to be careful about this one. https://issues.apache.org/jira/browse/YUNIKORN-1961 was filed a few days ago that indicates that Spark uses different auth info when submitting executor pods, so somehow we need to take that into acccount here.

craigcondit:
Closing this for now. We can't make this change, as it will break a lot of things (including Spark) as executors are launched using different security credentials (the spark service account) than the original driver.

PR #629, opened 2023-07-05
[YUNIKORN-1842] improve updatePVCRefCounts

FrankYang0529:
### What is this PR for?

In the previous version, we call `updatePVCRefCounts` in `updateNode`, `updatePod`, and `removePod`. These functions call `NodeInfo.SetNode`, `NodeInfo.AddPod`, or `NodeInfo.RemovePod`. All these `NodeInfo` functions update `Generation`, so I think we don't need to use `pvcGeneration` to determine whether we need to update `pvcRefCounts`. We can just give `nodeInfo` object to `updatePVCRefCounts` and update `pvcRefCounts` in `SchedulerCache`.

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1842

Add unit tests for `updatePVCRefCounts`

### Questions:

### DIFF
--- a/pkg/cache/external/scheduler_cache.go
+++ b/pkg/cache/external/scheduler_cache.go
@@ -55,7 +55,6 @@ type SchedulerCache struct {
 	assumedPods           map[string]bool   // map of assumed pods, value indicates if pod volumes are all bound
 	pendingAllocations    map[string]string // map of pod to node ID, presence indicates a pending allocation for scheduler
 	inProgressAllocations map[string]string // map of pod to node ID, presence indicates an in-process allocation for scheduler
-	pvcGeneration         int64
 	pvcRefCounts          map[string]map[string]int
 	lock                  sync.RWMutex
 	clients               *client.Clients // client APIs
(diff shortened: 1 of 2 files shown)


pkg/cache/external/scheduler_cache.go
```diff
@@ -603,46 +603,24 @@ func (cache *SchedulerCache) IsPVCUsedByPods(key string) bool {
...
 }
 
-func (cache *SchedulerCache) maxNodeGeneration() int64 {
-	var maxGeneration int64
-	for _, node := range cache.nodesMap {
-		if maxGeneration < node.Generation {
-			maxGeneration = node.Generation
+func (cache *SchedulerCache) updatePVCRefCounts(node *framework.NodeInfo, removeNode bool) {
+	for k, v := range cache.pvcRefCounts {
+		delete(v, node.Node().Name)
```
craigcondit, line 608: Save node.Node().Name to a variable at the top of the function so it can be re-used.

craigcondit (changes requested):
Minor nit - avoid calling node.Node().Name multiple times.

pkg/cache/external/scheduler_cache.go
FrankYang0529: Updated it. Thank you!

craigcondit (approved):
+1 LGTM.

PR #630, opened 2023-07-06
[YUNIKORN-1855] fix comments

FrankYang0529:
### What is this PR for?

In the test case `TestAssignedPod`, the comment is `// the cached node has no pods assigned`, but we do `assert.Equal(t, len(cachedNode.Pods), 1)`.

In the test case `TestAddPod`, the comment is `// assumed pod should no longer be assumed if node changes`, but we do `assert.Check(t, cache.isAssumedPod(podUID1), "pod is not assumed")`.

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1855

### Questions:

### DIFF
--- a/pkg/cache/external/scheduler_cache_test.go
+++ b/pkg/cache/external/scheduler_cache_test.go
@@ -98,7 +98,7 @@ func TestAssignedPod(t *testing.T) {
 
 			// verify the node is added to cache
 			// the cached node has reference to the v1.Node object
-			// the cached node has no pods assigned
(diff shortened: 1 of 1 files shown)


craigcondit (approved):
+1 LGTM.

PR #636, opened 2023-07-16
[YUNIKORN-1852] Remove namespace from bootstrap function call

wusamzong:
### What is this PR for?
remove namespace from client.LoadBootstrapConfigMaps function call

### What type of PR is it?
* [x] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN1852/

### Questions:


wilfred-s (approved):
LGTM

PR #637, opened 2023-07-17
[YUNIKORN-1867] add priority offset e2e test

FrankYang0529:
### What is this PR for?

Add `priority.offset` property e2e test.

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1867

### Questions:

### DIFF
--- a/test/e2e/framework/helpers/k8s/k8s_utils.go
+++ b/test/e2e/framework/helpers/k8s/k8s_utils.go
@@ -343,6 +343,23 @@ func (k *KubeCtl) CreateNamespace(namespace string, annotations map[string]strin
 	return ns, nil
 }
 
+// Func to update a namespace with annotations
+func (k *KubeCtl) UpdateNamespace(namespace string, annotations map[string]string) (*v1.Namespace, error) {
+	ns, err := k.clientSet.CoreV1().Namespaces().Get(context.TODO(), namespace, metav1.GetOptions{})
+	if err != nil {
+		return ns, err
+	}
+	nsCopy := ns.DeepCopy()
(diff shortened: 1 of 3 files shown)


craigcondit (changes requested):
Seeing quite a few test failures on this, need to investigate and fix.

FrankYang0529:
> Seeing quite a few test failures on this, need to investigate and fix.

The root cause of error test cases is that we can't add child queue to a draining queue. In the first case, we add three queues `root.fence`, `root.fence.child1`, and `root.fence.child2`. After we reset the configuration, these three queues will get into the draining state. Then, another case tries to add child queues to `root.fence` and the case gets error.

Currently, the yunikorn-core will clean draining queues every 10 seconds. The duration is hardcoded in the source code. Users can't update it. Do we want to add the setting to the configuration? If yes, I can help to add it.

CI failed job: https://github.com/apache/yunikorn-k8shim/actions/runs/5575867672/jobs/10202470269
Cleaning draining queues source code: https://github.com/apache/yunikorn-core/blob/129c0de6b12b21b4282631b2c0b6df188ce7ae80/pkg/scheduler/partition_manager.go#L70-L87

wilfred-s (approved):
LGTM
Change looked more complicated due to diff layout

craigcondit (approved):
+1 LGTM.

PR #638, opened 2023-07-18
[YUNIKORN-1869] update depguard config to new format

FrankYang0529:
### What is this PR for?

We update golangci-lint from v1.51.2 to v1.53.3. In this update, upstream migrated depguard to v2, and changed the default configuration. We also need to update it, or `./tools/golangci-lint run` will show depguard errors like:

```
import 'github.com/apache/yunikorn-k8shim/pkg/cache/external' is not allowed from list 'Main' (depguard)
```

Upstream reference: https://github.com/golangci/golangci-lint/commit/fb746c4b04c3a2f2f55400d22d40192b7e659b3b

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1869

### Questions:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -45,14 +45,15 @@ linters-settings:
     lines: 120
     statements: 80
   depguard:
-    list-type: blacklist
-    include-go-root: false
-    packages:
-      - github.com/sirupsen/logrus
-      - github.com/stretchr/testify
-    packages-with-error-messages:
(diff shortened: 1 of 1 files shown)


pbacsko (approved):
+1 let's wait for the build

PR #639, opened 2023-07-19
[YUNIKORN-1874] Data race: unlocked access in Context.updatePodCondition()

pbacsko:
### What is this PR for?
The `*v1.Pod` object is updated inside `Task` without locking and it's constantly accessed to create copies.
Since  `Condition` is mutable and we need to update it, it's now stored separately so `Task.pod` can be accessed without locking.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1874

### Questions:


pkg/cache/task.go
```diff
@@ -558,3 +560,25 @@ func (task *Task) sanityCheckBeforeScheduling() error {
 	}
 	return nil
 }
+
+func (task *Task) UpdatePodCondition(podCondition *v1.PodCondition) (bool, *v1.Pod) {
+	task.lock.Lock()
+	defer task.lock.Unlock()
+
+	status := task.podStatus.DeepCopy()
+	pod := task.pod.DeepCopy()
```
zhuqi-lucas, line 569: If we have the pod deepcopy, why we also need a deepcopy of podStatus?

zhuqi-lucas (approved):
LGTM Thanks @pbacsko , a minor question about the deepcopy pod and podstatus.

pkg/cache/task.go
pbacsko: These are two distinct variables. PodStatus is updated later, but not inside "pod". Otherwise we'd need to obtain locks every time we deepcopy the pod, which is a legit approach, but in this PR I avoided this.

zhuqi-lucas: Thanks clarify, it makes sense to me now.

wilfred-s (approved):
LGTM

PR #645, opened 2023-07-27
[YUNIKORN-1748] Create test case which measures binding throughput using MockScheduler

pbacsko:
### What is this PR for?
The PR contains a new test case which is disabled by default. It can measure the performance of Yunikorn core scheduler allocation/binding throughput with k8s mocks using MockScheduler. The result is a maximum theoretical allocation rate.
CPU and heap profiles are also generated.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1748

### Questions:

### DIFF
--- a/pkg/shim/scheduler_mock_test.go
+++ b/pkg/shim/scheduler_mock_test.go
@@ -87,10 +87,11 @@ func (fc *MockScheduler) start() {
 	fc.started.Store(true)
 }
 
-func (fc *MockScheduler) updateConfig(queues string) error {
+func (fc *MockScheduler) updateConfig(queues string, extraConfig map[string]string) error {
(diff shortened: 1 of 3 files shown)


craigcondit (changes requested):
Can we implement this as a Benchmark rather than a Test? Then it can run via `make bench` without having to mess around with enabled flags, etc.

craigcondit (approved):
+1 LGTM. One observation: It would be nicer if logging were disabled during the benchmark as it makes it very difficult to see the output.

PR #651, opened 2023-08-09
[YUNIKORN-1908] Support GPU in placeholders

wilfred-s:
### What is this PR for?
Extended resources and huge pages can not over commit. The pod validator checks for a Limit setting in the pod spec. The pod fails validation without it. Placeholder creation must make sure that the pods pass the validator.
Set limits for all extended resources and hugepages-* as they cannot be over committed.

Clean up of duplicated task group definition in the test.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-1908

Unit test added
e2e test for placeholders should be extended but we do not have GPU nodes in kind, follow up jira logged to

### DIFF
--- a/pkg/cache/placeholder_test.go
+++ b/pkg/cache/placeholder_test.go
@@ -33,26 +33,71 @@ import (
 	siCommon "github.com/apache/yunikorn-scheduler-interface/lib/go/common"
 )
 
+const (
+	gpu       = "nvidia.com/gpu"
+	ephemeral = "ephemeral-storage"
+	hugepages = "hugepages-1Gi"
+)
+
+var taskGroups = []interfaces.TaskGroup{
+	{
+		Name:      "test-group-1",
+		MinMember: 10,
+		MinResource: map[string]resource.Quantity{
+			"cpu":     resource.MustParse("500m"),
+			"memory":  resource.MustParse("1024M"),
+			gpu:       resource.MustParse("2"),
+			ephemeral: resource.MustParse("2G"),
+			hugepages: resource.MustParse("2"),
+		},
+		Labels: map[string]string{
+			"labelKey0": "labelKeyValue0",
+			"labelKey1": "labelKeyValue1",
+		},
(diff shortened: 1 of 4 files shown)


pkg/cache/placeholder_test.go
```diff
@@ -33,26 +33,71 @@ import (
 	siCommon "github.com/apache/yunikorn-scheduler-interface/lib/go/common"
 )
 
+const (
+	gpu       = "nvidia.com/gpu"
+	ephemeral = "ephemeral-storage"
+	hugepages = "hugepages-1Gi"
```
pbacsko, line 39: Nit: these can referenced from core/v1: `v1.ResourceEphemeralStorage` / `v1.ResourceHugePagesPrefix`

pkg/common/utils/gang_utils.go
```diff
@@ -32,6 +32,11 @@ import (
 	"github.com/apache/yunikorn-k8shim/pkg/log"
 )
 
+const (
+	k8sDomain       = "kubernetes.io"
+	hugepagesPrefix = "hugepages-"
```
pbacsko, line 37: Nit: these can be referenced by (core/v1)  `v1.ResourceHugePagesPrefix`

pkg/cache/placeholder_test.go
pbacsko: BTW, "hugepages-xxx" can have multiple values, do we only care about `1Gi`?

wilfred-s: No, in the check we only look at the prefix `hugepages-` anything with that prefix is filtered out

pkg/common/utils/gang_utils.go
wilfred-s: done

craigcondit (approved):
+1 LGTM.

PR #634, opened 2023-07-14
[YUNIKORN-1687] add e2e for user & group quota enforcement

FrankYang0529:
### What is this PR for?

- [X] Single specific user limit
  - [X] A user can't create more than maxapplications.
  - [X] A user can't create more than maxresoruces.
  - [X] A user can create apps in other queues.
  - [X] Users not in the limit can create more than maxapplications.
  - [X] Users not in the limit can create more than maxresources.
- [X] Single specific group limit
  - [X] A group can't create more than maxapplications.
  - [X] A group can't create more than maxresoruces.
  - [X] A group can create apps in other queues.
  - [X] Groups not in the limit can create more than maxapplications.
  - [X] Groups not in the limit can create more than maxresources.
- [X] User limit is lower than group limit.
- [X] Group limit is lower than user limit.

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1687

### How should this be tested?

All e2e test cases pass.

PR #654, opened 2023-08-15
[YUNIKORN-1901] A basic example for the user tracing and the group tracing

0yukali0:
### What is this PR for?
Validate the user resource usage and group resource usage from the REST API.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [x] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1901

### Questions:


craigcondit (changes requested):
user_qauta_tracing should be user_quota_tracing ; needs to be updated everywhere.

0yukali0:
> user_qauta_tracing should be user_quota_tracing ; needs to be updated everywhere.
Hi @craigcondit .
Thanks for the review.
I am going to fix the wrong word problem later.

test/e2e/user_quota_tracing/user_qauta_tracing_test.go
```diff
@@ -0,0 +1,174 @@
...
+		queue, err = restClient.GetQueueFromUserResourceUsage(usersUsage, queuePath[1], users[2])
+		Ω(err).NotTo(HaveOccurred())
+		Ω(queue.RunningApplications).To(Equal([]string{configs[3].AppID}))
+		usedResource.ParseResourceUsage(parseResource(queue.ResourceUsage))
+		Ω(usedResource).To(Equal(map[string]int64{siCommon.CPU: 100, siCommon.Memory: 50}))
+		queue, err = restClient.GetQueueFromUserResourceUsage(usersUsage, queuePath[2], users[2])
+		Ω(err).NotTo(HaveOccurred())
+		Ω(queue.RunningApplications).To(Equal([]string{configs[2].AppID}))
+		usedResource.ParseResourceUsage(parseResource(queue.ResourceUsage))
+		Ω(usedResource).To(Equal(map[string]int64{siCommon.CPU: 100, siCommon.Memory: 50}))
```
manirajv06, line 133: Can we organize asserts one by one? first all **user** tracker rest api response assertions and then **users** tracker rest api response assertions.

Similarly for group and groups

test/e2e/user_quota_tracing/user_qauta_tracing_test.go
```diff
@@ -0,0 +1,174 @@
...
+	})
+})
+
+func parseResource(res *resources.Resource) map[string]int64 {
+	result := make(map[string]int64)
+	for key, value := range res.Resources {
+		result[key] = int64(value)
+	}
+	return result
+}
```
manirajv06, line 174: Can we keep this in rest api utils?

manirajv06 (changes requested):

manirajv06 (changes requested):
looks good. 

Some comments has been addressed partially. please check those.

manirajv06:
Now that https://github.com/apache/yunikorn-k8shim/pull/634 has gone in, you will need to rebase and resolve conflicts.

test/e2e/user_quota_tracing/user_quota_tracing_test.go
```diff
@@ -0,0 +1,261 @@
...
+			}
+			for _, queueConfig := range queuesConfigs {
+				By(fmt.Sprintf("Add child queue %s to the parent queue %s", queueConfig.QueueName, queueConfig.parentQueue))
+				config := configs.QueueConfig{
+					Name: queueConfig.QueueName,
+				}
+				if err := common.AddQueue(sc, queueConfig.partition, queueConfig.parentQueue, config); err != nil {
+					return err
+				}
+			}
```
pbacsko, line 77: I'm not overly familiar with this helper method, but can't we just replace the queue config inside the `SchedulerConfig` object with sth which initialized before? It's very hard to read even for something as simple as this.

test/e2e/user_quota_tracing/user_quota_tracing_test.go
```diff
@@ -0,0 +1,261 @@
...
+	var ns string
+	BeforeEach(func() {
+		ns = "ns-" + common.RandSeq(NANESPACE_LENGTH)
+		By(fmt.Sprintf("Create namespace: %s", ns))
+		var ns1, err1 = kClient.CreateNamespace(ns, nil)
+		Ω(err1).NotTo(HaveOccurred())
+		Ω(ns1.Status.Phase).To(Equal(v1.NamespaceActive))
+	})
+
+	It("User qauta trace with 3 users and 2 groups", func() {
```
pbacsko, line 53: Typo: "qauta"

test/e2e/user_quota_tracing/user_quota_tracing_test.go
```diff
@@ -0,0 +1,261 @@
...
+			{"assistant-app-02", queuePath[1], users[2], []string{groups[0]}},
+		}
+		for _, config := range configs {
+			By(fmt.Sprintf("Deploy the sleep app %s to the %s namespace", config.AppID, ns))
+			sleepObj, podErr := k8s.InitSleepPod(k8s.SleepPodConfig{
+				AppID:  config.AppID,
+				NS:     ns,
+				Labels: map[string]string{"queue": config.Queue},
+			})
+			sleepObj.ObjectMeta.Annotations["yunikorn.apache.org/user.info"] = fmt.Sprintf("{username:%s, groups:{%s}}", config.UserName, config.GroupsNames)
```
pbacsko, line 104: This will be rejected by the admission controller unless the testcase (or rather the user which the TC uses, which is "kubernetes-admin" in case of KIND) is allowed to do it. A new entry is necessary in the configmap:
`admissionController.accessControl.externalUsers: "kubernetes-admin"`

You can look at the existing admission tests to see how it's configured.

test/e2e/user_quota_tracing/user_quota_tracing_test.go
```diff
@@ -0,0 +1,261 @@
...
+		*  user3 -> group1, group2
+		 */
+		configs := []struct {
+			AppID, Queue, UserName string
+			GroupsNames            []string
+		}{
+			{"teacher-app-01", queuePath[1], users[0], []string{groups[0]}},
+			{"students-app-01", queuePath[2], users[1], []string{groups[1]}},
+			{"assistant-app-01", queuePath[2], users[2], []string{groups[1]}},
+			{"assistant-app-02", queuePath[1], users[2], []string{groups[0]}},
```
pbacsko, line 95: If these are strings which do not change inside the slice, I'd prefer referencing them as literals rather than slice elements. This is difficult to interpret. Extract them to const if that helps.

test/e2e/user_quota_tracing/user_quota_tracing_test.go
```diff
@@ -0,0 +1,261 @@
...
+						queuePath[2]: {[]string{}, zeroPod},
+					},
+				},
+				users[2]: {
+					map[string]string{configs[2].AppID: groups[1], configs[3].AppID: groups[0]},
+					map[string]Queue{
+						queuePath[0]: {[]string{configs[2].AppID, configs[3].AppID}, twoPods},
+						queuePath[1]: {[]string{configs[3].AppID}, onePod},
+						queuePath[2]: {[]string{configs[2].AppID}, onePod},
+					},
```
pbacsko, line 214: Refer to strings directly

test/e2e/user_quota_tracing/user_quota_tracing_test.go
```diff
@@ -0,0 +1,261 @@
...
+			Ω(getGroupErr).NotTo(HaveOccurred())
+			Ω(groupUsage.Applications).To(Equal(quota.RunningApplications), "running application IDs are not expected")
+			for queuePath, queueQuota := range quota.Queues {
+				queue, getQueueErr := yunikorn.GetQueueResourceUsage(groupUsage.Queues, queuePath)
+				Ω(getQueueErr).NotTo(HaveOccurred())
+				Ω(queue.RunningApplications).To(Equal(queueQuota.RunningApplications))
+				usedResource.ParseResourceUsage(yunikorn.ParseResource(queue.ResourceUsage))
+				Ω(usedResource).To(Equal(queueQuota.Resources))
+			}
+		}
```
pbacsko, line 162: I'm wondering if using functions is better... because you're trying to avoid code duplication, which is good.
But perhaps having all these checks in a separate func is more readable, like:

```
type PerGroup struct {
  queuePath string
  runningApps []string
  resources map[string]int64
}

func checkGroups(group string, expected []PerGroup) {
  ... do things with "group" ...
 groupUsage := restClient.GetGroupResourceUsage(DEFAULT_PARTITION, group)
  for _, exp := range expected {
    checkForGroup(groupUsage, exp.queuePath, exp.runningApps, exp.resources)
  }
}

func checkForGroup(groupUsage *dao.GroupResourceUsageDAOInfo, queuePath string, expectedRunningApps []string,
expectedResource map[string]int64) {
 queue, getQueueErr := yunikorn.GetQueueResourceUsage(groupUsage.Queues, queuePath)
 Ω(getQueueErr).NotTo(HaveOccurred())
 Ω(queue.RunningApplications).To(Equal(expectedRunningApps))
 usedResource.ParseResourceUsage(yunikorn.ParseResource(queue.ResourceUsage))
 Ω(usedResource).To(Equal(expectedResource))
}
```

Then you call `checkGroups()` separately for each group, and you create `[]PerGroup` inline with string literals:
```
checkGroups("group1", []PerGroup{
  {
    queuePath: "root.path1",
    runningApps: "app1", "app2", "app3",
    resources: map[string]int64{
      "cpu": 1,
      "memory": 100,
   },
  {
    queuePath: "root.path2",
    runningApps: "app4", "app5",
    resources: map[string]int64{
      "cpu": 2,
      "memory": 200,
   },
  }
})
```
I'm not saying it has to look like this (it might not even compile), but it's a start.

The main idea here is to reduce nesting and the number of structs. I'm really having a hard time understanding what the code does with all these types.

test/e2e/user_quota_tracing/user_quota_tracing_suite_test.go
```diff
@@ -0,0 +1,47 @@
...
+var kClient = k8s.KubeCtl{} //nolint
+
+var _ = BeforeSuite(func() {
+	Ω(kClient.SetClient()).To(BeNil())
+	yunikorn.EnsureYuniKornConfigsPresent()
+})
+
+var _ = AfterSuite(func() {
+	yunikorn.RestoreConfigMapWrapper(oldConfigMap, annotation)
+})
```
pbacsko, line 47: Maybe I'm missing something... Don't you need a Test* function here?

test/e2e/user_quota_tracing/user_quota_tracing_test.go
pbacsko, line 104: As I mentioned below (that was the one which I spotted first), here we avoid code duplication by using a for loop and new types.

How about this?
```
deploySleepPod("teacher-app-01", ns, "root.group1_resources", []{"group1"})
deploySleepPod("students-app-01", ns, "root.group2_resources", []{"group2"})
deploySleepPod("somemoreapp-111", ns, "root.somequeue", []{"group2"})   // fix params
deploySleepPod("evenmoreapp-222", ns, "root.anotherqueue", []{"group1,group2"}) // fix params
...

func deploySleepPod(appId, ns, queue, username string,  groups []string) {
  By(fmt.Sprintf("Deploy the sleep app %s to the %s namespace", appId, ns))
			sleepObj, podErr := k8s.InitSleepPod(k8s.SleepPodConfig{
				AppID:  appId,
				NS:     ns,
				Labels: map[string]string{"queue": queue},
			})
			sleepObj.ObjectMeta.Annotations["yunikorn.apache.org/user.info"] = fmt.Sprintf("{username:%s, groups:{%s}}", username, groups)
}
```

Much clearer I believe.

0yukali0:
> Now that #634 has gone in, you will need to rebase and resolve conflicts.

Understood, I will rebase in this weekend.

test/e2e/user_quota_tracing/user_quota_tracing_test.go
0yukali0: Thanks for comments.
I will check the the way to configure external users in admin e2e test

test/e2e/user_quota_tracing/user_quota_tracing_test.go
0yukali0: Hi @pbacsko , thanks for review.
Your suggestion sounds great in order to reduce the code and make code readable.
I  will try to adopt this suggestion to make this code better.
I am going to start the progress in this weekend.

test/e2e/user_quota_tracing/user_quota_tracing_test.go
0yukali0: Understood, I will add this method to avoid code duplication.

0yukali0:
The commit is updated but it still lack of the part of admincontroll.externalUser
I will update the code of that part in next weekend.

PR #655, opened 2023-08-18
[YUNIKORN-1894] remove old gcr image locations in examples

wilfred-s:
### What is this PR for?
Update pause image to 3.7 aligning with default image for placeholders.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-1894

Not applicable, example references only

### DIFF
--- a/deployments/examples/predicates/node-affinity-example.yaml
+++ b/deployments/examples/predicates/node-affinity-example.yaml
@@ -37,7 +37,7 @@ spec:
                   - e2e-az2
   containers:
     - name: with-node-affinity
-      image: k8s.gcr.io/pause:2.0
+      image: registry.k8s.io/pause:3.7
       resources:
         requests:
           cpu: "100m"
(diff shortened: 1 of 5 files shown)


deployments/examples/predicates/node-affinity-example.yaml
```diff
@@ -37,8 +37,8 @@ spec:
...
   containers:
     - name: with-node-affinity
-      image: k8s.gcr.io/pause:2.0
+      image: registry.k8s.io/pause:3.7
       resources:
         requests:
           cpu: "100m"
-          memory: "500M"
\ No newline at end of file
+          memory: "500M"
```
manirajv06, line 44: as long as this change is not because of any tab char, indent etc we should be good. When users try these examples as is, they should not face any unnecessary issues.

craigcondit (approved):
+1 LGTM.

PR #658, opened 2023-08-23
[YUNIKORN-1930] Don't send all node events to the shim

pbacsko:
### What is this PR for?
Don't send all node events to the shim, because that will generate a lot of network traffic & load on the API server.
Filter all events except node added/removed.
Also don't output WARN message if the type is not QUEUE or NODE.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1930

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -982,6 +982,9 @@ func (ctx *Context) PublishEvents(eventRecords []*si.EventRecord) {
 						zap.Stringer("event", record))
 				}
 			case si.EventRecord_NODE:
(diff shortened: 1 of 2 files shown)


craigcondit (approved):
+1 LGTM.

PR #660, opened 2023-08-25
[YUNIKORN-1928] Remove the unused skip logic for auto scaling container.

zhuqi-lucas:
### What is this PR for?
In shim side we have skip autoscaling logic both for yunikorn mode and plugin mode.

```
switch request.State {
case si.UpdateContainerSchedulingStateRequest_SKIPPED:
    // auto-scaler scans pods whose pod condition is PodScheduled=false && reason=Unschedulable
    // if the pod is skipped because the queue quota has been exceed, we do not trigger the auto-scaling
    task.SetTaskSchedulingState(interfaces.TaskSchedSkipped)
    if ctx.updatePodCondition(task,
       &v1.PodCondition{
          Type:    v1.PodScheduled,
          Status:  v1.ConditionFalse,
          Reason:  "SchedulingSkipped",
          Message: request.Reason,
       }) {
       events.GetRecorder().Eventf(task.pod.DeepCopy(), nil,
          v1.EventTypeNormal, "PodUnschedulable", "PodUnschedulable",
          "Task %s is skipped from scheduling because the queue quota has been exceed", task.alias)
    }
```
We need to remove the unused misleading logic.

  

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1928

### Questions:

### DIFF
--- a/pkg/appmgmt/interfaces/task_sched_state.go
+++ b/pkg/appmgmt/interfaces/task_sched_state.go
@@ -22,7 +22,6 @@ type TaskSchedulingState int8
 
 const (
 	TaskSchedPending TaskSchedulingState = iota
-	TaskSchedSkipped
 	TaskSchedFailed
 	TaskSchedAllocated
 )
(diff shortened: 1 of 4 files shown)


craigcondit:
This logic is 100% needed and should not be removed. It is used to flag pods that have been skipped so that they may be treated differently by the scheduler plugin logic. Unlike the standalone scheduler mode (which does not need this), plugin mode does, because the default scheduler itself is the one responsible for setting the Pod state. Without this, PreEnqueue functionality is broken.

PR #661, opened 2023-08-25
[YUNIKORN-1938] [WIP] Build against Kubernetes v1.28.1

craigcondit:
### What is this PR for?
Updates the shim build dependency to Kubernetes v1.28.1. This release adds a QueuingHint function that can be utilized to quickly rule out previously unschedulable Pods that are not yet ready for scheduling. On a busy cluster with a large backlog, this can potentially save a lot of CPU cycles.

WIP as this approach is not yet tested; just want to see if e2e tests pass and what performance impact this may have.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1938

### Questions:


pkg/cache/external/scheduler_cache.go
```diff
@@ -520,6 +550,48 @@ func (cache *SchedulerCache) forgetPod(pod *v1.Pod) {
...
+	limit := len(taskID)
+	if limit > 4 {
+		limit = 4
+	}
+	for i := 0; i < limit; i++ {
+		if !filter.data[i][taskID[i]] {
+			return false
+		}
+	}
+	return true
```
pbacsko, line 594: Nit: I did some benchmarks with this piece of code and this performs better, although I have no idea whether it matters in the grand scheme of things:

```
func (filter *taskBloomFilter) addTask(taskID string) {
	filter.data[0][taskID[0]] = true
	filter.data[1][taskID[1]] = true
	filter.data[2][taskID[2]] = true
	filter.data[3][taskID[3]] = true
}

func (filter *taskBloomFilter) isTaskMaybePresent(taskID string) bool {
	return filter.data[0][taskID[0]] &&
		filter.data[1][taskID[1]] &&
		filter.data[2][taskID[2]] &&
		filter.data[3][taskID[3]]
}
```

I wanted to see if it was possible to speed it up further (mutex, counting bloom filter to avoid creating new instances, using atomic variable as a spin lock), but no, this approach is the fastest (at least based on my testing).

craigcondit: Yes I originally tried that but it's not 100% guaranteed that we have at least 4 characters in the ID. 

How much of a delta (ns/op and percentage) did you observe?

pbacsko: I realized that I messed up the benchmark, I forgot to call a method and it gave misleading results... I re-ran the fixed code. There's a small performance gain, but much smaller than what I measured on Thursday:

```
BenchmarkBloomFilter
BenchmarkBloomFilter-12     	  401397	      2733 ns/op	    6144 B/op	       6 allocs/op
BenchmarkBloomFilter2
BenchmarkBloomFilter2-12    	  478338	      2459 ns/op	    6144 B/op	       6 allocs/op
PASS
```

The difference is negligible.

craigcondit:
Closing this as no real performance gains have been seen.

PR #663, opened 2023-08-29
[YUNIKORN-1933]-Verify_allow_preemption_tag e2e test is flaky

wusamzong:
### What is this PR for?
Merely utilizing `yunikorn.UpdateCustomConfigMapWrapper` to modify the config map doesn't ensure instantaneous and accurate reconfiguration of YuniKorn's queues. This leads to the rejection of `Verify_allow_preemption_tag` by inappropriate placement rules in testing. As a more dependable alternative, I consequently attempted to restart YuniKorn in the AfterEach phase.

### What type of PR is it?
* [*] - Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1933/

### Questions:

### DIFF
--- a/test/e2e/preemption/preemption_test.go
+++ b/test/e2e/preemption/preemption_test.go
@@ -386,7 +386,7 @@ var _ = ginkgo.Describe("Preemption", func() {
 			// Wait for pod to move to running state
 			podErr = kClient.WaitForPodBySelectorRunning(dev,
 				fmt.Sprintf("app=%s", sleepRespPod.ObjectMeta.Labels["app"]),
-				60)
+				60*60*2)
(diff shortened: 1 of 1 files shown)


wusamzong:
After the github check fail, I've changed the solution.
After checking the log, I found that if the original queue name may be `A,B,C`, and the updated queue name is `A,D,E`, then the sheduler will create `A,D,E` first, then delete `A,B,C`.
To avoid queue deletion due to the same name, change the name of the test queue.

craigcondit (approved):
+1 LGTM.

PR #664, opened 2023-08-30
[YUNIKORN-1931] Improve placeholder name generation

craigcondit:
### What is this PR for?
The current placeholder name generation algorithm has numerous weaknesses including the possibility of collisions. Use a better algorithm that also introduces a random suffix to ensure that duplicates are never generated.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1931

### Questions:


craigcondit:
Removed a few e2e tests related to invalid gang scheduling handling as they are not correct and we have an open issue around handling validation anyway.

pkg/common/utils/gang_utils.go
```diff
@@ -57,11 +58,24 @@ func FindAppTaskGroup(appTaskGroups []*interfaces.TaskGroup, groupName string) (
 // GeneratePlaceholderName creates the placeholder name for a pod, pod name can not be longer than 63 chars,
 // taskGroup name and appID will be truncated if they go over 20/28 chars respectively,
 // each taskGroup is assigned with an incremental index starting from 0.
-func GeneratePlaceholderName(taskGroupName, appID string, index int32) string {
-	// taskGroup name no longer than 20 chars
+func GeneratePlaceholderName(taskGroupName, appID string) string {
```
wilfred-s, line 61: NIT: Update the comment to align with the new generation logic.

pkg/common/utils/gang_utils.go
```diff
@@ -57,11 +58,24 @@ func FindAppTaskGroup(appTaskGroups []*interfaces.TaskGroup, groupName string) (
...
+}
+
+const letterBytes = "abcdefghijklmnopqrstuvwxyz0123456789"
+
+// generateNonce generates an alphanumeric string (i.e. base-36) of the given length
+func generateNonce(length int) string {
+	buf := make([]byte, length)
+	for i := range buf {
+		// this uses the built-in rand package as it does not need to be hardened
+		buf[i] = letterBytes[rand.Intn(len(letterBytes))] //nolint:gosec
```
wilfred-s, line 76: math/rand is deterministic which means it generates the same values each time. That could lead to clashes in naming. crypto/rand is much safer.
```
	maxIndex := big.NewInt(int64(len(letterBytes)))
	buf := make([]byte, length)
	for i := range buf {
		num, err := rand.Int(rand.Reader, maxIndex)
		if err != nil {
			// something bad happened handle it...
			return ""
		}
	}
	return string(buf)
```

wilfred-s (changes requested):

pkg/cache/placeholder_manager.go
```diff
@@ -77,21 +77,26 @@ func (mgr *PlaceholderManager) createAppPlaceholders(app *Application) error {
...
 	for _, tg := range app.getTaskGroups() {
-		for i := int32(0); i < tg.MinMember; i++ {
-			placeholderName := utils.GeneratePlaceholderName(tg.Name, app.GetApplicationID(), i)
-			// when performing recovery, do not create pods that are already running
-			if _, ok := existingPlaceHolders[placeholderName]; ok {
-				log.Log(log.ShimCachePlaceholder).Info("Placeholder pod already exists",
-					zap.String("name", placeholderName))
-				continue
-			}
+		count, ok := tgCounts[tg.Name]
```
pbacsko, line 93: Nit: similar, just `count := tgCount[tg.Name]`

pkg/cache/placeholder_manager.go
```diff
@@ -77,21 +77,26 @@ func (mgr *PlaceholderManager) createAppPlaceholders(app *Application) error {
...
-	for _, phTasks := range app.GetPlaceHolderTasks() {
-		existingPlaceHolders[phTasks.GetTaskPod().GetName()] = struct{}{}
+	// map task group to count of already created placeholders
+	tgCounts := make(map[string]int32)
+	for _, ph := range app.getPlaceHolderTasks() {
+		tgName := ph.getTaskGroupName()
+		if count, ok := tgCounts[tgName]; ok {
+			tgCounts[tgName] = count + 1
+		} else {
+			tgCounts[tgName] = 1
```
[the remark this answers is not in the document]
craigcondit: Done

pkg/common/utils/gang_utils.go
craigcondit: Done

pkg/cache/placeholder_manager.go
craigcondit: Done

craigcondit:
Rebased on latest master and addressed review comments.

pbacsko (approved):
+1

PR #668, opened 2023-09-03
[YUNIKORN-1952] Verify_allow_preemption_tag e2e test doesn't really r…

wusamzong:
…ely on annotation on PriorityClass

### What is this PR for?
In the Admission Controller, it adds the allowPreemption annotation based on the pod's priority class[1]. Initially, it checks the pod's annotation and then the priorityClass, which cannot be tested in the Verify_allow_preemption_tag e2e test, since InitSleepPod() creates the configuration with the allowPreemption annotation.

I've introduced an enum to control whether InitSleepPod() should include the allowAnnotation or not, and it can be set to either True or False. 

[1] https://github.com/apache/yunikorn-k8shim/blob/master/pkg/admission/admission_controller.go#L379C2-L392C3

### Todos
* [x] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1952/

### Questions:

### DIFF
--- a/test/e2e/framework/helpers/k8s/pod_conf.go
+++ b/test/e2e/framework/helpers/k8s/pod_conf.go
@@ -39,9 +39,17 @@ type SleepPodConfig struct {
 	CPU          int64
 	Mem          int64
 	RequiredNode string
-	Optedout     bool
+	Optedout     AllowPreemptOpted
 	Labels       map[string]string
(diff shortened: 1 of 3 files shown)


zhuqi-lucas (approved):
LGTM

PR #677, opened 2023-09-15
[YUNIKORN-1351] alway prefer annotations above labels

chenyulin0719:
### What is this PR for?
Since we're looking for deprecated labels and move everything to annotation in the future.
This PR change newly added labels to annotations in admission controller when mutating the pods. 
This PR also fix existing mismatch to align the logic of "fetching annotations from Pods takes precedence over labels."

Below are the detailed change list:
1. Get queue name from annotation first. (GetQueueNameFromPod)
2. Get isStateAwareDisabled from annotation first  (isStateAwareDisabled)
3. Move lables applicationId/queue in newPlaceholder to annotation (newPlaceholder)
4. Add new annotation 'yunikorn.apache.org/disable-state-aware' to replace equivalent label.
5. Rename updateLabels to updateApplicationInfo in AdmissionController
6. Rename util.updatePodLabel to util.getAnnotationsForApplicationInfoUpdate

### What type of PR is it?

* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1351
Before vs After result. (Pod with existing applicationId, queue labels.)
YUNIKORN-1351  Before vs After (Pod with labels)

Before vs After result. (Pod with no label.)
YUNIKORN-1351  Before vs After (Pod without labels)

### Questions:
* [ ] - The original functions admission_controller.updateLabels(), util.updatePodLabel() can't reflect the behaviors anymore, so I renamed it. Please let me know if we have better name for those two functions.
* [ ] - When state aware annotation set but can't be converted to boolean, I'll return false without check label's value. The reason is we're not expecting the state aware policy be set in label. ([refer this test code](https://github.com/apache/yunikorn-k8shim/compare/master...chenyulin0719:yunikorn-k8shim:YUNIKORN-1351#diff-8965c48e09eaf9638694b621a5881ab8ef9d195d813cabb4816e33967e4e1ee2R681))
* [ ] - There will be another [PR ](https://github.com/apache/yunikorn-site/pull/337)to update this [doc](https://yunikorn.apache.org/docs/user_guide/labels_and_annotations_in_yunikorn/)

### DIFF
--- a/pkg/admission/admission_controller.go
+++ b/pkg/admission/admission_controller.go
@@ -196,7 +196,7 @@ func (c *AdmissionController) processPod(req *admissionv1.AdmissionRequest, name
 	patch = updateSchedulerName(patch)
 
 	if c.shouldLabelNamespace(namespace) {
-		patch = c.updateLabels(namespace, &pod, patch)
+		patch = c.updateApplicationInfo(namespace, &pod, patch)
 		patch = c.updatePreemptionInfo(&pod, patch)
 	} else {
 		patch = disableYuniKorn(namespace, &pod, patch)
@@ -369,8 +369,8 @@ func (c *AdmissionController) checkUserInfoAnnotation(getAnnotation func() (stri
 func updateSchedulerName(patch []common.PatchOperation) []common.PatchOperation {
 	log.Log(log.Admission).Info("updating scheduler name")
 	return append(patch, common.PatchOperation{
-		Op:    "add",
-		Path:  "/spec/schedulerName",
+		Op:    constants.AddPatchOp,
+		Path:  constants.SchedulerNamePatchPath,
 		Value: constants.SchedulerName,
 	})
 }
@@ -397,7 +397,7 @@ func (c *AdmissionController) updatePreemptionInfo(pod *v1.Pod, patch []common.P
 
 	// check for an existing patch on annotations and update it
 	for _, p := range patch {
-		if p.Op == "add" && p.Path == "/metadata/annotations" {
+		if p.Op == constants.AddPatchOp && p.Path == constants.AnnotationPatchPath {
 			if annotations, ok := p.Value.(map[string]string); ok {
 				annotations[constants.AnnotationAllowPreemption] = value
 				return patch
@@ -407,29 +407,61 @@ func (c *AdmissionController) updatePreemptionInfo(pod *v1.Pod, patch []common.P
 
 	result := updatePodAnnotation(pod, constants.AnnotationAllowPreemption, value)
 	patch = append(patch, common.PatchOperation{
-		Op:    "add",
-		Path:  "/metadata/annotations",
+		Op:    constants.AddPatchOp,
+		Path:  constants.AnnotationPatchPath,
 		Value: result,
 	})
 
 	return patch
 }
 
-func (c *AdmissionController) updateLabels(namespace string, pod *v1.Pod, patch []common.PatchOperation) []common.PatchOperation {
-	log.Log(log.Admission).Info("updating pod labels",
+func (c *AdmissionController) updateApplicationInfo(namespace string, pod *v1.Pod, patch []common.PatchOperation) []common.PatchOperation {
+	log.Log(log.Admission).Info("updating pod application labels and annotations",
 		zap.String("podName", pod.Name),
 		zap.String("generateName", pod.GenerateName),
 		zap.String("namespace", namespace),
-		zap.Any("labels", pod.Labels))
+		zap.Any("labels", pod.Labels),
+		zap.Any("annotations", pod.Annotations))
 
-	result := updatePodLabel(pod, namespace, c.conf.GetGenerateUniqueAppIds(), c.conf.GetDefaultQueueName())
+	newLabels, newAnnotations := getNewApplicationInfo(pod, namespace, c.conf.GetGenerateUniqueAppIds(), c.conf.GetDefaultQueueName())
 
-	patch = append(patch, common.PatchOperation{
-		Op:    "add",
-		Path:  "/metadata/labels",
-		Value: result,
-	})
+	patch = updatePatch(pod, newLabels, constants.LabelPatchPath, patch)
+	patch = updatePatch(pod, newAnnotations, constants.AnnotationPatchPath, patch)
 
+	return patch
+}
+
+func updatePatch(pod *v1.Pod, newValues map[string]string, patchPath string, patch []common.PatchOperation) []common.PatchOperation {
+	// if found an existing patch, add new values to it
+	for _, p := range patch {
+		if p.Op == constants.AddPatchOp && p.Path == patchPath {
+			if existingPatchValues, ok := p.Value.(map[string]string); ok {
+				for k, v := range newValues {
+					existingPatchValues[k] = v
+				}
+				return patch
+			}
+		}
+	}
+
+	// Add a new patch
+	if len(newValues) != 0 {
+		// combine new values with existing values in pod to create first patch
+		var existingPodValues map[string]string
+		if patchPath == constants.LabelPatchPath {
+			// newly add labels patch should include existing labels in pod
+			existingPodValues = pod.Labels
+		} else if patchPath == constants.AnnotationPatchPath {
+			// newly add annotations patch should include existing annotations in pod
+			existingPodValues = pod.Annotations
+		}
+		patchValue := utils.MergeMaps(existingPodValues, newValues)
+		patch = append(patch, common.PatchOperation{
+			Op:    constants.AddPatchOp,
+			Path:  patchPath,
+			Value: patchValue,
+		})
+	}
 	return patch
 }
 
@@ -442,8 +474,8 @@ func disableYuniKorn(namespace string, pod *v1.Pod, patch []common.PatchOperatio
 	result := updatePodAnnotation(pod, constants.AnnotationIgnoreApplication, constants.True)
 
 	patch = append(patch, common.PatchOperation{
-		Op:    "add",
-		Path:  "/metadata/annotations",
+		Op:    constants.AddPatchOp,
+		Path:  constants.AnnotationPatchPath,
 		Value: result,
 	})
 
--- a/pkg/admission/util.go
+++ b/pkg/admission/util.go
@@ -29,42 +29,86 @@ import (
 	"github.com/apache/yunikorn-k8shim/pkg/log"
 )
 
-func updatePodLabel(pod *v1.Pod, namespace string, generateUniqueAppIds bool, defaultQueueName string) map[string]string {
-	existingLabels := pod.Labels
-	result := make(map[string]string)
-	for k, v := range existingLabels {
-		result[k] = v
+func getNewApplicationInfo(pod *v1.Pod, namespace string, generateUniqueAppIds bool, defaultQueueName string) (map[string]string, map[string]string) {
+	newLabels := make(map[string]string)
+	newAnnotations := make(map[string]string)
+
+	appID, isAppIdFromLabel := getApplicationIDFromPod(pod)
+	disableStateAware, isDisableStateAwareFromLabel := getDisableStateAwareFromPod(pod)
+
+	// for undefined configuration, am_conf will add 'root.default' as default queue name
+	// if a custom name is configured for default queue, it will be used instead of root.default
+	queueName, isQueueNameFromLabel := utils.GetQueueNameFromPod(pod, defaultQueueName)
+
+	if isAppIdFromLabel || isDisableStateAwareFromLabel || isQueueNameFromLabel {
+		log.Log(log.Admission).Warn("Pod contains label for Yunikorn configuration. This is deprecated and will be ignored in a future release. Please migrate to annotation-based configuration.",
+			zap.Bool("applicationId from label", isAppIdFromLabel),
+			zap.Bool("disableStateAware from label", isDisableStateAwareFromLabel),
+			zap.Bool("queue name from label", isQueueNameFromLabel))
 	}
 
-	sparkAppID := utils.GetPodLabelValue(pod, constants.SparkLabelAppID)
-	appID := utils.GetPodLabelValue(pod, constants.LabelApplicationID)
-	if sparkAppID == "" && appID == "" {
+	if appID == "" {
 		// if app id not exist, generate one
 		// for each namespace, we group unnamed pods to one single app - if GenerateUniqueAppId is not set
 		// if GenerateUniqueAppId:
 		//		application ID convention: ${NAMESPACE}-${POD_UID}
 		// else
 		// 		application ID convention: ${AUTO_GEN_PREFIX}-${NAMESPACE}-${AUTO_GEN_SUFFIX}
-		generatedID := utils.GenerateApplicationID(namespace, generateUniqueAppIds, string(pod.UID))
-		result[constants.LabelApplicationID] = generatedID
+		appID = utils.GenerateApplicationID(namespace, generateUniqueAppIds, string(pod.UID))
 
 		// if we generate an app ID, disable state-aware scheduling for this app
-		if _, ok := existingLabels[constants.LabelDisableStateAware]; !ok {
-			result[constants.LabelDisableStateAware] = "true"
+		// skip it if disableStateAware has already been set in the pod
+		if disableStateAware == "" {
+			disableStateAware = "true"
 		}
 	}
 
-	// if existing label exist, it takes priority over everything else
-	if _, ok := existingLabels[constants.LabelQueueName]; !ok {
-		// if defaultQueueName is "", skip adding default queue name to the pod labels
-		if defaultQueueName != "" {
-			// for undefined configuration, am_conf will add 'root.default' to retain existing behavior
-			// if a custom name is configured for default queue, it will be used instead of root.default
-			result[constants.LabelQueueName] = defaultQueueName
-		}
+	// we're looking forward to deprecate the labels and move everything to annotations
+	// but for backforward compatibility, we still add to labels
+	newLabels = updateLabelIfNotPresentInPod(pod, newLabels, constants.LabelApplicationID, appID)
+	newAnnotations = updateAnnotationIfNotPresentInPod(pod, newAnnotations, constants.AnnotationApplicationID, appID)
+
+	// skip adding empty disableStateAware to the pod
+	if disableStateAware != "" {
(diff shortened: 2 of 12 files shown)


pkg/cache/placeholder.go
```diff
@@ -92,11 +92,12 @@ func newPlaceholder(placeholderName string, app *Application, taskGroup interfac
 			Name:      placeholderName,
 			Namespace: app.tags[constants.AppTagNamespace],
 			Labels: utils.MergeMaps(taskGroup.Labels, map[string]string{
-				constants.LabelApplicationID:   app.GetApplicationID(),
-				constants.LabelQueueName:       app.GetQueue(),
 				constants.LabelPlaceholderFlag: "true",
```
FrankYang0529, line 95: If we prefer to use annotations, we may need to set the value in annotations.

pkg/admission/util.go
```diff
@@ -31,38 +31,50 @@ import (
...
-			result[constants.LabelDisableStateAware] = "true"
+		disableStateAware := "true"
+		if value := utils.GetPodLabelValue(pod, constants.LabelDisableStateAware); value != "" {
+			disableStateAware = value
 		}
+		result[constants.AnnotationDisableStateAware] = disableStateAware
+	}
+
+	// if app id not in pod annotation, add it
+	if value := utils.GetPodAnnotationValue(pod, constants.AnnotationIgnoreApplication); value == "" {
```
FrankYang0529, line 60: @craigcondit The `yunikorn.apache.org/ignore-application` was from PR https://github.com/apache/yunikorn-k8shim/pull/579, but it's not on the [document](https://yunikorn.apache.org/docs/user_guide/labels_and_annotations_in_yunikorn). I'm not sure whether users can set the value?

If not, we don't need to check the value here. If yes, we also need to check whether value can be parsed as bool. Thank you.
cc @chenyulin0719

FrankYang0529 (changes requested):
Need some minor changes. Thanks for the contribution.

pkg/admission/util.go
chenyulin0719: Hi @FrankYang0529, 

Sorry, I just realized that I used the wrong constants here. 
I should change the constant `constants.AnnotationIgnoreApplication -> constants.AnnotationApplicationID`.

We should always add app-id to annotation if it wasn't in pod.  Will fix it.

pkg/common/utils/utils.go
```diff
@@ -94,10 +94,10 @@ func IsAssignedPod(pod *v1.Pod) bool {
 
 func GetQueueNameFromPod(pod *v1.Pod) string {
```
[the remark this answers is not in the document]
chenyulin0719: Thanks for the comment. Have changed it in this commit cf8be1428c6c72500447433c25d0dbf46ace71a6.

pkg/admission/util.go
chenyulin0719: Committed a fixed version b08d5cda0d031e630e48928b4810a752b0b2d670.
Also add more comments/tests to make the logic clearer.

pkg/cache/placeholder.go
chenyulin0719: Hi @FrankYang0529,
Base on my understanding, I've set queueName and AppID to annotation in admission controller.
And have set annotation value is newPlaceholder.  [(Code)](https://github.com/apache/yunikorn-k8shim/pull/677/files#diff-2c7c04bd874ac06a9c003357d6129e83fff3e161e92f3a46cbcd1be545219f75R97-R100)

In the admission controller function pkg/admission/admission_controller#updateApplicationInfo,
I've moved the labels below to annotations. So the newPlaceholder should be able to fetch it.
- applicationId
- queue
- disableStateAware

Please kindly let me know if I have any misunderstanding.

pkg/admission/util.go
```diff
@@ -31,38 +31,50 @@ import (
...
-	// if existing label exist, it takes priority over everything else
-	if _, ok := existingLabels[constants.LabelQueueName]; !ok {
+	queueName := utils.GetQueueNameFromPod(pod, "")
+	if queueName == "" {
 		// if defaultQueueName is "", skip adding default queue name to the pod labels
 		if defaultQueueName != "" {
 			// for undefined configuration, am_conf will add 'root.default' to retain existing behavior
 			// if a custom name is configured for default queue, it will be used instead of root.default
-			result[constants.LabelQueueName] = defaultQueueName
+			queueName = defaultQueueName
```
FrankYang0529, line 70: We can just use `utils.GetQueueNameFromPod(pod, defaultQueueName)`.

pkg/cache/placeholder.go
FrankYang0529: Sorry for the unclear comment. I thought we can replace `constants.LabelPlaceholderFlag` with `constants.AnnotationPlaceholderFlag` in annotations, so we can just use `Labels: taskGroup.Labels` here.

chenyulin0719: Thanks @FrankYang0529's comment.

Actually I'm not sure whether we should move placeholder to annotation. 
In some code of e2e test. We use label selector to count how many placeholder pod is running.
- https://github.com/apache/yunikorn-k8shim/blob/5fe1b952bd5eb88bf21f49914216b9d49312cec0/test/e2e/gang_scheduling/gang_scheduling_test.go#L147
- https://github.com/apache/yunikorn-k8shim/blob/5fe1b952bd5eb88bf21f49914216b9d49312cec0/test/e2e/framework/helpers/k8s/k8s_utils.go#L975-L977

Label selector is native support by Kubernetes, but annotation is not.
If I moved placeholder to annotation, I have to list all pod and filter by each pod's annotation value to get placeholder pod. 
This change might be ok in e2e test, but will we require label selector for placeholder pod in the future?

pkg/admission/util.go
chenyulin0719: Updated it in new commit: 26dc73107f5206ad7cdd021e1af8635d9efe803d

pkg/cache/placeholder.go
FrankYang0529: I check the code again and found that we already set `constants.AnnotationPlaceholderFlag`.

https://github.com/apache/yunikorn-k8shim/blob/5fe1b952bd5eb88bf21f49914216b9d49312cec0/pkg/cache/placeholder.go#L53-L56

@wilfred-s, do you think we also need to set `constants.LabelPlaceholderFlag` for label selector purpose? If yes, I think we may need to add some comments here. Thank you.

chenyulin0719:
Thanks @FrankYang0529 , @craigcondit 's comment.

I'll update a new version which will keep the original labels for backforward, and copy missing annotations from corresponding labels.

chenyulin0719:
Hi @FrankYang0529, @craigcondit, @wilfred-s 

I just uploaded a newer version of commit, could I have your review?  The existing labels are still kept in admission controller and will be added to annotation if the corresponding tag are not existing in pod.   

Below is the newer change list of the whole PR:

1. Get queue name from annotation first. (GetQueueNameFromPod)
   (Also add default queue in parameter)
2. Get isStateAwareDisabled from annotation first (isStateAwareDisabled)
3. Move labels applicationId/queue in newPlaceholder to annotation (newPlaceholder)
4. Add new annotation 'yunikorn.apache.org/disable-state-aware' to replace equivalent label.
5. **Rename updateLabels to updateApplicationInfo in AdmissionController** 
   (I made some changes here, updating label/annotations to patch separately so we can easily remove labels in the future.)
6. **Rename util.updatePodLabel to util.getNewApplicationInfo**
   (I made some changes here, checking new label/annotation and the existence in pod, return the to-be update list for admission controller.)

Since this PR modified the core part of admission controller, I might need your help to review, expecially for # 5 and # 6.
Please kindly let me know if I have any misunderstadnings.

pkg/cache/placeholder.go
chenyulin0719: In the view of back-forward compatibility. I will keep PlaceholderFlag's label and annotation in placeholder pod.

For the same reason, I should move applicationId/queue back to labels in placeholder pod, 
@wilfred-s Do I have the correct understanding about the newPlaceHolder change.

chenyulin0719:
Hi @FrankYang0529 ,  after dived into the error in CI (v1.24.15)
I found there have a racing issue in e2e test.  I opened a Jira issue here: 
-  [YUNIKORN-1998 Stale AdmissionControllerConf was used in e2e test](https://issues.apache.org/jira/browse/YUNIKORN-1998?jql=project%20%3D%20YUNIKORN%20ORDER%20BY%20key%20DESC%2C%20priority%20DESC%2C%20updated%20DESC)

Could you help to re-trigger CI? I don't think it will happen every time

pkg/cache/placeholder.go
chenyulin0719: To minimum the change of this PR and keep back-forward compatibility.
Just add label applicationId/queue back to placeholder pod. 26dc73107f5206ad7cdd021e1af8635d9efe803d

chenyulin0719:
Add screenshot to make this PR more readable.
Before vs After result.  (Pod with existing applicationId, queue labels.)

YUNIKORN-1351  Before vs After (Pod with labels)

chenyulin0719:
Before vs After result. (Pod with no label.)

YUNIKORN-1351  Before vs After (Pod without labels)

pkg/admission/util.go
```diff
@@ -29,42 +29,79 @@ import (
 	"github.com/apache/yunikorn-k8shim/pkg/log"
 )
 
-func updatePodLabel(pod *v1.Pod, namespace string, generateUniqueAppIds bool, defaultQueueName string) map[string]string {
-	existingLabels := pod.Labels
-	result := make(map[string]string)
-	for k, v := range existingLabels {
-		result[k] = v
-	}
+func getNewApplicationInfo(pod *v1.Pod, namespace string, generateUniqueAppIds bool, defaultQueueName string) (map[string]string, map[string]string) {
```
pbacsko, line 32: In this function, I think it's useful to issue a warning to the user that labels will be deprecated in a later release, so they must adapt their app submission logic to this. It's very easy to forget about this and we must reduce the surprise factor.

What I'd do: return an extra bool from `getApplicationIDFromPod()`, `getDisableStateAwareFromPod()`, `utils.GetQueueNameFromPod()` to indicate whether a source was a label. If there's at least one of them is true, log something useful:
```
if appIdLabel || stateAwareFromLabel || queueNameFromLabel {
  log.Log(log.Admission).Warn("Pod contains label for Yunikorn configuration. This is deprecated and will be 
  ignored in a future release. Please migrate to annotation-based configuration.",
  zap.Bool("appId from label", appIdLabel),
  zap.Bool("stateAware from label", stateAwareFromLabel),
  zap.Bool("queue name from label", queueNameFromLabel))
}
```

pbacsko (changes requested):
First round from me.

pkg/admission/admission_controller.go
```diff
@@ -415,21 +415,71 @@ func (c *AdmissionController) updatePreemptionInfo(pod *v1.Pod, patch []common.P
...
+		existingAnnotations := pod.Annotations
+		annotations = utils.MergeMaps(existingAnnotations, annotations)
+		patch = append(patch, common.PatchOperation{
+			Op:    "add",
+			Path:  "/metadata/annotations",
+			Value: annotations,
+		})
+	}
 	return patch
 }
```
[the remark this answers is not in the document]
chenyulin0719: Combined updateLabelPatch() and updateAnnotationPatch() to updatePatch() (f48e089baf9f0e85a60d10e9f8df38870a8af6ea)

I add an ifelse to control how to fetch existingPodValues.
- constants.LabelPatchPath
- constants.AnnotationPatchPath

Rename
- First existingLabel/existingAnnotation to existingPatchValues
- Second existingLabel/existingAnnotation to existingPodValues

Also rewrite test to enhance readibility.

pkg/admission/admission_controller.go
```diff
@@ -415,21 +415,71 @@ func (c *AdmissionController) updatePreemptionInfo(pod *v1.Pod, patch []common.P
...
+	patch = updateLabelPatch(pod, newLabels, patch)
+	patch = updateAnnotationPatch(pod, newAnnotations, patch)
+
+	return patch
+}
 
+func updateLabelPatch(pod *v1.Pod, labels map[string]string, patch []common.PatchOperation) []common.PatchOperation {
+	// if found an existing patch on labels, add new labels to it
+	for _, p := range patch {
+		if p.Op == "add" && p.Path == "/metadata/labels" {
```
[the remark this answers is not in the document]
chenyulin0719: Removed  "/metadata/labels" and "/metadata/annotations" and other literal strings, 
create below constants for patching:
- const AddPatchOp = "add"
- const AnnotationPatchPath = "/metadata/annotations"
- const LabelPatchPath = "/metadata/labels"
- const SchedulerNamePatchPath = "/spec/schedulerName"

([code link](https://github.com/apache/yunikorn-k8shim/pull/677/files#diff-0501a6e1a48b49ee780c6183edb77b222c1e4f5195e3521170c8847380e69d9dR116-R119))

pkg/admission/admission_controller.go
```diff
@@ -415,21 +415,71 @@ func (c *AdmissionController) updatePreemptionInfo(pod *v1.Pod, patch []common.P
 	return patch
 }
 
-func (c *AdmissionController) updateLabels(namespace string, pod *v1.Pod, patch []common.PatchOperation) []common.PatchOperation {
-	log.Log(log.Admission).Info("updating pod labels",
+func (c *AdmissionController) updateApplicationInfo(namespace string, pod *v1.Pod, patch []common.PatchOperation) []common.PatchOperation {
+	log.Log(log.Admission).Info("updating pod application annotations",
```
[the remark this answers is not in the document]
chenyulin0719: Change to "updating pod application labels and annotations" ([code link](https://github.com/apache/yunikorn-k8shim/pull/677/files#diff-a2a009dc3a1a3d2c466a066cf9a12a1321b097476c6c442b827ed295680361bfR419))

pkg/admission/util.go
chenyulin0719: Add deprecated warning if applicationInfo is from labels (cf71eeff4cef7f432cc8778de11dfb14202efb91)

- Return extra bool 'isFromLabel' as suggested

chenyulin0719:
Thanks @pbacsko's review,  I've updated a newer version based on the feedbacks.
The e2e test passed in my local env. Appreciate any advice.

chenyulin0719:
Hi @pbacsko, sorry for late update.  I just rebase the commits to solve some conficts.

pbacsko:
@chenyulin0719 thanks, I'll take a look shortly.

pkg/admission/util.go
```diff
@@ -29,42 +29,86 @@ import (
...
+
+	appID, isAppIdFromLabel := getApplicationIDFromPod(pod)
+	disableStateAware, isDisableStateAwareFromLabel := getDisableStateAwareFromPod(pod)
+
+	// for undefined configuration, am_conf will add 'root.default' as default queue name
+	// if a custom name is configured for default queue, it will be used instead of root.default
+	queueName, isQueueNameFromLabel := utils.GetQueueNameFromPod(pod, defaultQueueName)
+
+	if isAppIdFromLabel || isDisableStateAwareFromLabel || isQueueNameFromLabel {
+		log.Log(log.Admission).Warn("Pod contains label for Yunikorn configuration. This is deprecated and will be ignored in a future release. Please migrate to annotation-based configuration.",
```
pbacsko, line 44: I've already talked to @craigcondit about this, we might not be in a position to ever remove this, it's going to be too dangerous. Also, certain settings actually should be specified as labels because that allows pods to be grouped/selected based on certain criteria. Let's think about it.

chenyulin0719: I agree with the benefits of using label in terms of grouping and selecting.
Deprecate it might result in users unable to implement alternative solution.

Regarding this PR:
1. always prefer annotation above label.
2. Move labels to annotation, but keep existing labels before deprecation.

Before making any change, I think we need to align the scope again.
Below are my questions if we are not going to deprecate the labels:
1. Will AM patch missing attributes to label only?  Or patch to label and annotation both?
-> I think for  'applicationID' and 'queue', we should patch to both.
2. How about the existing label "disableStateAware"?
-> I think we can just change it to annotation because it's only be set in AM.  
-> This label could be deprecated. (Or suggest not using it)
3. If we have a different applicationID set in label/annotation,  do we still prefer annotation above label.
-> I think the answer is yes for consistency, we must guide user not to set different value if they require labels for selecting/grouping.
4. If only annotations were set in pod, should AM patch it(app-id, queue) to label?
-> Yes, align with my answer of question 1
5. If only labels were set in pod, should AM patch it(applicationId, queue) to annoataion?
-> Yes, align with my answer of question 1

My summary of new version:
1. Always prefer annotation above labels
2. AM ensure applicationID/queue existing in label and annotation.
3. Change label "disableStateAware" to annotation.

@pegasas, do you have any feedback?

chenyulin0719: Also needs @wilfred-s 's review too.

pkg/admission/util.go
craigcondit: I don't think we can even deprecate the `applicationId` label -- it's far too entrenched in the existing ecosystem (and arguably it should be a label anyway to allow querying pods by applicationId). I think our best best is simply to use the annotation if it exists, and fallback to the label if it does not.

pkg/admission/util.go
chenyulin0719: Sorry, I just pinged the wrong person.

@pbacsko,@craigcondit , do you have any feedback for the new version AM behaviror?

pkg/admission/util.go
chenyulin0719: It make sense to me, let's discuss the expecting AM behavior in below discussion.

pkg/admission/util.go
chenyulin0719: @pbacsko, @craigcondit ,
After reconsideration, if deprecated labels are no longer needed, I can revert most of the AM changes to keep this PR simple. We can maintain the previous UpdateLabel behavior for AM. Just make sure to always get annotation before labels when accessing the value.

Will update a version with this idea, please let me know if you have any concern.

craigcondit: I'd like to get @wilfred-s to weigh in on this one.

chenyulin0719:
Close it as new policy for labels and annotations is defined in [YUNIKORN-1351](https://issues.apache.org/jira/browse/YUNIKORN-1351).

PR #683, opened 2023-09-28
[YUNIKORN-1998] Stale AdmissionControllerConf was used in e2e test

chenyulin0719:
### What is this PR for?
The AdmissionControllerConf update happen asynchronously. There's no guarantee that new configuration has been applied after update. We introduced 5 secs delay between updating the configmap and submitting new sleep pods in e2e tests to avoid the racing issue.

Because only the e2e tests for the user_group_limit require the admission controller's config to be up-to-date, I only added a 5-second delay here to avoid unnecessary delays in other tests.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
NA

### What is the Jira issue?
[[YUNIKORN-1998] Stale AdmissionControllerConf was used in e2e test](https://issues.apache.org/jira/browse/YUNIKORN-1998)

Can add 1 sec delay in admission controller to reproduce the error. (Please refer to my desc in YUNIKORN-1998 )
Run the e2e test user_group_limit to validate it. 

NA

### Questions:
NA


pbacsko:
@chenyulin0719 Back then I added a solution to this problem for the admission controller tests:
```
		configMap.Data[amConf.AMAccessControlTrustControllers] = "false"
		ginkgo.By("Update configmap")
		stopChan := make(chan struct{})
		eventHandler := &EventHandler{updateCh: make(chan struct{})}
		err = kubeClient.StartConfigMapInformer(configmanager.YuniKornTestConfig.YkNamespace, stopChan, eventHandler)
		defer close(stopChan)
		gomega.Ω(err).ShouldNot(gomega.HaveOccurred())
		_, err = kubeClient.UpdateConfigMap(configMap, configmanager.YuniKornTestConfig.YkNamespace)
		gomega.Ω(err).ShouldNot(gomega.HaveOccurred())
		updateOk := eventHandler.WaitForUpdate(30 * time.Second)
```

It's still not perfect because we can receive the update earlier than the controller, but the tests are fairly stable, I haven't seen any issues. Would you mind trying this out?

chenyulin0719:
Hi @pbacsko, 
In my test, the infomer delay is only in milliseconds, so the “1 extra second to cover our backs”  part should be sufficient in most situations.  The implementation of a local informer could serve as a buffer if there is consistent and long delay (>1s) between K8S and the informer. I believe it is a better option to have it too.

Since we don’t have a perfect solution for now, it is better to align the solution so we can easily change it in future,  I will modify this PR to follow current implementation in admission controller tests.

Thanks for your comment. :)

wilfred-s:
Lets keep the approach and solution for all these wait for config changes the same. Since we already have the informer with the wait in other tests it is better to use it here also.

wilfred-s:
@pbacsko please check again so we can add it to YuniKorn 1.4

pbacsko (approved):
+1

PR #684, opened 2023-09-30
[YUNIKORN-2009] ensure placeholder pod's limits carry the same amount of limits regardless of resource type

Huang-Wei:
### What is this PR for?

Ensure placeholder pod's `limits` carry the same amount as `request` regardless of resource type. Otherwise, placeholder pod cannot be created on environments that LimitRange is in place, and hence gang-scheduling feature breaks.

### What type of PR is it?
* [x] - Bug Fix
* [x] - Improvement

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-2009

### Questions:


wilfred-s (approved):
LGTM

PR #686, opened 2023-10-05
[YUNIKORN-2014] update yunikorn-core version

FrankYang0529:
### What is this PR for?

Update yunikorn-core version to include extra config response.

ref: https://github.com/apache/yunikorn-core/pull/652

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2014


FrankYang0529:
Close the PR since we will update core version before release.

PR #688, opened 2023-10-12
[YUNIKORN-2029] remove deprecated function

FrankYang0529:
### What is this PR for?

Remove deprecated functions

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2029

### DIFF
--- a/test/e2e/framework/helpers/common/utils.go
+++ b/test/e2e/framework/helpers/common/utils.go
@@ -22,14 +22,12 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"math/rand"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"reflect"
 	"strings"
-	"time"
 
 	"github.com/google/uuid"
 	"github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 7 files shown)


test/e2e/framework/helpers/common/utils.go
```diff
@@ -116,7 +114,6 @@ func GetUUID() string {
 
 func RandSeq(n int) string {
 	var letters = []rune("abcdefghijklmnopqrstuvwxyz0123456789")
-	rand.Seed(time.Now().UnixNano())
```
wilfred-s, line 119: Not sure why this one is removed. The seeding can stay as far I can tell.

wilfred-s (changes requested):

test/e2e/framework/helpers/common/utils.go
wilfred-s: This is no longer needed in Go 1.20, same discussion I had with Craig on the placeholder name generation. Yes we can remove the Seed() call.

wilfred-s (approved):
LGTM

wilfred-s:
@FrankYang0529 can you check why the e2e tests are failing? The code changes look good but we cannot commit will all e2e tests broken.

PR #691, opened 2023-10-18
[YUNIKORN-2049] Fix incorrect placement rule examples

brandboat:
### What is this PR for?
There are some useful placement rule examples under https://github.com/apache/yunikorn-k8shim/tree/master/deployments/examples/placements

while there are some incorrect settings in them:
- `queue.yaml` should be `queues.yaml`
- label `yunikorn.apache.org/username` is legacy and should be switched to annotation `yunikorn.apache.org/user.info`

Correct me if I'm wrong, the backward compatibility for the label `yunikorn.apache.org/username` seems broken. The placement rules examples in the link mentioned will not work with the label, but will only be effective with the annotation (i.e., `yunikorn.apache.org/user.info`).

### What type of PR is it?
* [x] - Documentation

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2049

N/A

N/A

### Questions:
N/A

### DIFF
--- a/deployments/examples/placements/README.md
+++ b/deployments/examples/placements/README.md
@@ -28,12 +28,12 @@ The sleep pod is described in the example file and the partition is provided in 
 
 ## How to adopt the queues in `config.yaml`
 Before deploying the pods, make sure that the data in the `yunikorn-configs` configmap is correct.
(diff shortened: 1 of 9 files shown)


wilfred-s (approved):
LGTM

PR #697, opened 2023-10-23
[YUNIKORN-2068] Fix deadlock when evaluating preemption candidates

craigcondit:
### What is this PR for?
Fixes a deadlock that occurs during node evaluation for preemption that causes the scheduler cache read lock to be acquired twice.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2068

### Questions:


wilfred-s (approved):
LGTM
Simple enough change, please create a e2e test for this case.

PR #707, opened 2023-10-27
[YUNIKORN-2082] Make autogenerated applicationIDs reproducible

craigcondit:
### What is this PR for?
Moved the admission-controller private function generateAppID() to the shim utils package and renamed to GenerateApplicationID() to allow for re-use by the shim. Updated implementation to use the pod UID instead of a random UUID so that the same identifier will always be created for the same Pod.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2082

Updated unit tests.

### Questions:

### DIFF
--- a/pkg/admission/util.go
+++ b/pkg/admission/util.go
@@ -19,10 +19,8 @@
 package admission
 
 import (
-	"fmt"
 	"reflect"
 
-	"github.com/google/uuid"
 	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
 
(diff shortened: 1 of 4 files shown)


pbacsko (approved):
+1 LGTM

PR #708, opened 2023-10-27
[YUNIKORN-2083] Shim: Handle missing applicationID cleanly in standard mode

craigcondit:
### What is this PR for?
If a Pod with `schedulerName: yunikorn` is encountered with missing applicationID metadata, generate an applicationID using the same algorithm as the admission controller would have. This allows these Pods to be scheduled successfully.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2083

Updated unit tests to handle new logic (both plugin / standard mode). Also manually created a Pod with the necessary conditions and verified that it scheduled properly and was assigned the appropriate applicationID by YuniKorn.

### Questions:

### DIFF
--- a/pkg/common/utils/utils.go
+++ b/pkg/common/utils/utils.go
@@ -43,6 +43,16 @@ import (
 
 const userInfoKey = siCommon.DomainYuniKorn + "user.info"
 
+var pluginMode bool
+
+func SetPluginMode(value bool) {
+	pluginMode = value
+}
+
+func IsPluginMode() bool {
+	return pluginMode
+}
+
 func Convert2Pod(obj interface{}) (*v1.Pod, error) {
 	pod, ok := obj.(*v1.Pod)
 	if !ok {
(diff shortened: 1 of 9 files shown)


pkg/conf/schedulerconf.go
```diff
@@ -73,19 +74,24 @@ const (
 	CMKubeQPS   = PrefixKubernetes + "qps"
 	CMKubeBurst = PrefixKubernetes + "burst"
 
+	// admissioncontroller
+	PrefixAMFiltering               = PrefixAdmissionController + "filtering."
+	AMFilteringGenerateUniqueAppIds = PrefixAMFiltering + "generateUniqueAppId"
```
craigcondit, line 79: NOTE: This is needed as the specific algorithm used for generating the appID is controlled by an admission-controller specific property. We use that here as well to maintain backwards compatibility. The value is duplicated form am_conf.go as we would create a circular package reference otherwise.

pkg/common/utils/utils.go
```diff
@@ -43,6 +43,16 @@ import (
 
 const userInfoKey = siCommon.DomainYuniKorn + "user.info"
 
+var pluginMode bool
+
+func SetPluginMode(value bool) {
```
craigcondit, line 48: NOTE: Moved plugin flag here (from context) so that it can be used by utility functions. This effectively makes it global instead of per-context (which is fine since it's always set on startup).

pkg/conf/schedulerconf.go
```diff
@@ -208,6 +216,7 @@ func handleNonReloadableConfig(old *SchedulerConf, new *SchedulerConf) {
 	checkNonReloadableBool(CMSvcDisableGangScheduling, &old.DisableGangScheduling, &new.DisableGangScheduling)
 	checkNonReloadableString(CMSvcPlaceholderImage, &old.PlaceHolderImage, &new.PlaceHolderImage)
 	checkNonReloadableString(CMSvcNodeInstanceTypeNodeLabelKey, &old.InstanceTypeNodeLabelKey, &new.InstanceTypeNodeLabelKey)
+	checkNonReloadableBool(AMFilteringGenerateUniqueAppIds, &old.GenerateUniqueAppIds, &new.GenerateUniqueAppIds)
```
craigcondit, line 219: NOTE: We mark the generateUniqueAppIds property as non-reloadable since changing this value would cause the computed applicationIDs of existing Pods to change. This would completely mess up internal state tracking.

pbacsko (approved):
+1 LGTM

pbacsko (approved):
+1

PR #713, opened 2023-10-31
[YUNIKORN-2094] Add install kind cluster option to run-e2e-test script

doupache:
### What is this PR for?
Currently, the run-e2e-test script supports only two actions: cleanup and test.

It's common for us to want to test specific e2e tests instead of the entire set of e2e tests.

So, let's add a new action to only install the kind cluster. Then, we can run a particular e2e test for a specific folder using:

e.g.

```shell 
make e2e_test E2E_TEST=user_group_limit
```

For even finer granularity:

e.g.

```shell 
ginkgo run -r -v --focus "Verify_maxapplications_with_a_specific_group_limit" \
-- -yk-namespace "yunikorn" \
-kube-config "$HOME/.kube/config"
```

 ### What type of PR is it?
* [x] - Improvement

### What is the Jira issue?
[YUNIKORN-2094](https://issues.apache.org/jira/browse/YUNIKORN-2094)

### DIFF
--- a/scripts/run-e2e-tests.sh
+++ b/scripts/run-e2e-tests.sh
@@ -243,13 +243,16 @@ echo "  web image          : ${WEBTEST_IMAGE}"
 check_opt "action" "${ACTION}"
 check_opt "kind-cluster-name" "${CLUSTER_NAME}"
 
-# this script only supports 2 actions
+# this script only supports 3 actions
 #   1) test
 #     - install a K8s cluster with kind
 #     - install latest yunikorn
 #     - run e2e tests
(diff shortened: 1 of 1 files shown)


doupache:
@wusamzong @FrankYang0529 @pbacsko Please take a look 🙏

I think the duplicated code within the "install" and "test" sections of run-e2e-test is acceptable. However, I am also happy to removing the duplicated code.

wilfred-s (approved):
LGTM, I don't think we need to worry about the duplicated code lines.

PR #715, opened 2023-11-01
[YUNIKORN-2098] Change go lint SHA detection (following)

doupache:
### What is this PR for?
Currently, we will always use the "ORIGIN/HEAD" ref.(In Github Action env) Fallback to "HEAD^" when "ORIGIN/HEAD" doesn't exist.
Also print what git ref the ci have. 

I get `fatal: Needed a single revision` on my fork github action ci.
https://github.com/doupache/yunikorn-k8shim/actions/runs/6713607756/job/18245448950
image

image
image

CI in the forked repo doesn't seem to have the "ORIGIN/HEAD" ref.
I keep getting the 'fatal: Needed a single revision' error, similar to [YUNIKORN-285](https://issues.apache.org/jira/browse/YUNIKORN-285)

### What type of PR is it?
* [x] - Bug Fix

### What is the Jira issue?
[YUNIKORN-2098](https://issues.apache.org/jira/browse/YUNIKORN-2098)

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -14,6 +14,8 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+      - name: Fetch all branches and tags
+        run: git fetch --all
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
@@ -46,6 +48,8 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+      - name: Fetch all branches and tags
+        run: git fetch --all
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Checkout source code
         uses: actions/checkout@v3
         with:
-          fetch-depth: 2
+          fetch-depth: 0
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
@@ -45,7 +45,7 @@ jobs:
       - name: Checkout source code
         uses: actions/checkout@v3
         with:
-          fetch-depth: 2
+          fetch-depth: 0
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
--- a/.github/workflows/push-master.yml
+++ b/.github/workflows/push-master.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Checkout source code
         uses: actions/checkout@v3
         with:
-          fetch-depth: 2
+          fetch-depth: 0
       - name: Set up Go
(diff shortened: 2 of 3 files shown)


doupache:
@wilfred-s Please take a look ! 

This is the following PR of https://github.com/apache/yunikorn-k8shim/pull/210/

targetoee:
I have the same problem in my forked project.
link: [https://github.com/targetoee/yunikorn-k8shim/actions/runs/6716879779/job/18253781242](https://github.com/targetoee/yunikorn-k8shim/actions/runs/6716879779/job/18253781242)

doupache:
Pass on my fork with branch [doupache:YUNIKORN-2098](https://github.com/doupache/yunikorn-k8shim/tree/YUNIKORN-2098)

https://github.com/doupache/yunikorn-k8shim/actions/runs/6714213847 

It seems there is another issue with the Pull Request CI environment

doupache:
In [YUNIKORN-285](https://issues.apache.org/jira/browse/YUNIKORN-285), we previously included git symbolic-ref -q HEAD because Travis CI often encountered a detached head situation. However, now that we have switched to GitHub Actions, we can utilize the checkout@v3 action, which efficiently fetches all Git references.

Consequently, we can remove `git symbolic-ref -q HEAD` Instead, we can initially set `REV='origin/HEAD'` If git rev-parse fails, we can fall back to `REV=HEAD^`

Regarding the modification from `fetch-depth: 2` to `fetch-depth: 0`, I have compared the GitHub Action times before this pull request and with the master. The duration of the git checkout action remained around 2 seconds. Therefore, I believe this change is acceptable, as it does not introduce significant overhead.

@wilfred-s  I think this PR is ready for review.

.github/workflows/pre-commit.yml
```diff
@@ -45,7 +47,9 @@ jobs:
       - name: Checkout source code
         uses: actions/checkout@v3
         with:
-          fetch-depth: 2
+          fetch-depth: 0
+      - name: Fetch all branches and tags
+        run: git fetch --all
```
wilfred-s, line 52: Do not think this is needed, the fetch-depth or any history to build for the e2e tests only.

.github/workflows/push-master.yml
```diff
@@ -12,7 +12,9 @@ jobs:
       - name: Checkout source code
         uses: actions/checkout@v3
         with:
-          fetch-depth: 2
+          fetch-depth: 0
+      - name: Fetch all branches and tags
+        run: git fetch --all
```
wilfred-s, line 17: Do not think this is needed, the fetch-depth is a left over of the old pre-split flow and we should not need it here for the code coverage we run.

wilfred-s:
With the cleanup for lint issues we no longer need SHA detection in our lint runs. It was removed as part of an earlier change.

PR #733, opened 2023-11-22
[YUNIKORN-2135] Add integration test code coverage

FrankYang0529:
### What is this PR for?

If `ENABLE_GO_COVER_DIR` is `TRUE`:
* Add cover flags to go build.
* Use root user to build docker images.
* Add `uninstall_yunikorn` and `copy_go_cover_dir` to run-e2e-tests.sh.

### What type of PR is it?
* [X] - Improvement

### Todos
* [ ] Wait for https://github.com/apache/yunikorn-release/pull/164.
* [x] Update workflow to send codecov data

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2135

### DIFF
--- a/scripts/run-e2e-tests.sh
+++ b/scripts/run-e2e-tests.sh
@@ -87,7 +87,8 @@ function install_cluster() {
   if [ "${GIT_CLONE}" = "true" ]; then
     check_cmd "git"
     rm -rf ./build/yunikorn-release
-    git clone --depth 1 https://github.com/apache/yunikorn-release.git ./build/yunikorn-release
+    # TODO: update following branch if PR in yunikorn-release is merged
+    git clone --depth 1 https://github.com/FrankYang0529/yunikorn-release.git -b YUNIKORN-2135 ./build/yunikorn-release
   fi
   if [ ! -d "${CHART_PATH}" ]; then
     exit_on_error "helm charts not found in path: ${CHART_PATH}"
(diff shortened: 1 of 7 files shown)


pbacsko:
@FrankYang0529 could you resolve the merge conflicts?

craigcondit:
@FrankYang0529 can we write the logs to a directory under `/tmp` in the container? I'm not a fan of a top-level dir for this...

scripts/run-e2e-tests.sh
```diff
@@ -87,7 +87,8 @@ function install_cluster() {
   if [ "${GIT_CLONE}" = "true" ]; then
     check_cmd "git"
     rm -rf ./build/yunikorn-release
-    git clone --depth 1 https://github.com/apache/yunikorn-release.git ./build/yunikorn-release
+    # TODO: update following branch if PR in yunikorn-release is merged
+    git clone --depth 1 https://github.com/FrankYang0529/yunikorn-release.git -b YUNIKORN-2135 ./build/yunikorn-release
```
craigcondit, line 91: As indicated in the yunikorn-release repo PR, I think we should find a way to deploy this without polluting the upstream helm charts.

craigcondit: Here's the approach we should use: https://helm.sh/docs/topics/advanced/#post-rendering

Helm supports a post-renderer that can be used to update the generated output. We can use this in the e2e test script during install to "patch" the installed charts with what is required here. An ideal post-renderer would be kustomize (see https://kustomize.io/). An example using the two together can be found here: https://github.com/thomastaylor312/advanced-helm-demos/tree/master/post-render

We can setup a kustomize script to patch only the objects we need so that the original helm charts can stay pristine. All that should be required is an ENV var and build-mount for /tmp/coverage (or something like it).

pbacsko (changes requested):
@FrankYang0529 could you rebase the changes please? I'll come back to this once it can be merged.

PR #736, opened 2023-11-24
[YUNIKORN-2184] Fix goconst linter issues

targetoee:
### What is this PR for?
This PR focus on fix linter issues about `goconst` warning. Some const are set for better reuse.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
This PR only fixes the parts which are pointed out by the linter. 
While in `application_test.go`, the naming of variable names doesn't have a unified format. 
It would be better to reformat that, and one way to achieve this is by setting common variables in `util_test.go` and reuse them.

### What is the Jira issue?
[YUNIKORN-2184](https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-2184)

It's been test using `./tools/golanci-lint run ./... | grep goconst` to checkout all the goconst waring are fixed.

N/A

### Questions:
N/A

### DIFF
--- a/pkg/admission/webhook_manager.go
+++ b/pkg/admission/webhook_manager.go
@@ -50,6 +50,7 @@ const (
 	caCert2Path       = "cacert2.pem"
 	caPrivateKey1Path = "cakey1.pem"
 	caPrivateKey2Path = "cakey2.pem"
+	webhookLabel      = "yunikorn"
 )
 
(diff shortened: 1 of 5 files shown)


FrankYang0529 (approved):
+1. Thank you

PR #740, opened 2023-12-01
[YUNIKORN-2211] Replace Allocation uuid with allocationID

manirajv06:
### What is this PR for?
Replace all `uuid` references with `allocationID `

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2211

Using existing unit tests

### Questions:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -523,5 +523,6 @@ arch:
 e2e_test: tools
 	@echo "running e2e tests"
 	cd ./test/e2e && \
-	ginkgo $(E2E_TEST) -v -keep-going -- -yk-namespace "yunikorn" -kube-config $(KUBECONFIG) state_aware_app_scheduling
+	ginkgo -r -v state_aware_app_scheduling -- -yk-namespace yunikorn -kube-config $(KUBECONFIG)
+	#ginkgo $(E2E_TEST) -v -keep-going -- -yk-namespace "yunikorn" -kube-config $(KUBECONFIG) state_aware_app_scheduling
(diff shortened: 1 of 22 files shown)


pbacsko:
Need to update `go.mod`.

pkg/common/si_helper_test.go
```diff
@@ -1,19 +1,19 @@
 /*
- Licensed to the Apache Software Foundation (ASF) under one
```
chia7712, line 2: Those changes are unrelated to this PR. Should we address it in another ticket? I guess there are other files having similar issue.

chia7712 (commented):
nice patch and just one trivial comment left.

pkg/common/si_helper_test.go
manirajv06: unintended changes because of IDE formatting (I think so). Please check the reason if possible so that we can fix it next time. For now, let it go through for this fix.

chia7712:
@manirajv06 Please take a look at failed test. overall +1

Makefile
```diff
@@ -523,4 +523,6 @@ arch:
 e2e_test: tools
 	@echo "running e2e tests"
 	cd ./test/e2e && \
-	ginkgo -r $(E2E_TEST) -v -keep-going -- -yk-namespace "yunikorn" -kube-config $(KUBECONFIG)
```
chia7712, line 526: I guess those changes are used to run test locally. Please revert it before merging :)

wilfred-s (approved):
LGTM +1

PR #742, opened 2023-12-03
[YUNIKORN-2229] Fix GetQueueNameFromPod default use root.default

brandboat:
### What is this PR for?
The default queue name for the admission controller is `root.default` whereas in the GetQueueNameFromPod function on the shim side, the default queue name used is `root.sandbox`. It is necessary to ensure alignment between these two default queue names.

### What type of PR is it?
* [x] - Bug Fix

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2229

covered by existing unit test

N/A

### Questions:
N/A


pbacsko (approved):
+1

PR #747, opened 2023-12-07
[YUNIKORN-2205] remove the warning of processing nonexistent "namespa…

Haser0305:
### What is this PR for?
The warning message is printed when "namespace.guaranteed" does not exist. This PR checking if the "namespaceGuaranteed" value is empty and returning early to avoid the warning message. This ensures that the warning is not displayed when the expected behavior occurs. 

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2205

### DIFF
--- a/pkg/common/utils/utils.go
+++ b/pkg/common/utils/utils.go
@@ -179,19 +179,18 @@ func PodUnderCondition(pod *v1.Pod, condition *v1.PodCondition) bool {
 // get namespace guaranteed resource from namespace annotation
 func GetNamespaceGuaranteedFromAnnotation(namespaceObj *v1.Namespace) *si.Resource {
 	// retrieve guaranteed resource info from annotations
-	namespaceGuaranteed := GetNameSpaceAnnotationValue(namespaceObj, constants.NamespaceGuaranteed)
-	if namespaceGuaranteed == "" {
-		return nil
+	if guaranteed := GetNameSpaceAnnotationValue(namespaceObj, constants.NamespaceGuaranteed); guaranteed != "" {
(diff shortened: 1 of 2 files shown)


chia7712 (approved):
LGTM

pbacsko (changes requested):
Nit: please add coverage in `utils_test.go`

pkg/common/utils/utils.go
```diff
@@ -180,6 +180,9 @@ func PodUnderCondition(pod *v1.Pod, condition *v1.PodCondition) bool {
 func GetNamespaceGuaranteedFromAnnotation(namespaceObj *v1.Namespace) *si.Resource {
 	// retrieve guaranteed resource info from annotations
 	namespaceGuaranteed := GetNameSpaceAnnotationValue(namespaceObj, constants.NamespaceGuaranteed)
```
pbacsko, line 182: I believe this is more Go-like:
```
if guaranteed := GetNameSpaceAnnotationValue(namespaceObj, constants.NamespaceGuaranteed); guaranteed != "" {
  ...
}
return nil
```

No need to name it "namespaceGuaranteed", it's pretty obvious from the context. Prefer shorter variables if possible.

Haser0305: Thanks for reviewing my code and the helpful recommendation.  I'll make that change to implement your feedback.

zhuqi-lucas (approved):
LGTM

PR #748, opened 2023-12-08
[YUNIKORN-2230] Placement rule does not behave as expected

brandboat:
### What is this PR for?
Fix issue with the placement rules, specifically the one where the provided rule (when `created` is true) is applied before the tag rule, is not functioning as intended. The root cause lies in the GetQueueNameFromPod function in the shim, which automatically adds a default queue name to a pod if none is specified. As a result, the provided rule (when `created` is true) is consistently applied, preventing the tag rule from being triggered.

### What type of PR is it?
* [x] - Bug Fix

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2230

covered by e2e test

N/A

### Questions:
* [x] - There is breaking changes for older versions. This Pull Request removed the default queue name on the shim side. If a user is using YuniKorn without enabling the Admission Controller (which adds a default queue name), pods that do not provide a queue name will be rejected.

### DIFF
--- a/pkg/common/utils/utils.go
+++ b/pkg/common/utils/utils.go
@@ -102,13 +102,12 @@ func IsAssignedPod(pod *v1.Pod) bool {
 }
 
 func GetQueueNameFromPod(pod *v1.Pod) string {
-	queueName := constants.ApplicationDefaultQueue
-	if an := GetPodLabelValue(pod, constants.LabelQueueName); an != "" {
-		queueName = an
-	} else if qu := GetPodAnnotationValue(pod, constants.AnnotationQueueName); qu != "" {
-		queueName = qu
+	if queueName := GetPodLabelValue(pod, constants.LabelQueueName); queueName != "" {
+		return queueName
+	} else if queueName := GetPodAnnotationValue(pod, constants.AnnotationQueueName); queueName != "" {
+		return queueName
 	}
-	return queueName
+	return ""
 }
 
 // GenerateApplicationID generates an appID based on the namespace value
--- a/pkg/common/utils/utils_test.go
+++ b/pkg/common/utils/utils_test.go
@@ -832,7 +832,7 @@ func TestGetQueueNameFromPod(t *testing.T) {
 			pod: &v1.Pod{
 				ObjectMeta: metav1.ObjectMeta{},
 			},
-			expectedQueue: constants.ApplicationDefaultQueue,
+			expectedQueue: "",
 		},
 	}
 
--- a/test/e2e/placement_rules/placement_rules_suite_test.go
+++ b/test/e2e/placement_rules/placement_rules_suite_test.go
@@ -1,0 +1,63 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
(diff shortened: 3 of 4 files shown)


wilfred-s:
This is more a design and behavioural question than a code review.
Should we not have a similar behaviour as we have on the admission controller and allow specifying the default queue? In other words should we re-use the setting for the default queue in the K8shim. We do something similar also for the application ID generation.

brandboat:
> This is more a design and behavioural question than a code review.
Should we not have a similar behaviour as we have on the admission controller and allow specifying the default queue? In other words should we re-use the setting for the default queue in the K8shim. We do something similar also for the application ID generation.

Thank's for the comment! Indeed, that is a better way and doing so won't break anything. I'll make the setting `admissionController.filtering.defaultQueue` to also determine whether K8shim should add the default queue name or not. And we should emphasize in document that this setting still takes effect even without the admission controller, I'll open another pr to do that after this one is merged.

brandboat:
I also have another minor question: should we consider renaming these two settings (admissionController.filtering.defaultQueueName and admissionController.filtering.generateUniqueAppId)? Since they not only control the admission controller but also influence K8Shim, perhaps a more descriptive name would be appropriate.

brandboat:
gentle ping @wilfred-s 
I'm trying to follow the logic of `GenerateUniqueAppIds` and add a new variable `DefaultQueueName` in schedulerconf.go. However, I encountered a challenge. I believe DefaultQueueName must be placed in handleNonReloadableConfig to make it a configuration that takes effect only after restarting YuniKorn. Without this, changes to DefaultQueueName might lead to inconsistencies when getAppMetadata returns the queueName in the existing pod, as it could change due to configmap modifications in DefaultQueueName.

However, if I set DefaultQueueName as non-reloadable, changing the default queue name to `""` would require restarting the entire YuniKorn. Otherwise, the K8shim might still return the queue name as `root.default` (the current default value of DefaultQueueName). I think this could make things more complex, not unsure if there are better suggestions or approaches... thank you.

chia7712:
> if I set DefaultQueueName as non-reloadable, changing the default queue name to "" would require restarting the entire YuniKorn

pardon me, is this a potential issue? Are there any reasons that users have to change the default queue name by re-reconfiguring?

wilfred-s:
The admission controller handles it as a reloadable value. We can do either. However I think that the value should be reloadable as we can change the queue config and placement rules on the fly also.

If you do not allow it to be reloadable a queue config change could be applied by the default queue would not change causing rejects etc. Keeping them both reloadable would be better.

pkg/common/utils/utils.go
```diff
@@ -102,13 +102,12 @@ func IsAssignedPod(pod *v1.Pod) bool {
...
 
 func GetQueueNameFromPod(pod *v1.Pod) string {
-	queueName := constants.ApplicationDefaultQueue
-	if an := GetPodLabelValue(pod, constants.LabelQueueName); an != "" {
-		queueName = an
-	} else if qu := GetPodAnnotationValue(pod, constants.AnnotationQueueName); qu != "" {
-		queueName = qu
+	if queueName := GetPodLabelValue(pod, constants.LabelQueueName); queueName != "" {
+		return queueName
+	} else if queueName := GetPodAnnotationValue(pod, constants.AnnotationQueueName); queueName != "" {
```
wilfred-s, line 107: NIT: No need to have an `else` here as we return if we have a queue above

wilfred-s (changes requested):

craigcondit:
Obsolete after YUNIKORN-2711 and YUNIKORN-2703.

PR #749, opened 2023-12-15
[YUNIKORN-2271] Incorrect handling of GPU only resources

zhuqi-lucas:
### What is this PR for?
https://github.com/apache/yunikorn-k8shim/blob/a118ba6c4d84804e2a407f9d91196ece4690cf09/pkg/common/resource.go#L61-L63
 
this code seems to have a bug. When I define resource like this:

request:
  nvidia.com/gpu: 1
limit:
  nvidia.com/gpu: 1
 
this is considered as QoS best effort and returned with just

Resources:
  pod:1
but I think this is a valid configuration that a pod only specifies GPU resource without memory or CPU. It seems this is the K8s upstream code: qos.GetPodQOS() causes this..

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:

### DIFF
--- a/pkg/common/resource.go
+++ b/pkg/common/resource.go
@@ -57,11 +57,6 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
 		Resources: map[string]*si.Quantity{"pods": {Value: 1}},
 	}
 
-	// A QosBestEffort pod does not request any resources, just a single pod
-	if qos.GetPodQOS(pod) == v1.PodQOSBestEffort {
-		return podResource
-	}
-
 	for _, c := range pod.Spec.Containers {
 		resourceList := c.Resources.Requests
 		containerResource := getResource(resourceList)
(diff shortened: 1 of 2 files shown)


craigcondit:
I think we should just remove the QoS handling entirely; the flag is not used anywhere in YuniKorn currently except here to mutate resource requests. Since a QoS pod can't refer to memory or CPU, K8s itself should handle this case and we can just use the requests that are already present regardless of the QoS setting.

yangwwei:
Agree with @craigcondit . IIRC, the reason we add this https://github.com/apache/yunikorn-k8shim/pull/52 is to handle the case when a pod has no resource types specified at all. But since now we added resource type "pods:1" to the resource all the time, we no longer have that issue.

zhuqi-lucas:
Thank you @craigcondit @yangwwei for review, agree we can remove the QoS handling entirely, adressed in latest PR now.

pbacsko (approved):
+1 LGTM

yangwwei (approved):

craigcondit (approved):
+1 LGTM

PR #751, opened 2023-12-21
[YUNIKORN-2283] Support different compression method zlib for config map

FrankYang0529:
### What is this PR for?

Currently, we only support gzip compression method for config map. We can support other compression methods in [golang compress package](https://pkg.go.dev/compress).

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-2283

Covered by unit tests.


craigcondit:
This is a duplicate issue and we have decided not to implement support for other algorithms.

PR #752, opened 2023-12-22
[YUNIKORN-2287] Decompress function doesn't need to decode base64

FrankYang0529:
### What is this PR for?

Fix `Decompress` function and check the following case can run:

1. Use gzip to compress and use base64 to encode the config

```sh
echo "
partitions:
  - name: default
    placementrules:
      - name: tag
        value: namespace
        create: true
    queues:
      - name: root
        submitacl: '*'
        queues:
          - name: parent
            submitacl: '*'" | gzip | base64

```

2. Set the result to `queues.yaml.gz` in `binaryData` field.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2287

Covered by e2e test.

### DIFF
--- a/pkg/conf/schedulerconf.go
+++ b/pkg/conf/schedulerconf.go
@@ -21,7 +21,6 @@ package conf
 import (
 	"bytes"
 	"compress/gzip"
-	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"flag"
@@ -452,18 +451,10 @@ func DumpConfiguration() {
 
 func Decompress(key string, value []byte) (string, string) {
 	var uncompressedData string
-	decodedValue := make([]byte, base64.StdEncoding.DecodedLen(len(value)))
-	n, err := base64.StdEncoding.Decode(decodedValue, value)
-	if err != nil {
-		log.Log(log.ShimConfig).Error("failed to decode schedulerConfig entry", zap.Error(err))
-		return "", ""
-	}
-	decodedValue = decodedValue[:n]
 	splitKey := strings.Split(key, ".")
 	compressionAlgo := splitKey[len(splitKey)-1]
 	if strings.EqualFold(compressionAlgo, constants.GzipSuffix) {
-		reader := bytes.NewReader(decodedValue)
-		gzReader, err := gzip.NewReader(reader)
+		gzReader, err := gzip.NewReader(bytes.NewReader(value))
 		if err != nil {
 			log.Log(log.ShimConfig).Error("failed to decompress decoded schedulerConfig entry", zap.Error(err))
 			return "", ""
--- a/test/e2e/configmap/configmap_suite_test.go
+++ b/test/e2e/configmap/configmap_suite_test.go
@@ -1,0 +1,88 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package configmap
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/ginkgo/v2/reporters"
+	"github.com/onsi/gomega"
+
+	"github.com/apache/yunikorn-k8shim/test/e2e/framework/configmanager"
+	"github.com/apache/yunikorn-k8shim/test/e2e/framework/helpers/common"
+	"github.com/apache/yunikorn-k8shim/test/e2e/framework/helpers/k8s"
+	"github.com/apache/yunikorn-k8shim/test/e2e/framework/helpers/yunikorn"
+)
+
+func init() {
+	configmanager.YuniKornTestConfig.ParseFlags()
+}
+
+func TestConfigMap(t *testing.T) {
+	ginkgo.ReportAfterSuite("TestConfigMap", func(report ginkgo.Report) {
+		err := common.CreateJUnitReportDir()
+		Ω(err).NotTo(gomega.HaveOccurred())
+		err = reporters.GenerateJUnitReportWithConfig(
+			report,
+			filepath.Join(configmanager.YuniKornTestConfig.LogDir, "TEST-configmap_junit.xml"),
+			reporters.JunitReportConfig{OmitSpecLabels: true},
+		)
+		Ω(err).NotTo(HaveOccurred())
+	})
+	gomega.RegisterFailHandler(ginkgo.Fail)
+	ginkgo.RunSpecs(t, "TestConfigMap", ginkgo.Label("TestConfigMap"))
+}
+
+var (
+	kClient    k8s.KubeCtl
+	restClient yunikorn.RClient
+)
+
+var _ = BeforeSuite(func() {
+
+	kClient = k8s.KubeCtl{}
+	Ω(kClient.SetClient()).To(BeNil())
+
+	restClient = yunikorn.RClient{}
+	Ω(restClient).NotTo(BeNil())
+
+	yunikorn.EnsureYuniKornConfigsPresent()
+
+	By("Port-forward the scheduler pod")
+	err := kClient.PortForwardYkSchedulerPod()
+	Ω(err).NotTo(HaveOccurred())
+})
+
+var _ = AfterSuite(func() {})
+
+var Describe = ginkgo.Describe
+var It = ginkgo.It
+var By = ginkgo.By
+var BeforeEach = ginkgo.BeforeEach
+var AfterEach = ginkgo.AfterEach
+var BeforeSuite = ginkgo.BeforeSuite
+var AfterSuite = ginkgo.AfterSuite
+
+// Declarations for Gomega Matchers
+var Equal = gomega.Equal
+var Ω = gomega.Expect
+var BeNil = gomega.BeNil
+var HaveOccurred = gomega.HaveOccurred
--- a/test/e2e/configmap/configmap_test.go
+++ b/test/e2e/configmap/configmap_test.go
@@ -1,0 +1,112 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package configmap
+
+import (
+	"bytes"
+	"compress/gzip"
+	"io"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/apache/yunikorn-core/pkg/common/configs"
+	"github.com/apache/yunikorn-k8shim/test/e2e/framework/configmanager"
+	"github.com/apache/yunikorn-k8shim/test/e2e/framework/helpers/k8s"
+	"github.com/apache/yunikorn-k8shim/test/e2e/framework/helpers/yunikorn"
+)
+
+var oldConfigMap = new(v1.ConfigMap)
+
+var _ = Describe("ConfigMap", func() {
+	BeforeEach(func() {
+		By("Get previous config")
+		var err error
+		oldConfigMap, err = kClient.GetConfigMaps(configmanager.YuniKornTestConfig.YkNamespace,
+			configmanager.DefaultYuniKornConfigMap)
+		Ω(err).NotTo(HaveOccurred())
+		Ω(oldConfigMap).NotTo(BeNil())
+	})
+
+	It("Verify_ConfigMap_File", func() {
+		configMap, err := k8s.GetConfigMapObj("../testdata/yunikorn-configs.yaml")
+		Ω(err).NotTo(HaveOccurred())
+		Ω(configMap).NotTo(BeNil())
+
+		By("Updating the config map with BinaryData")
+		configMap.Namespace = configmanager.YuniKornTestConfig.YkNamespace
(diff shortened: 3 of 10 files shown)


chia7712:
@FrankYang0529 I have another idea before reviewing. If this function is never in production, can we just get rid of it to simplify our code base? 

@wilfred-s @craigcondit WDYT?

test/e2e/testdata/compressed_yunikorn-configs.yaml
```diff
@@ -0,0 +1,33 @@
+#Licensed to the Apache Software Foundation (ASF) under one
```
wilfred-s, line 1: space between # and the wording on all lines

test/e2e/testdata/compressed_yunikorn-configs.yaml
```diff
@@ -0,0 +1,33 @@
...
+#distributed under the License is distributed on an "AS IS" BASIS,
+#WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#See the License for the specific language governing permissions and
+#limitations under the License.
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: yunikorn-configs
+binaryData:
+# echo "
```
wilfred-s, line 21: Explain what this commented info is about

wilfred-s (changes requested):
minor tweaks, see comment in the jira would be nice to document why the decode is removed

chia7712:
> why the decode is removed

@wilfred-s did you feel that decode is a kind of behavior change of k8s? This yunikorn feature is already existent for a period of time. We have to keep compatibility for previous k8s if previous k8s does not do base64 decoding.

If the base64 decoding happens way down in the k8s always, it seems we have not got binarydata work in yunikorn. That is to say, that yunikorn feature is nonexistent actually :)

In short, removing this feature seems to be acceptable. The benefit includes 

1) don't worry for k8s compatibility (if it exists)
2) simplify code base

test/e2e/testdata/compressed_yunikorn-configs.yaml
FrankYang0529: Other test data also need to add space to it. Do we want to fix this together in this PR?

https://github.com/apache/yunikorn-k8shim/blob/master/test/e2e/testdata/sleeppod_template.yaml

FrankYang0529:
> see comment in the jira would be nice to document why the decode is removed

I don't have a good reference to show why the configmap in Golang object is already decoded, but I may be able to proof that if object from our utility function [GetConfigMapObj](https://github.com/apache/yunikorn-k8shim/blob/3ab3b2d03dd21896e803dc871df706d2563e7064/test/e2e/framework/helpers/k8s/k8s_utils.go#L626) is decoded, then object from object from [CoreV1().ConfigMaps(namespace).Get](https://github.com/apache/yunikorn-k8shim/blob/3ab3b2d03dd21896e803dc871df706d2563e7064/pkg/client/kubeclient.go#L156) is decoded too. The reason is that both functions use same serializers.

### GetConfigMapObj

1. It uses [scheme.Codecs.UniversalDeserializer().Decode](https://github.com/apache/yunikorn-k8shim/blob/3ab3b2d03dd21896e803dc871df706d2563e7064/test/e2e/framework/helpers/common/deserializer.go#L28).
2. `Codecs` is from [serializer.NewCodecFactory](https://github.com/kubernetes/client-go/blob/bec45d3c69b597c35e0246cce5c5001127edf994/kubernetes/scheme/register.go#L81).
3. In `NewCodecFactory`, it gets `serializers` from [newSerializersForScheme](https://github.com/kubernetes/apimachinery/blob/194b6f8ddf62f55d1b8d997223e94ace2c1def31/pkg/runtime/serializer/codec_factory.go#L182).
4. In [newCodecFactory](https://github.com/kubernetes/apimachinery/blob/194b6f8ddf62f55d1b8d997223e94ace2c1def31/pkg/runtime/serializer/codec_factory.go#L183), it collects all `serializers` as [decoders](https://github.com/kubernetes/apimachinery/blob/194b6f8ddf62f55d1b8d997223e94ace2c1def31/pkg/runtime/serializer/codec_factory.go#L193-L194) and sets it as `runtime.Decoder` to [universal](https://github.com/kubernetes/apimachinery/blob/194b6f8ddf62f55d1b8d997223e94ace2c1def31/pkg/runtime/serializer/codec_factory.go#L235).
5. The `universal` is a `runtime.Decoder` from [recognizer.NewDecoder](https://github.com/kubernetes/apimachinery/blob/194b6f8ddf62f55d1b8d997223e94ace2c1def31/pkg/runtime/serializer/recognizer/recognizer.go#L45-L49).
6. In [recognizer.Decoder.Decode](https://github.com/kubernetes/apimachinery/blob/194b6f8ddf62f55d1b8d997223e94ace2c1def31/pkg/runtime/serializer/recognizer/recognizer.go#L80-L128), it tries all `serializers`.

### CoreV1().ConfigMaps(namespace).Get

1. In `CoreV1().ConfigMaps(namespace).Get`, it calls [Do](https://github.com/kubernetes/client-go/blob/bec45d3c69b597c35e0246cce5c5001127edf994/kubernetes/typed/core/v1/configmap.go#L78).
2. In `Do`, it calls [Request.transformResponse](https://github.com/kubernetes/client-go/blob/bec45d3c69b597c35e0246cce5c5001127edf994/rest/request.go#L1040).
3. In `transformResponse`, it calls [r.c.content.Negotiator.Decoder](https://github.com/kubernetes/client-go/blob/bec45d3c69b597c35e0246cce5c5001127edf994/rest/request.go#L1114).
4. In `Decoder`, [it checks media type and content type, and gets `serializer` from `SerializerInfo`.](https://github.com/kubernetes/apimachinery/blob/194b6f8ddf62f55d1b8d997223e94ace2c1def31/pkg/runtime/negotiate.go#L58-L68)
5. The `SerializerInfo` is from [SerializerInfoForMediaType](https://github.com/kubernetes/apimachinery/blob/194b6f8ddf62f55d1b8d997223e94ace2c1def31/pkg/runtime/codec.go#L265-L277). The `SerializerInfoForMediaType` is just a utility function to check which `mediaTypes` match to `contentType`, so we have to check where `mediaTypes` from.
7. The `mediaTypes` is from [n.serializer.SupportedMediaTypes](https://github.com/kubernetes/apimachinery/blob/194b6f8ddf62f55d1b8d997223e94ace2c1def31/pkg/runtime/negotiate.go#L59).
8. In [DecoderToVersion](https://github.com/kubernetes/apimachinery/blob/194b6f8ddf62f55d1b8d997223e94ace2c1def31/pkg/runtime/serializer/codec_factory.go#L334-L338), the `serializer` is equal to `runtime.Decoder`.

If we can proof the `runtime.Decoder` in both functions are same, then they will have same behavior - getting a decoded configmap. We have to check which structure implements `SupportedMediaTypes`.

1. In k8shim, we get `kubernetes.Clientset` from [kubernetes.NewForConfig](https://github.com/apache/yunikorn-k8shim/blob/3ab3b2d03dd21896e803dc871df706d2563e7064/pkg/client/kubeclient.go#L45).
2. In `NewForConfig`, it calls [NewForConfigAndClient](https://github.com/kubernetes/client-go/blob/bec45d3c69b597c35e0246cce5c5001127edf994/kubernetes/clientset.go#L473).
3. In `NewForConfigAndClient`, it calls [corev1.NewForConfigAndClient](https://github.com/kubernetes/client-go/blob/bec45d3c69b597c35e0246cce5c5001127edf994/kubernetes/clientset.go#L583).
4. In `corev1.NewForConfigAndClient`.
4-1a. In `corev1.NewForConfigAndClient`, it calls [setConfigDefaults](https://github.com/kubernetes/client-go/blob/bec45d3c69b597c35e0246cce5c5001127edf994/kubernetes/typed/core/v1/core_client.go#L137).
4-1b. In `setConfigDefaults`, it sets `NegotiatedSerializer` as [scheme.Codecs.WithoutConversion](https://github.com/kubernetes/client-go/blob/bec45d3c69b597c35e0246cce5c5001127edf994/kubernetes/typed/core/v1/core_client.go#L166). The `scheme.Codecs` is same as first step in `k8s.GetConfigMapObj`.
4-1c. In `WithoutConversion`, it return [WithoutConversionCodecFactory{f}](https://github.com/kubernetes/apimachinery/blob/194b6f8ddf62f55d1b8d997223e94ace2c1def31/pkg/runtime/serializer/codec_factory.go#L245-L247). This structure has `DecoderToVersion` function in step 7 (last list).
4-2a. In `corev1.NewForConfigAndClient`, it calls [rest.RESTClientForConfigAndClient](https://github.com/kubernetes/client-go/blob/bec45d3c69b597c35e0246cce5c5001127edf994/kubernetes/typed/core/v1/core_client.go#L140).
4-2b. In `rest.RESTClientForConfigAndClient`, it sets `Negotiator` as [runtime.NewClientNegotiator](https://github.com/kubernetes/client-go/blob/bec45d3c69b597c35e0246cce5c5001127edf994/rest/config.go#L374). This function uses `config.NegotiatedSerializer` which is equal to step 4-1b.

Finally, `NegotiatedSerializer` is [WithoutConversionCodecFactory](https://github.com/kubernetes/apimachinery/blob/194b6f8ddf62f55d1b8d997223e94ace2c1def31/pkg/runtime/serializer/codec_factory.go#L319-L338) and it also is [CodecFactory](https://github.com/kubernetes/apimachinery/blob/194b6f8ddf62f55d1b8d997223e94ace2c1def31/pkg/runtime/serializer/codec_factory.go#L176-L184). The [SupportedMediaTypes](https://github.com/kubernetes/apimachinery/blob/194b6f8ddf62f55d1b8d997223e94ace2c1def31/pkg/runtime/serializer/codec_factory.go#L250-L252) in `CodecFactory` returns `f.accepts` which gets same serializers as `f.universal`.

craigcondit:
> see comment in the jira would be nice to document why the decode is removed

I think there's more than enough comments in the JIRA relating to why this is not needed. Removing an unnecessary line of code from the PR doesn't warrant a comment either. What would it mean to a reader a year from now? (probably nothing at all).

chia7712:
@FrankYang0529 could you please fix the conflicts?

test/e2e/bin_packing/bin_packing_test.go
```diff
@@ -198,8 +198,9 @@ var _ = Describe("", func() {
 			tests.LogTestClusterInfoWrapper(testDescription.FailureMessage(), []string{ns})
 			tests.LogYunikornContainer(testDescription.FailureMessage())
 		}
+
 		By("Tear down namespace: " + ns)
-		err := kClient.DeleteNamespace(ns)
+		err := kClient.TearDownNamespace(ns)
```
chia7712, line 203: Is this another bug fix? If so, could you file another JIRA to handle it?

test/e2e/configmap/configmap_suite_test.go
```diff
@@ -0,0 +1,94 @@
...
+	kClient = k8s.KubeCtl{}
+	Ω(kClient.SetClient()).To(BeNil())
+
+	restClient = yunikorn.RClient{}
+	Ω(restClient).NotTo(BeNil())
+
+	yunikorn.EnsureYuniKornConfigsPresent()
+
+	// Restart yunikorn and port-forward
+	// Required to change node sort policy.
```
chia7712, line 70: It seems those code are copied from other e2e, right? if so, please make sure this e2e requires them also.

chia7712 (commented):
@FrankYang0529 thanks for this fix. please take a look at following comments.

test/e2e/configmap/configmap_suite_test.go
FrankYang0529: Yeah, we don't need it. Thanks for catching this.

test/e2e/bin_packing/bin_packing_test.go
FrankYang0529: In `Verify_BinPacking_Node_Order_Memory`, we create jobs without waiting for pod terminating. It's better to use `TearDownNamespace` to cleanup workload before removing the namespace. Do we want to create another JIRA or just fix it in this PR?

https://github.com/apache/yunikorn-k8shim/blob/3e59ddcdd917c292635c6be0ddb8170356ae6511/test/e2e/bin_packing/bin_packing_test.go#L171-L192

chia7712: > Do we want to create another JIRA or just fix it in this PR?

yep,that is not related to this PR/issue. It would be better to have a separate ticket for it.

chia7712:
@FrankYang0529 please rebase PR as #761 is already merged.

FrankYang0529:
The e2e failed case is `Verify_basic_preemption` which is not related to this PR and is filed in another JIRA [YUNIKORN-2313](https://issues.apache.org/jira/browse/YUNIKORN-2313).

test/e2e/configmap/configmap_test.go
```diff
@@ -0,0 +1,144 @@
...
+	BeforeEach(func() {
+		By("Get previous config")
+		var err error
+		oldConfigMap, err = kClient.GetConfigMaps(configmanager.YuniKornTestConfig.YkNamespace,
+			configmanager.DefaultYuniKornConfigMap)
+		Ω(err).NotTo(HaveOccurred())
+		Ω(oldConfigMap).NotTo(BeNil())
+	})
+
+	It("Verify_Compressed_ConfigMap", func() {
```
chia7712, line 48: It seems to me this e2e should focus on the way that YuniKorn loads the config file. Hence, loading the file is more suitable for this e2e rather than creating a compressed file directly. WDYT?

chia7712 (commented):
@FrankYang0529 thanks for updating patch. a couple of comments below.

chia7712:
@FrankYang0529 please fix conflicts and rebase code

test/e2e/configmap/configmap_test.go
FrankYang0529: Do you mean that we just use `Verify_Compressed_ConfigMap_From_File` and remove `Verify_Compressed_ConfigMap`?

test/e2e/configmap/configmap_suite_test.go
```diff
@@ -0,0 +1,89 @@
...
+	kClient = k8s.KubeCtl{}
+	Ω(kClient.SetClient()).To(BeNil())
+
+	restClient = yunikorn.RClient{}
+	Ω(restClient).NotTo(BeNil())
+
+	yunikorn.EnsureYuniKornConfigsPresent()
+
+	ginkgo.By("Port-forward scheduler pod after restart")
+	yunikorn.RestorePortForwarding(&kClient)
```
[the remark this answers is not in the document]
FrankYang0529: We remove restarting pod, so we can just do port forward directly. Updated it. Thanks.

pkg/conf/schedulerconf.go
```diff
@@ -452,17 +451,10 @@ func DumpConfiguration() {
...
-		log.Log(log.ShimConfig).Error("failed to decode schedulerConfig entry", zap.Error(err))
-		return "", ""
-	}
-	decodedValue = decodedValue[:n]
 	splitKey := strings.Split(key, ".")
 	compressionAlgo := splitKey[len(splitKey)-1]
 	if strings.EqualFold(compressionAlgo, constants.GzipSuffix) {
-		reader := bytes.NewReader(decodedValue)
+		reader := bytes.NewReader(value)
 		gzReader, err := gzip.NewReader(reader)
```
chia7712, line 458: How about `gzReader, err := gzip.NewReader(bytes.NewReader(value))`?

FrankYang0529: Yes, updated it. Thank you.

chia7712 (approved):
LGTM

PR #754, opened 2023-12-29
[YUNIKORN-2292] Flaky E2E Test: Orphan pods still exist after TearDownNamespace()

chenyulin0719:
### What is this PR for?
- Check and delete Jobs in TearDownNamespace()

2024/01/02 Update:
- Also clenaup Deployment,DaemonSet,ReplicaSet,StatefulSet,PVC in TearDownNamespace()

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
NA

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2292

All existing test should pass.
> make test
make e2e_test

NA

### Questions:
NA

### DIFF
--- a/test/e2e/framework/helpers/k8s/k8s_utils.go
+++ b/test/e2e/framework/helpers/k8s/k8s_utils.go
@@ -165,6 +165,18 @@ func (k *KubeCtl) GetPodsByOptions(options metav1.ListOptions) (*v1.PodList, err
 	return k.clientSet.CoreV1().Pods("").List(context.TODO(), options)
 }
 
+func (k *KubeCtl) GetDeployments(namespace string) (*appsv1.DeploymentList, error) {
+	return k.clientSet.AppsV1().Deployments(namespace).List(context.TODO(), metav1.ListOptions{})
+}
+
+func (k *KubeCtl) GetDaemonSets(namespace string) (*appsv1.DaemonSetList, error) {
+	return k.clientSet.AppsV1().DaemonSets(namespace).List(context.TODO(), metav1.ListOptions{})
+}
+
+func (k *KubeCtl) GetStatefulSets(namespace string) (*appsv1.StatefulSetList, error) {
+	return k.clientSet.AppsV1().StatefulSets(namespace).List(context.TODO(), metav1.ListOptions{})
+}
+
 func (k *KubeCtl) GetJobs(namespace string) (*batchv1.JobList, error) {
 	return k.clientSet.BatchV1().Jobs(namespace).List(context.TODO(), metav1.ListOptions{})
 }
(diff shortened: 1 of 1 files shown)


chenyulin0719:
Have checked the error message is equal to [YUNIKORN-2291 Flaky e2e test](https://issues.apache.org/jira/browse/YUNIKORN-2291)

One potential issue involves checking the application's failing state. I will create another Jira ticket once this issue is confirmed.
https://github.com/apache/yunikorn-k8shim/blob/master/test/e2e/gang_scheduling/gang_scheduling_test.go#L288

chia7712:
@chenyulin0719 Is the failed e2e related to this PR?

test/e2e/framework/helpers/k8s/k8s_utils.go
```diff
@@ -391,7 +391,13 @@ func (k *KubeCtl) DeleteNamespace(namespace string) error {
 }
 
 func (k *KubeCtl) TearDownNamespace(namespace string) error {
```
chia7712, line 393: Should it cleanup both PV and PVC ?

chenyulin0719: Hi @chia7712,
I don't think we should cleanup both PV and PVC in TearDownNamespace.

PV/PVC were only used in "persistent_volume" e2e test.  I think the cleanup responsibility is on the e2e test itself. (Besides, PV is cluster resource and is not bounded to namesapce.)

One of the reason we cleanup Jobs/Pods in TearDownNamespace() is those objects are widely used in serveral e2e tests and sadly we didn't handle all the cleanup.

(IMO, the best practice is each e2e test clean Jobs/Pods by themself. However, it quite difficult to review all the e2e test.)

chenyulin0719:
> @chenyulin0719 Is the failed e2e related to this PR?

Hi @chia7712 ,
The failed e2e test is not related to this PR. I've created below Jira to cover it:

- [[YUNIKORN-2293] Flaky E2E Test: Failed asserts in LogTestClusterInfoWrapper() blocked the resources cleanup steps](https://issues.apache.org/jira/browse/YUNIKORN-2293)
- [[YUNIKORN-2294] Flaky E2E Test: "Verify_Hard_GS_Failed_State" polling short-lived "Failing" application status](https://issues.apache.org/jira/browse/YUNIKORN-2294)

test/e2e/framework/helpers/k8s/k8s_utils.go
chia7712: My point was that all e2e should reuse the code to cleanup resources. PV, PVC, pods, and jobs should be cleanup when the e2e is closing even if the e2e does not use such resource. That could make isolation between e2e test and all tests get benefit from it.

brandboat: Agree with Chia-Ping's point. I also believe that we should invoke `DeleteWorkloadAndPods` in the `tearDownNamespace` method.

chia7712: @chenyulin0719 any feedback?

chenyulin0719: @chia7712 
I'm working on extending TearDownNamespace() to cover below namespace level object:
1. Deployment
2. DaemonSet
3. GetReplicaSet
4. GetStatefulSet
5. Pods
6. PVC

>PV, PVC, pods, and jobs should be cleanup when the e2e is closing even if the e2e does not use such resource. 

My previous thought is resource cleanup should be 'explicit'.
ex: "If I create pvc in e2e test, I should delete it after this e2e test.  If I don't use pvc in e2e test,  I don't need to check pvc existing.

However, currently, we're not using the 'explicit' cleanup in every e2e test, and looks like TearDownNamespace() is the shared function to cleanup namespace for multiple e2e test. So I'll follow the suggestion and make it more generic.

@brandboat Thanks for the suggestion. I've checked the DeleteWorkloadAndPods() only deletes one workload object per trigger. (If we have 10 Jobs in namespace, we would need to call DeleteWorkloadAndPods() 10 times). Therefore, I will create a new function for each resource type.

chia7712: @chenyulin0719 thanks for update. sounds good!

chenyulin0719: New commit pushed, 
Also clenaup Deployment,DaemonSet,ReplicaSet,Stateful,PVC in TearDownNamespace()

test/e2e/framework/helpers/k8s/k8s_utils.go
```diff
@@ -391,7 +451,38 @@ func (k *KubeCtl) DeleteNamespace(namespace string) error {
...
-	err := k.DeletePods(namespace)
+	var err error
+	err = k.DeleteDeployments(namespace)
```
chia7712, line 455: @chenyulin0719 Could you add comments to explain the order of delete?

craigcondit (approved):

test/e2e/framework/helpers/k8s/k8s_utils.go
craigcondit: I think this is fairly self-explanatory. Objects that create other objects are deleted first.

PR #759, opened 2024-01-05
[YUNIKORN-2294] Flaky E2E Test: "Verify_Hard_GS_Failed_State" polling short-lived "Failing" application status

chenyulin0719:
### What is this PR for?
- Change the App state polling frequency from 300ms to 100 ms

2024/01/06 Update
- Change to check 'Failed' appInfo instead of 'Failing' appInfo.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
NA

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2294

All the existing test should pass

NA

### Questions:
NA

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ module github.com/apache/yunikorn-k8shim
 go 1.20
 
 require (
-	github.com/apache/yunikorn-core v0.0.0-20240103094035-ba62c5db9f61
+	github.com/apache/yunikorn-core v0.0.0-20240105094327-77e19f6aca27
 	github.com/apache/yunikorn-scheduler-interface v0.0.0-20240102192148-d4b43d6910c9
 	github.com/google/go-cmp v0.6.0
 	github.com/google/uuid v1.3.1
--- a/test/e2e/framework/helpers/yunikorn/rest_api_utils.go
+++ b/test/e2e/framework/helpers/yunikorn/rest_api_utils.go
@@ -180,12 +180,22 @@ func (c *RClient) GetCompletedAppInfo(partition string, appID string) (*dao.Appl
 		return nil, err
 	}
 
+	// ApplicationID is not unique in the completed applications list. Try to get the latest one.
+	var latestApp *dao.ApplicationDAOInfo
+	var latestSubmissionTime = int64(0)
+
 	for _, app := range apps {
-		if app.ApplicationID == appID {
-			return app, nil
+		if app.ApplicationID == appID && app.SubmissionTime > latestSubmissionTime {
+			latestApp = app
+			latestSubmissionTime = app.SubmissionTime
 		}
 	}
-	return nil, fmt.Errorf("App is not in 'Failed', 'Expired', 'Completed' state.")
+
+	if latestApp != nil {
+		return latestApp, nil
+	}
+
+	return nil, fmt.Errorf("No application found with ID %s in 'Failed', 'Expired', 'Completed' state", appID)
 }
 
 func (c *RClient) GetAllocationLog(partition string, queueName string, appID string, podName string) ([]*dao.AllocationAskLogDAOInfo, error) {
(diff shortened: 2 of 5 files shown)


craigcondit (changes requested):
This will very likely do nothing to fix this issue. Testing for transient states is not reliable, full stop. Instead, it would be better to use the REST API to grab the state transition log and look for the failing state. This is persistent for as long as the application object is visible. 

Even better would be to simply ignore the failing status and check for externally visible behavior instead of internal implementation. If the application fails, the pods will be killed off (at least if I remember correctly). Set up an infinite timed pause pod with immediate termination timeout, and wait for the pod to terminate (for say 30s). If the pod terminates, the test passes. If a timeout occurs, the test fails. 

Alternatively, if pods aren't automatically killed, we can wait for the application to disappear from the REST API (as that would have indicated failure as well). In either case, checking explicitly for FAILING state is unnecessary.

pbacsko (changes requested):
I agree, waiting for a state which only exists for a very short period of time is just asking for trouble in an e2e test. This cannot be tested reliably on such a high level. Why can't we just wait for the terminal "Failed" state?

chenyulin0719:
Hi @craigcondit, @pbacsko ,
I just realized that after [YUNIKORN-2235] (https://issues.apache.org/jira/browse/YUNIKORN-2235), we can fetch 'Failed' appInfo through the API:
- /ws/v1/partition/default/applications/completed

However, the result is a list and the appID is not unique, so I have to iterate it.
New commits are availiable now.

go.mod
```diff
@@ -21,7 +21,7 @@ module github.com/apache/yunikorn-k8shim
 go 1.20
 
 require (
-	github.com/apache/yunikorn-core v0.0.0-20240103094035-ba62c5db9f61
+	github.com/apache/yunikorn-core v0.0.0-20240105094327-77e19f6aca27
```
chenyulin0719, line 24: Have to update core version to use the latest api:
- ws/v1/partition/default/applications/completed

test/e2e/framework/helpers/yunikorn/rest_api_utils.go
```diff
@@ -235,6 +279,10 @@ func (c *RClient) WaitForAppStateTransition(partition string, queue string, appI
 	return wait.PollUntilContextTimeout(context.TODO(), time.Millisecond*300, time.Duration(timeout)*time.Second, false, c.isAppInDesiredState(partition, queue, appID, state).WithContext())
 }
 
+func (c *RClient) WaitForCompletedAppStateTransition(partition string, appID string, state string, timeout int) error {
+	return wait.PollUntilContextTimeout(context.TODO(), time.Millisecond*300, time.Duration(timeout)*time.Second, false, c.isAppInDesiredCompletedState(partition, appID, state).WithContext())
```
pbacsko, line 283: 1. Just use int64 for timeout so you don't need to cast
2. 300 millisec is too low in an e2e test, increase it to 1 second

Also fix existing methods with these changes. `WaitforQueueToAppear()` actually polls in every 300 usec (!), which is obviously a typo. Don't put such a load on Yunikorn. The other method is `WaitForAppStateTransition()`.

pbacsko (changes requested):
Please address the remaining issues.

test/e2e/framework/helpers/yunikorn/rest_api_utils.go
chenyulin0719: Hi @pbacsko,
Just set below functions to 1 sec (5c165af2539460e97e2ecc03979495644188814f)
1.  WaitforQueueToAppear
2. WaitForAppStateTransition
3. WaitForCompletedAppStateTransition

There have some polling functions with 100 interval under k8s_utils.go,  but I'm keeping it as is because they are widely used, and I don't want to increase the time for e2e testing.

For the int64 timeout,  I just realized that even we use int64 for timeout, we still need to cast it to time.Duration because the type of time.Second is 'time.Duration'(alias of int64).  We still need to cast timeout to 'time.Duration'. So I just keep it as it was.

pbacsko:
Current changes look good, I'll merge when the e2e tests finish.

pbacsko (approved):
+1

PR #770, opened 2024-01-26
[YUNIKORN-2366] Update GetPodResources() to handle in-place pod resource updates

craigcondit:
### What is this PR for?
Implement initial support for the upcoming Kubernetes InPlacePodVerticalScaling feature when doing Pod resource calculations. If present, the allocated resources of each container will be used in preference to the requested resources. Additionally, if the Pod status flag Resize is set to Proposed, use the larger of both requested and allocated for each resource type.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2366

Added unit tests to cover logic introduced by the upstream K8s feature.

### Questions:

### DIFF
--- a/pkg/common/resource.go
+++ b/pkg/common/resource.go
@@ -56,10 +56,9 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
 		Resources: map[string]*si.Quantity{"pods": {Value: 1}},
 	}
 
-	for _, c := range pod.Spec.Containers {
-		resourceList := c.Resources.Requests
-		containerResource := getResource(resourceList)
-		podResource = Add(podResource, containerResource)
+	count := len(pod.Spec.Containers)
+	for i := 0; i < count; i++ {
+		podResource = Add(podResource, containerResource(pod, i))
 	}
 
 	// each resource compare between initcontainer and sum of containers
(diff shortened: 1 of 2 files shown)


pkg/common/resource.go
```diff
@@ -87,6 +86,48 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
...
+	// alpha: v1.27
+	// beta: v1.31?
+	// If AllocatedResources are present, these need to be used in preference to pod resource requests.
+	// Additionally, if the Resize pod status is InProgress, then the maximum of the request and allocated values need
+	// to be used.
+	requested := pod.Spec.Containers[i].Resources.Requests
+	if len(pod.Status.ContainerStatuses) == 0 {
+		return getResource(requested)
+	}
+	allocated := pod.Status.ContainerStatuses[i].AllocatedResources
```
chia7712, line 100: Do `ContainerStatuses` and `Containers`  have same order?

craigcondit: Yes, this is part of the spec.

craigcondit: See https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/1287-in-place-update-pod-resources/README.md#scheduler-and-api-server-interaction for more details.

pkg/common/resource.go
```diff
@@ -87,6 +86,48 @@ func GetPodResource(pod *v1.Pod) (resource *si.Resource) {
...
+	if len(allocated) == 0 {
+		// no allocatedResources present, use requested
+		return getResource(requested)
+	}
+	if pod.Status.Resize == v1.PodResizeStatusProposed {
```
chia7712, line 105: from the spec: `For containers which have Status.Resize = "Proposed", it must be pessimistic and assume that the resize will be imminently accepted. Therefore it must use the larger of the Pod's Spec...Resources.Requests and Status...AllocatedResources values`

Should we replace `PodResizeStatusInProgress` by `PodResizeStatusProposed`?

chia7712: oh, you have updated the code. please ignore my comment.

craigcondit: Yes, I spotted that while re-reading the spec.

chia7712 (approved):
LGTM

pbacsko (approved):
+1

PR #773, opened 2024-01-31
[YUNIKORN-2369] AM: Add token suffix when using unique applicationId feature

craigcondit:
### What is this PR for?
Workaround missing pod UID in admission controller lifecycle by setting a suffix of "-uniqueautogen" when the unique applicationId feature is in use. This suffix will be detected and replaced by the scheduler since the pod UID will exist at that point.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2369

Added unit test cases for both generation and replacement.

### Questions:

### DIFF
--- a/pkg/common/utils/utils.go
+++ b/pkg/common/utils/utils.go
@@ -41,6 +41,7 @@ import (
 )
 
 const userInfoKey = siCommon.DomainYuniKorn + "user.info"
+const uniqueAutogenSuffix = "-uniqueautogen"
 
 var pluginMode bool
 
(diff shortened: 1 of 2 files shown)


pbacsko (approved):
+1 LGTM (pending tests)

PR #786, opened 2024-02-13
[YUNIKORN-2406] Ensure task status is updated for existing allocations

craigcondit:
### What is this PR for?
When core accepts an existing allocation, force task state to Bound to ensure that we track state properly.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2406

E2e testing only; this code is deeply tied to the core implementation.

### Questions:


pkg/cache/scheduler_callback.go
```diff
@@ -60,8 +60,15 @@ func (callback *AsyncRMCallback) UpdateAllocation(response *si.AllocationRespons
 			return err
 		}
 		if app := callback.context.GetApplication(alloc.ApplicationID); app != nil {
-			ev := NewAllocateTaskEvent(app.GetApplicationID(), alloc.AllocationKey, alloc.AllocationID, alloc.NodeID)
-			dispatcher.Dispatch(ev)
+			if task := callback.context.getTask(app.GetApplicationID(), alloc.AllocationKey); task != nil {
```
chia7712, line 63: Can we get `task` from `app`? for instance: `app.GetTask(alloc.AllocationKey)`

pkg/cache/scheduler_callback.go
```diff
@@ -60,8 +60,15 @@ func (callback *AsyncRMCallback) UpdateAllocation(response *si.AllocationRespons
...
+				if utils.IsAssignedPod(task.GetTaskPod()) {
```
chia7712, line 64: Should we exclude the `failed` pod?

chia7712 (commented):
@craigcondit thanks for quick fix. a couple of comments below.

For another, how to write test for this bug? I guess it needs a new e2e test? If so, a follow-up jira is needed :)

pkg/cache/scheduler_callback.go
craigcondit: Either method works fine and performance isn't a concern here as this happens at most once per running task which was recovered during startup. My preference would be to leave it as-is.

pkg/cache/task.go
```diff
@@ -260,6 +260,22 @@ func (task *Task) SetTaskSchedulingState(state TaskSchedulingState) {
 	task.schedulingState = state
 }
 
+func (task *Task) MarkPreviouslyAllocated(allocationID string, nodeID string) {
+	task.sm.SetState(TaskStates().Bound)
```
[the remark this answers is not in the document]
craigcondit: The state machine does it's own internal locking.

pkg/cache/scheduler_callback.go
craigcondit: Failed pods are handled via rejections and won't be here.

chia7712 (approved):
LGTM

PR #788, opened 2024-02-14
[YUNIKORN-2417] Build against Kubernetes 1.27.2

craigcondit:
### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [x] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2417

### Questions:


go.mod
```diff
@@ -165,33 +165,33 @@ require (
...
-	k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.29.1
-	k8s.io/code-generator => k8s.io/code-generator v0.29.1
-	k8s.io/component-base => k8s.io/component-base v0.29.1
-	k8s.io/component-helpers => k8s.io/component-helpers v0.29.1
-	k8s.io/controller-manager => k8s.io/controller-manager v0.29.1
-	k8s.io/cri-api => k8s.io/cri-api v0.29.1
-	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.29.1
-	k8s.io/dynamic-resource-allocation => k8s.io/dynamic-resource-allocation v0.29.1
-	k8s.io/endpointslice => k8s.io/endpointslice v0.29.1
+	k8s.io/api => k8s.io/api v0.29.2
```
chia7712, line 168: This was what I try to figure out. Is the purpose to "lock" the version of `k8s.io/api`? I mean the version declared by `require` may be updated by other dependencies, so we want to build the code with "fixed" version.

craigcondit: K8s is a special case. Their official repo does some weirdness with overrides to a staging version of the various components. The go.mod they use has everything at simply v0.0.0. That is completely unworkable for us. This is the only viable solution.

wilfred-s: This same setup is also used by K8s projects that are not included in the base. An example can be found in the cluster autoscaler. 

This is the reason:
In the Kubernetes repository, the packages listed under require directive are tagged with a pseudo-version v0.0.0 but then are replaced with a local code in the staging directory. These replace directive works fine when building Kubernetes itself but won’t work when someone is importing it.
The go doc says
replace directives only apply in the main module’s go.mod file and are ignored in other modules.source.

There is thus no choice. It is _baked_ into the way K8s publishes its code

wilfred-s (approved):
LGTM

go.mod
chia7712: @craigcondit @wilfred-s thanks for all your explanation! 

The indirect dependencies need to be fixed with specify version, and that makes sense to me. However, My question was: why we need put those dependencies which are declared by `require` (for example: "k8s.io/api") into `replace`? We can just fix the `0.0.0` version of indirect dependencies instead of adding all k8s deps to simplify the go.mod.

craigcondit: Because the K8s components themselves utilize the v0.0.0 versions. It's a horrible mess. Using replace is the only way to ensure that we don't have one random transitive dependency at the wrong version, which at best leads to non-reproducible builds and at worst causes all sorts of compilation issues.

chia7712: > Because the K8s components themselves utilize the v0.0.0 versions. It's a horrible mess.

we are on the same page 😄 

> Using replace is the only way to ensure that we don't have one random transitive dependency at the wrong version, which at best leads to non-reproducible builds and at worst causes all sorts of compilation issues.

I agreed to use replace to fix "v0.0.0" issue, and now what I try to discuss is: can we "minimize" the `replace`? For example, only the following indirect deps are required.
```go
	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.29.2
	k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.29.2
	k8s.io/code-generator => k8s.io/code-generator v0.29.2
	k8s.io/endpointslice => k8s.io/endpointslice v0.29.2
	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.29.2
	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.29.2
	k8s.io/kube-proxy => kk8s.io/kube-proxy v0.29.2
	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.29.2
	k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.29.2
	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.29.2
```

Also, the version "v.0.0.0" can break the build so we can fix it in time if we ignore any indirect version.

craigcondit: It's a maintenance mess to "minimize" the replaces. We update to new K8s releases fairly often and chasing down breakage is a huge pain. I'd rather "over-specify" the replaces and just not have to worry about it. I've probably personally done the last 10 K8s updates -- this is the path of least resistance.

PR #794, opened 2024-02-22
[YUNIKORN-2416] Cleanup replace directives

chenyulin0719:
### What is this PR for?

- Cleanup unnecessarily golang.org/x/lint modules in replace directives. 
- Upgrade the the latest yunikorn-core/yunikorn-sheduler-interface
- Replace github.com/opencontainers/runc:v1.1.10 by v1.1.12 to fix [CVE-2024-21626](https://nvd.nist.gov/vuln/detail/CVE-2024-21626)

Note: There are 2 vulnerable dependencies existing before this PR, and it can't be fixed by replace directives. 
- go:gopkg.in/square/go-jose.v2:v2.6.0 is vulnerable (Cxb6dee8d5-b814)
 (From k8s.io/kubernetes@v1.29.2) (The packege is under public archive.)
- go:k8s.io/kubernetes:v1.29.2 is vulnerable (CVE-2020-8562)

And it required should be fixed in K8S.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
NA

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2416

1. Run CVE check (or use your IDE's vulnerability analysis tool).
2. Run E2E test with this new dependencis

E2E passed in my side:
https://github.com/chenyulin0719/yunikorn-k8shim/actions/runs/8004883059

There are 2 existing CVE and can't be fixed by replace.
image

CVE check pass with govulncheck:
image

### Questions:
NA


chia7712:
Had you run `go mod tidy` to update indirects?

wilfred-s (changes requested):
-1  for this approach. 
We must not touch the indirect require entries. Those are maintained by the code. We leave them as is and use a replace to override with what we want/need.
When updating and the indirect dependency is updated we need to check if the replace can be cleaned up.
The last couple of changes made to go.mod stepped away from this but we need to move back to that as we have no guarantee that the indirect references are stable and we could regress.

craigcondit (changes requested):
I'm also -1 on this. As I've previously indicated, I'd rather "over-specify" our replaces and ensure that we don't trigger regressions by omission.

chia7712:
@wilfred-s @craigcondit I have a question which is unrelated to this PR for "indirect".

It seems to me whole "indirect" should be generated by `go mod tidy`, but the "indirect" in k8shim does not. Do I misunderstand anything?

craigcondit:
> It seems to me whole "indirect" should be generated by `go mod tidy`, but the "indirect" in k8shim does not. Do I misunderstand anything?
What makes you think it doesn't work that way in k8shim?

chia7712:
> What makes you think it doesn't work that way in k8shim?

I remove all indicates and then run "go mod tidy". Some indirect deps get changed.

craigcondit:
> > What makes you think it doesn't work that way in k8shim?
> 
> I remove all indicates and then run "go mod tidy". Some indirect deps get changed.

That's expected - there were probably some indirect references that got updated via core changes. That's one of the reasons we try to be explicit in the replace stanza so that we don't accidentally regress due to a seemingly unrelated change.

chia7712:
> That's expected - there were probably some indirect references that got updated via core changes. That's one of the reasons we try to be explicit in the replace stanza so that we don't accidentally regress due to a seemingly unrelated change.

just curious. How to maintain/update the version of indirect dependency for k8shim if we don't rely on the `go mod tidy` (to auto-generate indirect dependencies)?

craigcondit:
> The version of some `indirect` dependencies get older after I re-create `indirect` stanza. For example: the version of `github.com/imdario/mergo` is changed from `v0.3.7` to `v0.3.6`. It seems to me the version `v0.3.6` is resolved by `go` and we should use it unless there is known CVE or regression, right?

No, we should keep the newer version unless there is something that breaks. It's not just CVE updates to consider.

chia7712:
> No, we should keep the newer version unless there is something that breaks. It's not just CVE updates to consider.

agree to that is not just CVE updates to consider. However, how to choose the suitable and newer version for it? for instance, why using `v0.3.7` rather than `v0.3.10`? Sorry for raising a bunch of questions. I try to learn the version management via this case.

chenyulin0719:
Base on above discussion, may I ask two questions to confirm the best practice for maintaining replace directives?

Q1: When should we upgrade version in replace directives
- New CVE detect
- Compatible issue.

Q2: When should we remove replace directives: 
- When all the indirect module version is higer than replace directive's version. (No rollback to older version)

chenyulin0719:
Thanks for the prompt reply.

> The version of some indirect dependencies get older after I re-create indirect stanza. For example: the version of github.com/imdario/mergo is changed from v0.3.7 to v0.3.6. It seems to me the version v0.3.6 is resolved by Go and we should use it unless there is known CVE or regression, right?

I just did an investigation to understand why we manually set github.com/imdario/mergo to v0.3.7 in indirect require entry instead of the version Go resolved.

The indirect require entry`github.com/imdario/mergo v0.3.7 // indirect` was added in 2203ad8, and come from spark-on-k8s-operator.

image


But after we remove spark-on-k8s-operator, we didn't regenerate indirect require entry.
`go mod tidy` just keep the newer version between before and after.
This is the root cause of the inconsistency.

image

In this case, I think the indirect require can be change to the v0.3.6.
If in any reason we must keep v0.3.7, we should put it in replace directive and comments.

There are some other inconsistency.  The review might be exhausting, but it should be an one-time job.
After that we can ensure indirect require is equal to what Go resolved.

craigcondit:
> 1. Manually remove indirect require entry
> 2. Run `go mod tidy`
> 3. Run tests
> 4. If anything wrong, put the fix version in replace entry and add comment

Sorry, I have to respectfully disagree here. This is  not only time consuming, but error prone, because every time this is done we need to audit for CVEs that might have been inadvertently reintroduced. This is not a hypothetical issue - it has happened on several occasions.

> Pros:
> We know what module version is come from parent module's require.

Why is this important? We override these to pull newer versions, and if those newer versions cause no issues, downgrading is pointless.

> We document down what compatible/security issue happened before in this project.

This isn't something we do, nor is there a central place to track this within the project.

The very real cons of this approach significantly dwarf any of the perceived benefits.

chenyulin0719:
Understood, I can't find other reason to support it. 
Thanks for sharing your viewpoint. :)

For this PR, I'm not sure what should be fixed. I didn't manually change the indirect require entry. (It was change by `go mod tidy` with newer core version) 
Should I move golang.org/x/lint back?

For k8shim and sheduler-interface, I've removed the golang.org/x modules in replace entry. 
- Core: https://github.com/apache/yunikorn-core/commit/e81cf42a3981b95eb96d51ee0c57a996fea28432
- sheduler-interface: https://github.com/apache/yunikorn-scheduler-interface/commit/e404c64334d25156a32b5af4a68faaa653b9c457
Should I add a commit to move it back?

craigcondit:
> For this PR, I'm not sure what should be fixed. I didn't manually change the indirect require entry. (It was change by `go mod tidy` with newer core version) Should I move golang.org/x/lint back?

Put the x/lint replace back; also remove the comment from the other added replace.


> For core and sheduler-interface, I've removed the golang.org/x modules in replace entry.
> 
> * Core: [apache/yunikorn-core@e81cf42](https://github.com/apache/yunikorn-core/commit/e81cf42a3981b95eb96d51ee0c57a996fea28432)
> * sheduler-interface: [apache/yunikorn-scheduler-interface@e404c64](https://github.com/apache/yunikorn-scheduler-interface/commit/e404c64334d25156a32b5af4a68faaa653b9c457)
> 
> Should I add a commit to move it back?

Yes, please open JIRAs and PRs to replace the overrides that were already in place.

craigcondit:
@chenyulin0719 - The core / shim PRs have been. merged. In addition to the other changes requested, can you please update the scheduler-interface and core references to the latest? These are:

core: v0.0.0-20240222204527-4289253dc78e
si: v0.0.0-20240222194711-770b7b60b5c6

craigcondit (approved):
+1 pending pre-commit checks.

chia7712 (approved):
LGTM

go.mod
```diff
@@ -164,7 +160,12 @@ require (
 )
 
 replace (
-	golang.org/x/lint => golang.org/x/lint v0.0.0-20210508222113-6edffad5e616
```
chia7712, line 167: sorry that I have to raise the comment again.

We should put x/lint back, right? otherwise, the version will get downgrade from `v0.0.0-20210508222113-6edffad5e616` to `v0.0.0-20190313153728-d0100b6bd8b3`. It seems to me the most important conclusion is that we don't give up the newer version against which we've successfully tested.

craigcondit: Good catch. Let's put that back.

chenyulin0719: Sorry, let me fix it.

craigcondit (approved):
+1 pending e2e tests.

chia7712 (approved):
LGTM

PR #796, opened 2024-03-05
[YUNIKORN-2465] Remove Task objects from the shim upon pod completion

pbacsko:
### What is this PR for?
Remove completed Tasks from the Application to reduce memory pressure for long running applications.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2465

### Questions:

### DIFF
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -228,9 +228,13 @@ func (app *Application) addTask(task *Task) {
 	app.taskMap[task.taskID] = task
 }
 
-func (app *Application) removeTask(taskID string) {
+func (app *Application) RemoveTask(taskID string) {
 	app.lock.Lock()
 	defer app.lock.Unlock()
+	app.removeTask(taskID)
+}
+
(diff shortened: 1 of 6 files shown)


pkg/cache/application.go
```diff
@@ -363,6 +363,7 @@ func (app *Application) Schedule() bool {
 		app.scheduleTasks(func(t *Task) bool {
 			return t.placeholder
 		})
+		app.removeCompletedTasks()
```
pbacsko, line 366: We need to remove it from here to let the state transitions run for the Task. If we removing it prematurely, it cannot reach Completed state. Another solution is to call `Application.removeTask()` from the task itself but this seems cleaner.

craigcondit (approved):
+1 LGTM.

PR #803, opened 2024-03-13
[YUNIKORN-2485] Shim: Use Docker to build reproducible binaries

craigcondit:
### What is this PR for?
Introduces a build var (REPRODUCIBLE_BUILDS=1) to force building binaries using Docker. This results in builds with a consistent environment every time.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [x] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2485

### Questions:


Makefile
```diff
@@ -408,6 +430,17 @@ plugin: $(RELEASE_BIN_DIR)/$(PLUGIN_BINARY)
...
+	docker run -t --rm=true --volume "$(BASE_DIR):/buildroot" "golang:$(GO_REPRO_VERSION)" sh -c "cd /buildroot && \
+	CGO_ENABLED=0 GOOS=linux GOARCH=\"${EXEC_ARCH}\" go build \
+	-a \
+	-o=${RELEASE_BIN_DIR}/${PLUGIN_BINARY} \
+	-trimpath \
+	-ldflags '-buildid= -extldflags \"-static\" -X ${FLAG_PREFIX}.buildVersion=${VERSION} -X ${FLAG_PREFIX}.buildDate=${DATE} -X ${FLAG_PREFIX}.isPluginVersion=true -X ${FLAG_PREFIX}.goVersion=${GO_REPRO_VERSION} -X ${FLAG_PREFIX}.arch=${EXEC_ARCH} -X ${FLAG_PREFIX}.coreSHA=${CORE_SHA} -X ${FLAG_PREFIX}.siSHA=${SI_SHA} -X ${FLAG_PREFIX}.shimSHA=${SHIM_SHA}' \
+	-tags netgo \
+	-installsuffix netgo \
+	./pkg/cmd/schedulerplugin/"
+else
```
chia7712, line 443: Do we need to maintain the old way of build? 

The old way does have an advantage - it does not to re-download all dependencies :_

Makefile
```diff
@@ -74,6 +74,16 @@ else
   PLUGIN_OPTS := 
 endif
 
+# Reproducible builds mode
+GO_REPRO_VERSION := $(shell cat .go_repro_version)
+GO_REPRO_MAJOR := $(word 1,$(subst ., ,$(GO_REPRO_VERSION)))
+GO_REPRO_MINOR := $(word 2,$(subst ., ,$(GO_REPRO_VERSION)))
```
chia7712, line 80: `GO_REPRO_MINOR` seems to be unused.

Makefile
craigcondit: That was my thought. Downloading all the deps and not caching build artifacts is a little excessive for day-to-day dev work. So I think keeping both is the best plan.

Makefile
craigcondit: So it is... I can remove that on commit.

chia7712 (approved):
LGTM

PR #810, opened 2024-04-01
[YUNIKORN-2520] PVC errors in AssumePod() are not handled properly

pbacsko:
### What is this PR for?
Handle PV/PVC volume errors  in `Context.AssumePod()` properly.
Enhance test coverage.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2520

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -779,7 +779,7 @@ func (ctx *Context) bindPodVolumes(pod *v1.Pod) error {
 // be running on it. And we keep this cache in-sync between core and the shim.
 // this way, the core can make allocation decisions with consideration of
 // other assumed pods before they are actually bound to the node (bound is slow).
-func (ctx *Context) AssumePod(name, applicationID, allocationID, node string) error {
+func (ctx *Context) AssumePod(name, node string) error {
 	ctx.lock.Lock()
 	defer ctx.lock.Unlock()
 	if pod, ok := ctx.schedulerCache.GetPod(name); ok {
@@ -792,13 +792,6 @@ func (ctx *Context) AssumePod(name, applicationID, allocationID, node string) er
 			// assume pod volumes before assuming the pod
 			// this will update scheduler cache with essential PV/PVC binding info
 			var allBound = true
-			task := ctx.getTask(applicationID, name)
-			if task == nil {
-				log.Log(log.ShimContext).Error("BUG: task not found", zap.String("taskID", name))
-				return fmt.Errorf("task not found: %s", name)
-			}
-			task.setAllocationID(allocationID)
-
 			var err error
 			// retrieve the volume claims
 			podVolumeClaims, err := ctx.apiProvider.GetAPIs().VolumeBinder.GetPodVolumeClaims(ctx.klogger, pod)
@@ -806,7 +799,6 @@ func (ctx *Context) AssumePod(name, applicationID, allocationID, node string) er
 				log.Log(log.ShimContext).Error("Failed to get pod volume claims",
 					zap.String("podName", assumedPod.Name),
 					zap.Error(err))
-				task.failWithEvent(fmt.Sprintf("Failed to get pod volume claims: %s", err.Error()), "PodVolumeClaimsNotFound")
 				return err
 			}
 
@@ -817,7 +809,6 @@ func (ctx *Context) AssumePod(name, applicationID, allocationID, node string) er
 					zap.String("podName", assumedPod.Name),
 					zap.String("nodeName", assumedPod.Spec.NodeName),
 					zap.Error(err))
-				task.failWithEvent(fmt.Sprintf("Failed to get pod volumes: %s", err.Error()), "PodVolumeNotFound")
 				return err
 			}
 			if len(reasons) > 0 {
@@ -831,12 +822,10 @@ func (ctx *Context) AssumePod(name, applicationID, allocationID, node string) er
 					zap.String("podName", assumedPod.Name),
 					zap.String("nodeName", assumedPod.Spec.NodeName),
 					zap.Error(err))
-				task.failWithEvent(err.Error(), "ConflictingVolumeClaims")
 				return err
 			}
 			allBound, err = ctx.apiProvider.GetAPIs().VolumeBinder.AssumePodVolumes(ctx.klogger, pod, node, volumes)
 			if err != nil {
-				task.failWithEvent(fmt.Sprintf("Failed to assume pod volumes: %s", err.Error()), "AssumePodVolumeError")
 				return err
 			}
 
@@ -1161,13 +1150,9 @@ func (ctx *Context) RemoveTask(appID, taskID string) {
 	app.RemoveTask(taskID)
 }
 
-func (ctx *Context) GetTask(appID string, taskID string) *Task {
+func (ctx *Context) getTask(appID string, taskID string) *Task {
 	ctx.lock.RLock()
 	defer ctx.lock.RUnlock()
-	return ctx.getTask(appID, taskID)
-}
-
-func (ctx *Context) getTask(appID, taskID string) *Task {
 	app := ctx.getApplication(appID)
 	if app == nil {
 		log.Log(log.ShimContext).Debug("application is not found in the context",
@@ -1203,7 +1188,7 @@ func (ctx *Context) PublishEvents(eventRecords []*si.EventRecord) {
 			case si.EventRecord_REQUEST:
 				appID := record.ReferenceID
 				taskID := record.ObjectID
-				if task := ctx.GetTask(appID, taskID); task != nil {
+				if task := ctx.getTask(appID, taskID); task != nil {
 					events.GetRecorder().Eventf(task.GetTaskPod().DeepCopy(), nil,
 						v1.EventTypeNormal, "", "", record.Message)
 				} else {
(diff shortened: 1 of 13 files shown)


pkg/cache/context.go
```diff
@@ -792,45 +792,54 @@ func (ctx *Context) AssumePod(name string, node string) error {
 			// assume pod volumes before assuming the pod
 			// this will update scheduler cache with essential PV/PVC binding info
 			var allBound = true
-			// volume builder might be null in UTs
-			if ctx.apiProvider.GetAPIs().VolumeBinder != nil {
```
pbacsko, line 796: Removed "nil" check, now a VolumeBinder instance is always set.

pkg/cache/task.go
```diff
@@ -379,14 +379,12 @@ func (task *Task) postTaskAllocated() {
 			log.Log(log.ShimCacheTask).Debug("bind pod volumes",
 				zap.String("podName", task.pod.Name),
 				zap.String("podUID", string(task.pod.UID)))
-			if task.context.apiProvider.GetAPIs().VolumeBinder != nil {
```
pbacsko, line 382: Removed "nil" check, now a VolumeBinder instance is always set.

pkg/cache/task.go
```diff
@@ -523,18 +516,16 @@ func (task *Task) releaseAllocation() {
 		s := TaskStates()
 		switch task.GetTaskState() {
 		case s.New, s.Pending, s.Scheduling, s.Rejected:
-			releaseRequest = common.CreateReleaseAskRequestForTask(
-				task.applicationID, task.taskID, task.application.partition)
+			releaseRequest = common.CreateReleaseRequestForTask(task.applicationID, task.taskID, task.allocationID, task.application.partition, task.terminationType)
```
pbacsko, line 519: A single method to create a release request. From now on we'll always try to release both ask+allocation if possible.

pkg/cache/context.go
```diff
@@ -1152,9 +1161,13 @@ func (ctx *Context) RemoveTask(appID, taskID string) {
...
 }
 
-func (ctx *Context) getTask(appID string, taskID string) *Task {
+func (ctx *Context) GetTask(appID string, taskID string) *Task {
 	ctx.lock.RLock()
 	defer ctx.lock.RUnlock()
+	return ctx.getTask(appID, taskID)
+}
+
+func (ctx *Context) getTask(appID, taskID string) *Task {
```
wilfred-s, line 1170: This is not needed if you first retrieve the task using `getTask` before calling `AssumePod` in the tests. Also becomes unneeded if we handle the fail in the `UpdateAllocation`

pkg/cache/context.go
```diff
@@ -792,45 +792,54 @@ func (ctx *Context) AssumePod(name string, node string) error {
...
-						zap.String("podName", assumedPod.Name),
-						zap.Error(err))
-					return err
-				}
+			task := ctx.getTask(applicationID, name)
+			if task == nil {
+				log.Log(log.ShimContext).Error("BUG: task not found", zap.String("taskID", name))
+				return fmt.Errorf("task not found: %s", name)
+			}
+			task.setAllocationID(allocationID)
```
wilfred-s, line 800: This can be prevented if we handle the error returned in UpdateAllocation to fail the task etc.

pkg/cache/external/scheduler_cache.go
```diff
@@ -450,7 +450,7 @@ func (cache *SchedulerCache) StartPodAllocation(podKey string, nodeID string) bo
 }
 
 // return if pod is assumed in cache, avoid nil
-func (cache *SchedulerCache) isAssumedPod(podKey string) bool {
```
craigcondit, line 453: If we're making this public, we should keep the existing function name, and wrap with an IsAssumedPod() which takes a lock. Otherwise, we risk external callers not locking.

craigcondit (changes requested):
Mostly ok. There's a couple of issues I called out in comments.

pkg/cache/context_test.go
```diff
@@ -2138,6 +2142,168 @@ func TestTaskRemoveOnCompletion(t *testing.T) {
...
+func TestAssumePod_AssumePodVolumesError(t *testing.T) {
+	binder := test.NewVolumeBinderMock()
+	const errMsg = "error assuming pod volumes"
+	binder.SetAssumePodVolumesError(errMsg)
+	context := initAssumePodTest(binder)
+	defer dispatcher.UnregisterAllEventHandlers()
+	defer dispatcher.Stop()
+	err := context.AssumePod(pod1UID, appID, "alloc-0", fakeNodeName)
+	assert.Error(t, err, errMsg)
+	task := context.getTask(appID, pod1UID)
```
wilfred-s, line 2244: swap order getTask before AssumePod

pkg/cache/context_test.go
```diff
@@ -2138,6 +2142,168 @@ func TestTaskRemoveOnCompletion(t *testing.T) {
...
+func TestAssumePod_ConflictingVolumes(t *testing.T) {
+	binder := test.NewVolumeBinderMock()
+	binder.SetConflictReasons("reason1", "reason2")
+	context := initAssumePodTest(binder)
+	defer dispatcher.UnregisterAllEventHandlers()
+	defer dispatcher.Stop()
+
+	err := context.AssumePod(pod1UID, appID, "alloc-0", fakeNodeName)
+	assert.Error(t, err, "pod my-pod-1 has conflicting volume claims: reason1, reason2")
+	task := context.getTask(appID, pod1UID)
```
wilfred-s, line 2224: swap order getTask before AssumePod

pkg/cache/external/scheduler_cache.go
```diff
@@ -450,7 +450,7 @@ func (cache *SchedulerCache) StartPodAllocation(podKey string, nodeID string) bo
...
+func (cache *SchedulerCache) IsAssumedPod(podKey string) bool {
```
wilfred-s, line 453: No need to export, callers have not changed

wilfred-s (changes requested):
Instead of pushing the `allocationID` into `AssumePod` and fail the task from there we can fail the task in `UpdateAllocation` when the error is returned from `AssumePod`. Adding a second return value if we need  more detail is a simpler change.
Means no changes in the `task.go` needed. Removes the need for a locked and unlocked `getTask` etc

pkg/cache/context.go
pbacsko: Done

pkg/cache/context.go
pbacsko: Done

pbacsko:
> Instead of pushing the `allocationID` into `AssumePod` and fail the task from there we can fail the task in `UpdateAllocation` when the error is returned from `AssumePod`. Adding a second return value if we need more detail is a simpler change. Means no changes in the `task.go` needed. Removes the need for a locked and unlocked `getTask` etc

I made the changes to simplify the PR.

I also realized that there's no unit test code for `scheduler_callback.go` (`AsyncRMCallback`) which bothers me a bit. It's tested indirectly but we don't see if we miss anything. I suggest creating tests for it.

pkg/cache/context.go
```diff
@@ -792,45 +792,43 @@ func (ctx *Context) AssumePod(name string, node string) error {
...
+				log.Log(log.ShimContext).Error("Failed to find pod volumes",
+					zap.String("podName", assumedPod.Name),
+					zap.String("nodeName", assumedPod.Spec.NodeName),
+					zap.Error(err))
+				return err
+			}
+			if len(reasons) > 0 {
+				sReasons := make([]string, 0)
+				for _, reason := range reasons {
+					sReasons = append(sReasons, string(reason))
```
wilfred-s, line 817: This should be optimised to:
```
sReasons := make([]string, len(reasons))
for i, reason := range reasons {
    sReasons[i] = string(reason)
}
```

pkg/cache/scheduler_callback.go
```diff
@@ -56,11 +56,20 @@ func (callback *AsyncRMCallback) UpdateAllocation(response *si.AllocationRespons
...
+		}
 		if err := callback.context.AssumePod(alloc.AllocationKey, alloc.NodeID); err != nil {
+			if task != nil {
+				task.failWithEvent(err.Error(), "AssumePodError")
+			}
 			return err
 		}
 		if app := callback.context.GetApplication(alloc.ApplicationID); app != nil {
-			if task := callback.context.getTask(app.GetApplicationID(), alloc.AllocationKey); task != nil {
+			if task != nil {
```
wilfred-s, line 72: This needs a cleanup: `getTask()` already checks for the application. On top of that we have the appID already in the task so we do not need to get it from the app. That needs to be applied to the whole function (new jira)

wilfred-s (approved):

wilfred-s:
> I also realized that there's no unit test code for `scheduler_callback.go` (`AsyncRMCallback`) which bothers me a bit. It's tested indirectly but we don't see if we miss anything. I suggest creating tests for it.

Please file a jira for that. I filed two simple jiras to clean up the code in this same area a bit.

PR #811, opened 2024-04-01
[YUNIKORN-2525] dispatcher.Stop() waits an extra second unnecessarily

pbacsko:
### What is this PR for?
Speed up `dispatcher.Stop()` which also makes unit tests run faster.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2525

### Questions:

### DIFF
--- a/pkg/dispatcher/dispatcher.go
+++ b/pkg/dispatcher/dispatcher.go
@@ -56,6 +56,7 @@ type Dispatcher struct {
 	handlers  map[EventType]map[string]func(interface{})
 	running   atomic.Value
 	lock      sync.RWMutex
+	stopped   sync.WaitGroup
 }
 
 func initDispatcher() {
@@ -211,6 +212,7 @@ func Start() {
 		return
 	}
 	getDispatcher().stopChan = make(chan struct{})
+	getDispatcher().stopped.Add(1)
 	go func() {
 		for {
 			select {
@@ -229,6 +231,7 @@ func Start() {
 			case <-getDispatcher().stopChan:
 				log.Log(log.ShimDispatcher).Info("shutting down event channel")
 				getDispatcher().setRunning(false)
+				getDispatcher().stopped.Done()
 				return
 			}
 		}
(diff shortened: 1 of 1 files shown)


pkg/dispatcher/dispatcher.go
```diff
@@ -257,13 +260,23 @@ func Stop() {
...
-		log.Log(log.ShimDispatcher).Info("waiting for dispatcher to be stopped",
-			zap.Int("remainingSeconds", maxTimeout))
-		time.Sleep(1 * time.Second)
-		maxTimeout--
+	stopWait := make(chan struct{})
+	defer close(stopWait)
+
+	go func() {
+		getDispatcher().stopped.Wait()
+		stopWait <- struct{}{}
```
chia7712, line 268: We can close the channel directly here.

chia7712 (approved):
overall +1 and one minor comment left.

pkg/dispatcher/dispatcher.go
chia7712:
```go
	go func() {
		defer close(stopWait)
		getDispatcher().stopped.Wait()
	}()
```

pbacsko: That would mean double closure, which results in a panic. Closing inside `Stop()` seems to be a better place to me.

chia7712: > That would mean double closure

pardon me. why it causes double closure if we remove the `close(stopWait)` from `Stop()`.

```go
	stopWait := make(chan struct{})
        // defer close(stopWait) move this into goroutine

	go func() {
		defer close(stopWait)
		getDispatcher().stopped.Wait()
	}()

	// wait until the main event loop stops properly
	select {
	case <-stopWait:
		break
	case <-time.After(5 * time.Second):
		log.Log(log.ShimDispatcher).Info("dispatcher did not stop in time")
		break
	}
```

pbacsko: > pardon me. why it causes double closure if we remove the close(stopWait) from Stop().

In that case it doesn't. I just checked and a general principle is that channel shouldn't be closed on a receiver side. I don't think it's a big deal, but ok, I'll change it.

chia7712 (approved):

chenyulin0719 (approved):
+1 LGTM

PR #814, opened 2024-04-05
[YUNIKORN-2539] Shim: Add deadlock tracking feature

craigcondit:
### What is this PR for?

Replaces sync.{RW}Mutex with internal locking.{RW}Mutex implementations. The new implementation wraps the go-deadlock library with logic to conditionally enable deadlock detection based on the presence of environment variables:

To enable the feature:

- DEADLOCK_DETECTION_ENABLED=true

To customize the timeout before potential deadlocks are logged (default is 60 seconds):

- DEADLOCK_TIMEOUT_SECONDS=60

See https://github.com/sasha-s/go-deadlock for more details.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2539

Unit tests added which use the new locks.

### Questions:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ module github.com/apache/yunikorn-k8shim
 go 1.21
 
 require (
-	github.com/apache/yunikorn-core v0.0.0-20240402212227-bdf109b5432c
+	github.com/apache/yunikorn-core v0.0.0-20240405153113-5758d7ac3c85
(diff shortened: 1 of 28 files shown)


pbacsko (commented):
+1

PR #817, opened 2024-04-13
[YUNIKORN-2538]Shim cache context pre-allocate slice

SophieTech88:
### What is this PR for?
When building the reason string from all volume failure reasons we should allocate a slice once based on the size of the reasons object we get returned.

### What type of PR is it?
* [x] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2538

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -802,9 +802,9 @@ func (ctx *Context) AssumePod(name, node string) error {
 				return err
 			}
 			if len(reasons) > 0 {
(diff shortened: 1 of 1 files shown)


brandboat (approved):
LGTM !

PR #818, opened 2024-04-13
[YUNIKORN-2557]Update the doc for e2e-tests

SophieTech88:
YUNIKORN-2557

### What is this PR for?
When I want to run the e2e-tests for the first time, it is a time consuming process to install the tools one by one, then learn from the failure to figure out I need to link the path of each tool to the tools/ folder as pre-requisite step. And there is an easy way to run the command `make tools` to set up everything we need before we run the e2e-tests.

So it is great to update the doc for e2e-tests to set up the environment in an easy way.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2557

### Questions:

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -34,6 +34,8 @@ Below is the structure of the project.
 This project requires Go to be installed. On OS X with Homebrew you can just run `brew install go`.
 OR follow this doc for deploying go https://golang.org/doc/install
 
+Before we run the e2e tests, we need to set up the environment and install the relates tools using the command `make tools`. It will install the related tools including ginkgo, golangci-lint, helm, kind, kubectl, shellcheck and so on. At the same time it will link the path of the tools under the `./tools/` folder.
+
 ## Understanding the Command Line Arguments
 * `yk-namespace` - namespace under which YuniKorn is deployed. [Required]
 * `kube-config` - path to kube config file, needed for k8s client [Required]


craigcondit:
The makefile should already have the e2e-tests target depend on the tools. If it doesn't that's the proper fix.

chia7712:
> The makefile should already have the e2e-tests target depend on the tools. If it doesn't that's the proper fix.

The script skips the `make tools` if the tools folder is existent. That is a bit weird that we don't check all tools in the folder. Maybe the script should call `make tools` anyway, and let `make tools` do the check and install.

https://github.com/apache/yunikorn-k8shim/blob/bda57dda1a461517491d849bf8e3196ead3416a1/scripts/run-e2e-tests.sh#L74

craigcondit:
Yes that should be made unconditional. It should only install the ones that are missing anyway.

ryankert01:
+1 on removing the conditional. I've encountered this problem before.

ryankert01:
> +1 on removing the conditional(or the `install_tools` function, and use `make tools`). I've encountered this problem before.

opened a jira regarding this:
https://issues.apache.org/jira/browse/YUNIKORN-2558

chia7712:
#819 gets merged so I'm going to close this PR.

PR #820, opened 2024-04-17
Download ginkgo to tools folder

SophieTech88:
### What is this PR for?
`run-e2e-tests.sh` install the ginkgo in the `GOBIN` folder. However, our `make e2e_test` assumes all tools are installed at yunikorn-k8shim/tools, and hence we can see `Command 'ginkgo' not found` error if we don't export the `GOBIN`.

Actions:
* Update the Makefile to download ginkgo under tools folder
* Delete the ginkgo download and check in run-e2e-tests.sh file , and correct the steps order.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2560

### Questions:


Makefile
```diff
@@ -174,6 +174,10 @@ export SPARK_PYTHON_IMAGE=docker.io/apache/spark-py:v$(SPARK_PYTHON_VERSION)
 GO_LICENSES_VERSION=v1.6.0
 GO_LICENSES_BIN=$(TOOLS_DIR)/go-licenses
 
+# ginkgo
+GINKGO_VERSION=v2.17.1
```
chia7712, line 178: the version of `ginkgo` is defined by `go.mod`, and it would be better to keep them consistency. Hence, we don't need to define the specific version in makefile.

Makefile
```diff
@@ -273,6 +277,11 @@ $(GO_LICENSES_BIN):
 	@mkdir -p "$(TOOLS_DIR)"
 	@GOBIN="$(BASE_DIR)/$(TOOLS_DIR)" "$(GO)" install "github.com/google/go-licenses@$(GO_LICENSES_VERSION)"
 
+$(GINKGO_BIN):
+	@echo "installing go-licenses $(GINKGO_VERSION)"
```
chia7712, line 281: `go-licenses` -> `ginkgo`

chia7712 (commented):
@SophieTech88 thanks for this improvement. The error of `ginkgo not found` is panic to me before. Please take a look at following comments.

pbacsko:
Thanks for the PR @SophieTech88. Please make sure that the PR title and the commit contains the JIRA number:

`[YUNIKORN-2560] Download ginkgo to tools folder`

Thanks.

Makefile
SophieTech88: Oh . Sorry to make a stupid mistake.

Makefile
SophieTech88: Great. Delete the version of ginkgo in makefile.

craigcondit: Will `go install` honor the go.mod file?

craigcondit: It's better to match what's in go.mod instead. Perhaps extract from go.mod using a short bit of shell and explicitly install that version instead.

craigcondit (changes requested):
See my comments around version.

Makefile
craigcondit: This should extract it:

`GINKGO_VERSION := $(shell grep ginkgo go.mod | head -1 | awk '{print $$2}')`

chia7712: @craigcondit according to the docs (https://go.dev/ref/mod#go-install)
```
# Install a program at the version selected by the module in the current directory.
$ go install golang.org/x/tools/gopls
```
and

```
If the arguments don’t have version suffixes, go install may run in module-aware mode or GOPATH mode, depending on the GO111MODULE environment variable and the presence of a go.mod file.
```

Also, I have test `go install github.com/onsi/ginkgo/v2/ginkgo` on my local. the version is equal to what we define in go.mod

craigcondit: Ok, I stand corrected.

craigcondit (approved):
+1 pending e2e test completion successfully.

pbacsko (approved):
LGTM

chia7712 (approved):
LGTM

PR #826, opened 2024-04-24
[YUNIKORN-2577] Remove named returns from IsPodFitNodeViaPreemption

ryankert01:
### What is this PR for?
IsPodFitNodeViaPreemption has defined named returns but does not use them. They should be removed as the way they are used can cause issues that are hard to debug.

As part of this change we need to further cleanup:  
- The variable ok also gets shadowed multiple times, not just from the named return declaration.
- The if construct around GetPodNoLock() is not needed as it returns a nil for the pod if it returns false. Just adding the result for the pod always has the same effect.

**I did following changes:**
1. remove named returns
2. to avoid shadowing, I try not to use `ok`, instead, those function can represent `ok` by its first return value, whether if it's `index` or `pod`, `-1` and `nil` means `false` respectively.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2577

```
make test
```

N/A

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -679,10 +679,10 @@ func (ctx *Context) IsPodFitNodeViaPreemption(name, node string, allocations []s
 			defer ctx.schedulerCache.UnlockForReads()
 
 			// look up each victim in the scheduler cache
-			victims := make([]*v1.Pod, 0)
-			for _, uid := range allocations {
+			victims := make([]*v1.Pod, len(allocations))
+			for index, uid := range allocations {
 				victim, _ := ctx.schedulerCache.GetPodNoLock(uid)
-				victims = append(victims, victim)
+				victims[index] = victim
 			}
 
 			// check predicates for a match
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -663,15 +663,15 @@ func (ctx *Context) IsPodFitNode(name, node string, allocate bool) error {
 	return err
 }
 
-func (ctx *Context) IsPodFitNodeViaPreemption(name, node string, allocations []string, startIndex int) (index int, ok bool) {
+func (ctx *Context) IsPodFitNodeViaPreemption(name, node string, allocations []string, startIndex int) (int, bool) {
 	// assume minimal pods need killing if running in testing mode
 	if ctx.apiProvider.IsTestingMode() {
-		return startIndex, ok
+		return startIndex, false
 	}
 
 	ctx.lock.RLock()
 	defer ctx.lock.RUnlock()
-	if pod, ok := ctx.schedulerCache.GetPod(name); ok {
+	if pod, _ := ctx.schedulerCache.GetPod(name); pod != nil {
 		// if pod exists in cache, try to run predicates
 		if targetNode := ctx.schedulerCache.GetNode(node); targetNode != nil {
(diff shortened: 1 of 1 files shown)


pkg/cache/context.go
```diff
@@ -681,17 +681,13 @@ func (ctx *Context) IsPodFitNodeViaPreemption(name, node string, allocations []s
 			// look up each victim in the scheduler cache
 			victims := make([]*v1.Pod, 0)
```
chia7712, line 682: Could we set the size?

pkg/cache/context.go
```diff
@@ -681,17 +681,13 @@ func (ctx *Context) IsPodFitNodeViaPreemption(name, node string, allocations []s
...
-					victims = append(victims, nil)
-				}
+				victim, _ := ctx.schedulerCache.GetPodNoLock(uid)
+				victims = append(victims, victim)
 			}
 
 			// check predicates for a match
-			if index, ok := ctx.predManager.PreemptionPredicates(pod, targetNode, victims, startIndex); ok {
-				return index, ok
+			if index, _ := ctx.predManager.PreemptionPredicates(pod, targetNode, victims, startIndex); index != -1 {
```
chia7712, line 689: It seems `-1` is equal to "failure". Maybe we don't return extra `boolean` value.

chia7712 (commented):
@ryankert01 thanks for this refactor. two minor comments left.

pkg/cache/context.go
ryankert01: It looks like we can refactor PreemptionPredicates(). I will file a jira for it.

pkg/cache/context.go
ryankert01: Yes, we can initialized the size first to further enhance the performance.

ryankert01: Done.

pkg/cache/context.go
wilfred-s: The predicate manager also uses a mixture of named returns and just types. We should not use named returns there either. Filed a new jira for that one and a PR raised. The change to remove the boolean should be another jira.

wilfred-s (approved):
LGTM

pkg/cache/context.go
```diff
@@ -663,35 +663,31 @@ func (ctx *Context) IsPodFitNode(name, node string, allocate bool) error {
 	return err
 }
 
-func (ctx *Context) IsPodFitNodeViaPreemption(name, node string, allocations []string, startIndex int) (index int, ok bool) {
+func (ctx *Context) IsPodFitNodeViaPreemption(name, node string, allocations []string, startIndex int) (int, bool) {
 	// assume minimal pods need killing if running in testing mode
 	if ctx.apiProvider.IsTestingMode() {
-		return startIndex, ok
+		return startIndex, false
```
chia7712, line 669: This will get removed by #832, so we can revert this change to avoid conflicts later

chia7712: I guess @wilfred-s is going to merge it, so please ignore my comment :)

PR #827, opened 2024-04-24
[YUNIKORN-2586] Set GOROOT to avoid tool failure when using newer go compilers

craigcondit:
### What is this PR for?
Compiling with newer go compilers (tested on go 1.22.2) fails when generating 3rd party licenses. This was traced back to GOROOT not being set. Set it in the Makefile to avoid this issue.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2586

### Questions:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,9 @@ ifeq ($(GO),)
 GO := go
 endif
 
-GO_EXE_PATH := $(shell "$(GO)" env GOROOT)/bin
+GOROOT := $(shell "$(GO)" env GOROOT)
+export GOROOT := $(GOROOT)
+GO_EXE_PATH := $(GOROOT)/bin
 
 # Check if this GO tools version used is at least the version of go specified in
 # the go.mod file. The version in go.mod should be in sync with other repos.


chia7712:
@craigcondit just curios. how to reproduce the error? There is no `GOROOT` and then I run `make clean && make go-license-generate`. It works well.
螢幕快照 2024-04-25 06-44-27

craigcondit:
Not sure but it's 100% reproducible for me without the patch.

craigcondit:
Apparently go-license sometimes has difficulty detecting GOROOT properly under certain cases. There an open issue on it, but the symptom is that it misdetects standard library packages as non-modules and breaks the build.

chia7712 (approved):
Exporting `GOROOT` does not burn out server, and it works well on my local. +1

pbacsko (approved):
+1

PR #829, opened 2024-04-30
[YUNIKORN-2597] Improve error messages in Context

pbacsko:
### What is this PR for?
Improve event handler related logging for `cache.Context`.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2597

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


FrankYang0529 (approved):
LGTM.

pkg/cache/context.go
```diff
@@ -1294,41 +1295,68 @@ func (ctx *Context) HandleContainerStateUpdate(request *si.UpdateContainerSchedu
...
 			if app.canHandle(event) {
 				if err := app.handle(event); err != nil {
 					log.Log(log.ShimContext).Error("failed to handle application event",
 						zap.String("event", event.GetEvent()),
+						zap.String("applicationID", appID),
 						zap.Error(err))
+					return
 				}
 			}
+			log.Log(log.ShimContext).Error("application event cannot be handled in the current state",
```
chia7712, line 1315: This log message is weird to me since the application could be get handled successfully rather than "cannot be handled" in this path. Maybe we need to add return to `if app.canHandle(event) {`

pkg/cache/context.go
```diff
@@ -1294,41 +1295,68 @@ func (ctx *Context) HandleContainerStateUpdate(request *si.UpdateContainerSchedu
...
-						zap.String("applicationID", task.applicationID),
-						zap.String("taskID", task.taskID),
+						zap.String("applicationID", appID),
+						zap.String("taskID", taskID),
 						zap.String("event", event.GetEvent()),
 						zap.Error(err))
+					return
 				}
 			}
+			log.Log(log.ShimContext).Error("task event cannot be handled in the current state",
```
chia7712, line 1350: ditto

chia7712 (commented):
@pbacsko thanks for this patch. two comments left. PTAL

pkg/cache/context.go
pbacsko: Correct, I modified that. Initially it was a bunch of if-else which was hard to read, I do hope this is better.

pkg/cache/context.go
pbacsko: Fixed.

chia7712 (approved):
LGTM

PR #830, opened 2024-05-03
[YUNIKORN-2599] Certain shim events are not handled by the state machine

pbacsko:
### What is this PR for?
Certain events are not specified in the state machine, but we keep sending them.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2599

### Questions:

### DIFF
--- a/pkg/cache/application_state.go
+++ b/pkg/cache/application_state.go
@@ -50,13 +50,12 @@ const (
 	KilledApplication
 	ReleaseAppAllocation
 	ReleaseAppAllocationAsk
-	AppStateChange
 	ResumingApplication
 	AppTaskCompleted
 )
 
 func (ae ApplicationEventType) String() string {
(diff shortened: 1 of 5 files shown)


pkg/cache/application_state.go
```diff
@@ -50,13 +50,12 @@ const (
 	KilledApplication
 	ReleaseAppAllocation
 	ReleaseAppAllocationAsk
-	AppStateChange
 	ResumingApplication
 	AppTaskCompleted
 )
 
 func (ae ApplicationEventType) String() string {
-	return [...]string{"SubmitApplication", "RecoverApplication", "AcceptApplication", "TryReserve", "UpdateReservation", "RunApplication", "RejectApplication", "CompleteApplication", "FailApplication", "KillApplication", "KilledApplication", "ReleaseAppAllocation", "ReleaseAppAllocationAsk", "AppStateChange", "ResumingApplication", "AppTaskCompleted"}[ae]
```
pbacsko, line 59: Note: "RecoverApplication" had to be removed, the string output was actually incorrect.

chia7712: Could you please add unit test to keep consistency between enum and string?

chia7712 (approved):
LGTM

pkg/cache/application_state.go
pbacsko: Yes, I'll update the PR.

PR #835, opened 2024-05-10
[YUNIKORN-2323] Gang scheduling user experience issues

manirajv06:
### What is this PR for?

Send event for every ph allocation and also final event to convey "gang reseravtions has been done" to driver pod.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2323

### Questions:

### DIFF
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -519,7 +519,7 @@ func (app *Application) onReserving() {
 // If we have all the required placeholders progress the application status, otherwise nothing happens
 func (app *Application) onReservationStateChange() {
 	if app.originatingTask != nil {
-		events.GetRecorder().Eventf(app.originatingTask.GetTaskPod().DeepCopy(), nil, v1.EventTypeNormal, "Placeholder Allocated",
+		events.GetRecorder().Eventf(app.originatingTask.GetTaskPod().DeepCopy(), nil, v1.EventTypeNormal, "GangScheduling",
 			"Placeholder Allocated", "Application %s placeholder has been allocated.", app.applicationID)
 	}
 	desireCounts := make(map[string]int32, len(app.taskGroups))
@@ -549,7 +549,7 @@ func (app *Application) onReservationStateChange() {
 
 	if app.originatingTask != nil {
 		// Now that all placeholders has been allocated, send a final conclusion message
-		events.GetRecorder().Eventf(app.originatingTask.GetTaskPod().DeepCopy(), nil, v1.EventTypeNormal, "Gang reservations completed. All placeholders are allocated.",
+		events.GetRecorder().Eventf(app.originatingTask.GetTaskPod().DeepCopy(), nil, v1.EventTypeNormal, "GangScheduling",
 			"Gang reservations completed. All placeholders are allocated.", "Application %s all placeholders are allocated. Transitioning to running state.", app.applicationID)
 	}
 	dispatcher.Dispatch(NewRunApplicationEvent(app.applicationID))
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -518,6 +518,10 @@ func (app *Application) onReserving() {
 // onReservationStateChange is called when there is an add or a release of a placeholder
 // If we have all the required placeholders progress the application status, otherwise nothing happens
 func (app *Application) onReservationStateChange() {
+	if app.originatingTask != nil {
+		events.GetRecorder().Eventf(app.originatingTask.GetTaskPod().DeepCopy(), nil, v1.EventTypeNormal, "Placeholder Allocated",
+			"Placeholder Allocated", "Application %s placeholder has been allocated.", app.applicationID)
+	}
 	desireCounts := make(map[string]int32, len(app.taskGroups))
 	for _, tg := range app.taskGroups {
 		desireCounts[tg.Name] = tg.MinMember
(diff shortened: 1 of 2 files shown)


pkg/cache/application.go
```diff
@@ -518,6 +518,10 @@ func (app *Application) onReserving() {
 // onReservationStateChange is called when there is an add or a release of a placeholder
 // If we have all the required placeholders progress the application status, otherwise nothing happens
 func (app *Application) onReservationStateChange() {
+	if app.originatingTask != nil {
+		events.GetRecorder().Eventf(app.originatingTask.GetTaskPod().DeepCopy(), nil, v1.EventTypeNormal, "Placeholder Allocated",
+			"Placeholder Allocated", "Application %s placeholder has been allocated.", app.applicationID)
```
pbacsko, line 523: To enhance supportability, it's best to include the placeholder name. This requires changes in the `UpdateApplicationReservationEvent` type because right now it does not have the taskID. Once it's there, it can be extracted and passed to this method.

pkg/cache/application_test.go
```diff
@@ -1282,11 +1288,33 @@ func TestApplication_onReservationStateChange(t *testing.T) {
...
+					counter++
+					if counter == 4 {
+						return true
+					}
+				}
+			default:
+				return false
+			}
+		}
+	}, 5*time.Millisecond, 20*time.Millisecond)
```
pbacsko, line 1316: 20ms seems to be a very low value, this can be a problem on slower machines. Use a safer value like `time.Second`.

pkg/cache/application.go
```diff
@@ -518,6 +518,10 @@ func (app *Application) onReserving() {
 // onReservationStateChange is called when there is an add or a release of a placeholder
 // If we have all the required placeholders progress the application status, otherwise nothing happens
 func (app *Application) onReservationStateChange() {
+	if app.originatingTask != nil {
+		events.GetRecorder().Eventf(app.originatingTask.GetTaskPod().DeepCopy(), nil, v1.EventTypeNormal, "Placeholder Allocated",
```
pbacsko, line 522: For the action and reason, just use the text "GangScheduling", we already do it in Task.

pbacsko (changes requested):
Found some minor things.

pkg/cache/application.go
pbacsko: Update: after our discussion offline, it's a better not to include the taskID so that the K8s event system can use a counter to collapse the messages. But we can still include the task group name in the message, so we'll have a per task group counter.

manirajv06: Ok. Even to include task group name, we will need to make changes in many places passes through couple to app state transitions to pass task details. I am not really sure changes worth the benefit. Can we do when the time comes in?

pbacsko: If it's too much hassle, I'm ok with skipping it for the time being.

manirajv06: BTW, How does message collapse happens? Is it going to be based on action or reason or note? Because you asked me to use "GangScheduling" for action and reason later.

pbacsko: The action/reason is usually a short string, a single word. So that's why I used "GangScheduling" here. I assume the collapsing is based on the "message" itself.

6581be2c3cb45b721d99b5d0_3Hza2qBOHqaGN1ZSninHyGL3NIsLEKeMWmJT5z4N4NC-6RzUN2SNoXT5fd64oTFxlbKNZ2HasfPIYQlYBS_CReWca0yaoD9Zw8xQtIVWIYElNN1Q9cRhbKadAi2xUNU1Yxih2xrPedHg9PlOAH98Wcc

pbacsko: The image above does not contain the "count" field. Example:

https://stackoverflow.com/questions/40132277/can-kubectl-describe-show-timestamp-of-pod-events

```
$ kubectl get events -o custom-columns=FirstSeen:.firstTimestamp,LastSeen:.lastTimestamp,Count:.count,From:.source.component,Type:.type,Reason:.reason,Message:.message \
                     --field-selector involvedObject.kind=Pod,involvedObject.name=my-pod
FirstSeen              LastSeen               Count   From                Type      Reason             Message
2020-01-21T16:30:25Z   2020-01-21T16:30:25Z   1       default-scheduler   Normal    Scheduled          Successfully assigned my-pod to 10.10.1.3
2020-01-21T16:30:26Z   2020-01-21T16:30:26Z   1       kubelet             Normal    Pulling            Pulling image "my-image"
2020-01-21T16:30:26Z   2020-01-21T16:30:26Z   1       kubelet             Normal    Pulled             Successfully pulled image "my-image"
2020-01-21T16:30:26Z   2020-01-21T16:30:26Z   1       kubelet             Normal    Created            Created container my-container
2020-01-21T16:30:27Z   2020-01-21T16:30:27Z   1       kubelet             Normal    Started            Started container my-container
```

manirajv06: It seems "reason" would be displayed as is and "action" would be displayed as "message" in events list. I believe collapse should be based on action or message.

Please refer https://dev.to/lucassha/writing-events-to-pod-objects-using-client-go-1p1h.

I think we should use "GangScheduling" as reason and "Placeholder Allocation" as action. We need to follow the pattern in other places too. May need to change in task.go.

pbacsko (approved):
+1 lgtm

PR #837, opened 2024-05-10
[YUNIKORN-2578] Refactor SchedulerCache.GetPod() remove bool return

ryankert01:
### What is this PR for?
SchedulerCache GetPod() and GetPodNoLock() retrun two values:

1. *v1.Pod
2. bool  
The boolean value is redundant as it is false if the pod is not found and a nil is returned for the pod. The boolean is true if the pod has a value. Testing for a nil pod has the same result.

We do not cache a nil pod in the cache for a pod UID

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [x] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2578

make test 
github ci

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -644,8 +644,8 @@ func (ctx *Context) EventsToRegister(queueingHintFn framework.QueueingHintFn) []
 func (ctx *Context) IsPodFitNode(name, node string, allocate bool) error {
 	ctx.lock.RLock()
 	defer ctx.lock.RUnlock()
-	var pod *v1.Pod
-	if pod = ctx.schedulerCache.GetPod(name); pod == nil {
+	pod := ctx.schedulerCache.GetPod(name)
+	if pod == nil {
 		return ErrorPodNotFound
 	}
 	// if pod exists in cache, try to run predicates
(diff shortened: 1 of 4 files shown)


pbacsko (approved):
LGTM

pkg/cache/context.go
```diff
@@ -645,8 +645,7 @@ func (ctx *Context) IsPodFitNode(name, node string, allocate bool) error {
 	ctx.lock.RLock()
 	defer ctx.lock.RUnlock()
 	var pod *v1.Pod
-	var ok bool
-	if pod, ok = ctx.schedulerCache.GetPod(name); !ok {
```
chia7712, line 649:
```go
	pod := ctx.schedulerCache.GetPod(name)
	if pod == nil {
		return ErrorPodNotFound
	}
```

chia7712 (commented):
@ryankert01 thanks for this refactor. one small suggestion is left. PTAL

pkg/cache/context.go
ryankert01: Done

PR #838, opened 2024-05-10
[YUNIKORN-2621] Start the CSINode informer

jacobsalway:
### What is this PR for?

Start and wait for the CSINode informer to sync so that scheduler plugins and the volume binder are able to list and get CSINode objects.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-2621

### Questions:

### DIFF
--- a/pkg/client/apifactory.go
+++ b/pkg/client/apifactory.go
@@ -125,6 +125,7 @@ func NewAPIFactory(scheduler api.SchedulerAPI, informerFactory informers.SharedI
 			PVCInformer:           pvcInformer,
 			NamespaceInformer:     namespaceInformer,
 			StorageInformer:       storageInformer,
+			CSINodeInformer:       csiNodeInformer,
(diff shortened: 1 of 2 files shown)


pbacsko:
Thanks for the patch @jacobsalway. I'm wondering if it makes sense to enhance test coverage for this. We could implement a mock for `cache.SharedIndexInformer` because all informer returns this type when you call `Informer()`. Then we can validate if `HasSynced()` and `Run()` has been called, so `client.Clients` becomes testable.

wilfred-s (approved):
LGTM, leaving the commit to Craig or Peter as they might have more remarks.

pbacsko:
Created https://issues.apache.org/jira/browse/YUNIKORN-2623.

pbacsko (approved):
+1 will merge soon

PR #839, opened 2024-05-13
[YUNIKORN-2623] Create unit tests for Clients

pbacsko:
### What is this PR for?
Create mock for `SharedIndexedInformer` and update existing types.
Write new unit tests for `client.Clients`.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2623

### Questions:

### DIFF
--- a/pkg/client/apifactory_mock.go
+++ b/pkg/client/apifactory_mock.go
@@ -484,7 +484,7 @@ func NewMockedCSINodeInformer() *MockedCSINodeInformer {
 }
 
 func (m *MockedCSINodeInformer) Informer() cache.SharedIndexInformer {
-	return &test.SharedInformerMock{}
+	return m.informer
 }
 
 func (m *MockedCSINodeInformer) Lister() storagev1.CSINodeLister {
--- a/pkg/client/apifactory_mock.go
+++ b/pkg/client/apifactory_mock.go
@@ -85,12 +85,13 @@ func NewMockedAPIProvider(showError bool) *MockedAPIProvider {
 			PodInformer:           test.NewMockedPodInformer(),
 			NodeInformer:          test.NewMockedNodeInformer(),
 			ConfigMapInformer:     test.NewMockedConfigMapInformer(),
-			PVInformer:            &MockedPersistentVolumeInformer{},
-			PVCInformer:           &MockedPersistentVolumeClaimInformer{},
-			StorageInformer:       &MockedStorageClassInformer{},
+			PVInformer:            NewMockedPersistentVolumeInformer(),
+			PVCInformer:           NewMockedPersistentVolumeClaimInformer(),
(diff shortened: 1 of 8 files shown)


pkg/client/clients_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"github.com/apache/yunikorn-core/pkg/common"
+	"github.com/apache/yunikorn-k8shim/pkg/common/test"
+)
+
+const (
+	noOfInformers = 9
+)
+
+func TestWaitForSync(t *testing.T) {
```
chia7712, line 35: Maybe we can do a bit refactor to `Clients` to simplify the test. 

1. add a new method `informers` to  `Clients` to return all informers
2. rewrite `WaitForSync` and `Run` by looping all informers
3. in this test we can create a mock clients to override the new method `informers` to return a single mock informer

WDYT?

pbacsko: Could we do it a separate ticket? In this one I'd rather just focus on the new tests.

chia7712: ok, I have filed https://issues.apache.org/jira/browse/YUNIKORN-2625

pkg/client/apifactory_mock.go
```diff
@@ -416,38 +417,80 @@ func (m *MockedAPIProvider) GetBoundPods(clear bool) []BoundPod {
...
+}
+
+func NewMockedCSINodeInformer() *MockedCSINodeInformer {
+	return &MockedCSINodeInformer{
+		informer: &test.SharedInformerMock{},
+	}
+}
+
+func (m *MockedCSINodeInformer) Informer() cache.SharedIndexInformer {
+	return &test.SharedInformerMock{}
```
chia7712, line 487: maybe it should return `m.informer`?

pkg/client/clients_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"time"
+
+	"gotest.tools/v3/assert"
+
+	"github.com/apache/yunikorn-core/pkg/common"
+	"github.com/apache/yunikorn-k8shim/pkg/common/test"
+)
+
+const (
+	noOfInformers = 9
```
chia7712, line 32: Could you add comments for this magic number?

chia7712 (commented):
@pbacsko thanks for this patch!

pkg/client/apifactory_mock.go
pbacsko: Correct, fixed.

chia7712 (approved):
LGTM

PR #842, opened 2024-05-17
[YUNIKORN-2630] Release context lock early for config changes

wilfred-s:
### What is this PR for?
Release the lock of the context in the shim when processing is done. When the config changes are sent to the core the k8shim should not be locked. The context changes have been finalised at that point.

The core handles its own locking and serialises config changes that come in from the k8shim.

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-2630

Current unit and e2e tests cover the changed paths

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -603,7 +603,7 @@ func (ctx *Context) deletePriorityClass(obj interface{}) {
 func (ctx *Context) triggerReloadConfig(index int, configMap *v1.ConfigMap) {
 	// hot reload is turned off do nothing
 	// hot reload can be turned off by an update: safety first access under lock to prevent data race
-	if !ctx.apiProvider.GetAPIs().GetConf().IsConfigReloadable() {
+	if !schedulerconf.GetSchedulerConf().IsConfigReloadable() {
 		log.Log(log.ShimContext).Info("hot-refresh disabled, skipping scheduler configuration update")
 		return
 	}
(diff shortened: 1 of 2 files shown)


pbacsko (approved):
+1

pkg/cache/context.go
```diff
@@ -601,25 +601,18 @@ func (ctx *Context) deletePriorityClass(obj interface{}) {
...
 
 func (ctx *Context) triggerReloadConfig(index int, configMap *v1.ConfigMap) {
-	ctx.lock.Lock()
-	defer ctx.lock.Unlock()
-
-	conf := ctx.apiProvider.GetAPIs().GetConf()
-	if !conf.EnableConfigHotRefresh {
+	// hot reload is turned off do nothing
+	// hot reload can be turned off by an update: safety first access under lock to prevent data race
+	if !ctx.apiProvider.GetAPIs().GetConf().IsConfigReloadable() {
```
chia7712, line 606: This is unrelated to this PR, but should we replace this long call chain by`conf.GetSchedulerConf().IsConfigReloadable()`? The `SchedulerConf` is a global singleton, and it would be nice that we unify the usage of `ctx.apiProvider.GetAPIs().GetConf()` and `conf.GetSchedulerConf()`

chia7712 (approved):
@wilfred-s nice patch.

chenyulin0719 (approved):
+1 LGTM

pkg/cache/context.go
wilfred-s: removed the call and use scheduler conf directly

chia7712 (approved):

PR #846, opened 2024-05-21
[YUNIKORN-2618]Update Streamline AsyncRMCallback UpdateAllocation

SophieTech88:
### What is this PR for?
if task is not found, a nil is returned from context.getTask in  for response.New processing we should just log that fact and proceed to the next alloc. Simplifies the flow as we never need to check for a. nil task. We should never have a pod in the cache that does not exist as a task on an application.

We retrieve the application using the application ID from the response to never use the object. We only use the application ID to pass into an event. The context event handler then does the exact same lookup again to process the event on the app.

We need to become much smarter in this area, double or triple lookups, generate async events that just change the state of the app or task or kick off another event.

Actions:
* Check the task, if task == nil, make a logging and skip that loop and jump to next iteration.
* Retrieve the application using the application ID from the response

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2618

It should pass all CI tests.
I tested locally. It passed all CI tests and e2e test.

### Questions:

### DIFF
--- a/pkg/cache/scheduler_callback.go
+++ b/pkg/cache/scheduler_callback.go
@@ -56,27 +56,24 @@ func (callback *AsyncRMCallback) UpdateAllocation(response *si.AllocationRespons
 
 		// update cache
 		task := callback.context.getTask(alloc.ApplicationID, alloc.AllocationKey)
-		if task != nil {
-			task.setAllocationKey(alloc.AllocationKey)
-		} else {
+		if task == nil {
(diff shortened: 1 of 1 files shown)


pbacsko (approved):
+1 pending e2e

chia7712 (approved):

PR #847, opened 2024-05-23
[YUNIKORN-2637] finalizePods should ignore pods like registerPods

wilfred-s:
### What is this PR for?
If a pod was in a terminal state during registration it is skipped. The same principal should apply to finalisePods:
* only check pods that were added in registerPods.
* remove finished pods in the finalizePods call if they were registered.
* new pods are not added in finalizePods

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-2637

Unit tests now cover the code path
e2e tests coverage is not possible as we cannot time the pod removal well enough

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -1690,7 +1690,7 @@ func (ctx *Context) finalizeNodes(existingNodes []*v1.Node) error {
 }
 
 func (ctx *Context) registerPods() ([]*v1.Pod, error) {
-	log.Log(log.ShimContext).Info("Starting node registration...")
+	log.Log(log.ShimContext).Info("Starting pod registration...")
 
 	// list all pods via the informer
 	pods, err := ctx.apiProvider.GetAPIs().PodInformer.Lister().List(labels.Everything())
@@ -1705,14 +1705,16 @@ func (ctx *Context) registerPods() ([]*v1.Pod, error) {
 	})
 
 	// add all pods to the context
-	for _, pod := range pods {
-		// skip terminated pods
+	for i, pod := range pods {
+		// skip terminated pods: we do not add or finalise them later
 		if utils.IsPodTerminated(pod) {
+			pods[i] = nil
 			continue
 		}
 		ctx.AddPod(pod)
 	}
 
+	log.Log(log.ShimContext).Info("Finished pod registration...")
 	return pods, nil
 }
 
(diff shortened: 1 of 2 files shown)


pbacsko:
Unit test failure is due to YUNIKORN-2629.

pbacsko (approved):
+1 pending e2e

A related question: Context doesn't have the best unit test coverage. I do believe we need to enhance it. Eg `bindPodVolumes()` is completely uncovered, but there are a lot of missing paths, too. Thoughts?

wilfred-s:
> +1 pending e2e
> 
> A related question: Context doesn't have the best unit test coverage. I do believe we need to enhance it. Eg `bindPodVolumes()` is completely uncovered, but there are a lot of missing paths, too. Thoughts?

I would completely agree. That is why I wrote tests that cover the two functions that I changed.

wilfred-s:
I think we need to get a fix for YUNIKORN-2629 created, reviewed and committed to get the unit tests to pass on the k8shim and get things stable again. Until that point I think we need to wait with commits in the master.

pkg/cache/context.go
```diff
@@ -1727,13 +1729,21 @@ func (ctx *Context) finalizePods(existingPods []*v1.Pod) error {
...
-	// find any existing nodes that no longer exist
+	// find any existing pods that no longer exist
 	for _, pod := range existingPods {
+		// skip if the pod was already terminated during register
+		if pod == nil {
+			continue
+		}
 		if _, ok := podMap[pod.UID]; !ok {
-			// node no longer exists, delete it
+			// pod no longer exists, delete it
```
chia7712, line 1746: Just curious. In this phase we have added event handler, so those pods may be removed already from `context` already. It seems to me that is a race condition but it is fine as `ctx.DeletePod(pod)` is no-op for repeated delete. If this description is valid, maybe we should add comments for it. Also, we should add tests for `DeletePod` to make sure it is no-op in repeated deletes.

pbacsko: I'm planning to enhance test coverage for `Context`, we can have a separate test case for `DeletePod()` to make sure that's idempotent. But we can do this in this PR as well.

chia7712: +1 to address that in separate PR :)

wilfred-s: >  so those pods may be removed already

The _maybe_ is what we're trying to catch. It could be that the pod was removed between the list in registerPods and the event handler add later on. It will not show here but the event handler has not seen it being removed. Those we want to catch. Not the ones that have already been deleted by the handlers. We should not have to describe that in this function.

> Also, we should add tests for DeletePod to make sure it is no-op in repeated deletes.

See above for the comment made by Peter around the unit test coverage of the context code. It needs to be extended. Any code path or function that declares to be idempotent should have tests for it.

pbacsko (approved):
+1
All green, I'm merging this.

PR #849, opened 2024-05-23
[YUNIKORN-2629] Adding a node can result in a deadlock

pbacsko:
### What is this PR for?
Prevent deadlock in `registerNodes()` by releasing/re-acquiring the write lock.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2629

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -1417,7 +1417,7 @@ func (ctx *Context) InitializeState() error {
 		log.Log(log.ShimContext).Error("failed to load nodes", zap.Error(err))
 		return err
 	}
-	acceptedNodes, err := ctx.registerNodes(nodes)
+	acceptedNodes, err := ctx.RegisterNodes(nodes)
 	if err != nil {
 		log.Log(log.ShimContext).Error("failed to register nodes", zap.Error(err))
 		return err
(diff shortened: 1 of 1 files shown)


chenyulin0719:
Hi @pbacsko,

The temporary release context lock method makes sense to me. 

However, I guess the root cause of the deadlock is other tests.
When another test called 'dispatcher.UnregisterAllEventHandlers()', 

- https://github.com/apache/yunikorn-k8shim/blob/08a9b93dcc308f1e70bbf9a6596664877977e1bd/pkg/cache/scheduler_callback_test.go#L452-L455

the EventHandler will be cleaned and the wait group will never meet the count. 
- https://github.com/apache/yunikorn-k8shim/blob/08a9b93dcc308f1e70bbf9a6596664877977e1bd/pkg/cache/context.go#L1574

So the context lock is not released.

wilfred-s:
I am OK with the change as it is for 1.5.2. We need to have a proper look at the context lock for the 1.6.0 release and we should try to prevent this change from becoming the final solution.

pbacsko:
> I am OK with the change as it is for 1.5.2. We need to have a proper look at the context lock for the 1.6.0 release and we should try to prevent this change from becoming the final solution.

OK. Putting it back to draft. I'll commit it directly to branch-1.5.

pbacsko:
This is solved by #859, closing.

PR #852, opened 2024-05-31
[YUNIKORN-2640] Remove config from Clients

0lai0:
### What is this PR for?

The config (`conf.SchedulerConf`) [0] references to a global singleton object [1][2]. Also, in the code base `clients#GetConf()` is used 3 times [3] and `conf.GetSchedulerConf()` is used 61 times.
So this PR removes config from Clients.

### What type of PR is it?
* [x] - Improvement

### Todos
N/A

### What is the Jira issue?
[[YUNIKORN-2640]](https://issues.apache.org/jira/browse/YUNIKORN-2640)

It has been checked locally by make test.
N/A
### Questions:
N/A

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -46,7 +46,6 @@ import (
 	"github.com/apache/yunikorn-k8shim/pkg/common/constants"
 	"github.com/apache/yunikorn-k8shim/pkg/common/events"
 	"github.com/apache/yunikorn-k8shim/pkg/common/utils"
-	"github.com/apache/yunikorn-k8shim/pkg/conf"
 	schedulerconf "github.com/apache/yunikorn-k8shim/pkg/conf"
 	"github.com/apache/yunikorn-k8shim/pkg/dispatcher"
 	"github.com/apache/yunikorn-k8shim/pkg/locking"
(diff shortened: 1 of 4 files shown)


pkg/cache/context.go
```diff
@@ -93,7 +94,7 @@ func NewContextWithBootstrapConfigMaps(apis client.APIProvider, bootstrapConfigM
 	ctx := &Context{
 		applications: make(map[string]*Application),
 		apiProvider:  apis,
-		namespace:    apis.GetAPIs().GetConf().Namespace,
+		namespace:    conf.GetSchedulerConf().Namespace,
```
chia7712, line 97: It should be `schedulerconf.GetSchedulerConf()`

pkg/cache/context.go
```diff
@@ -46,6 +46,7 @@ import (
 	"github.com/apache/yunikorn-k8shim/pkg/common/constants"
 	"github.com/apache/yunikorn-k8shim/pkg/common/events"
 	"github.com/apache/yunikorn-k8shim/pkg/common/utils"
+	"github.com/apache/yunikorn-k8shim/pkg/conf"
```
chia7712, line 49: We don't need to re-import it.

chia7712 (commented):
@0lai0 thanks for this cleanup!

pkg/cache/context.go
0lai0: No problem, I've completed the modification.
PTAL

chia7712 (approved):
LGTM

PR #855, opened 2024-06-06
[YUNIKORN-2665] Gang app originator pod changes after restart

manirajv06:
### What is this PR for?
  
During recovery, ensure gang app originator remains same. Ordering of pods shouldn't have any impact on originator selection.

### What type of PR is it?
* [ ] - Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2665

unit test added

### Questions:

### DIFF
--- a/pkg/cache/context_test.go
+++ b/pkg/cache/context_test.go
@@ -2308,6 +2308,8 @@ func TestOriginatorPodAfterRestart(t *testing.T) {
 
 	app := context.getApplication("spark-app-01")
 	assert.Equal(t, app.originatingTask.taskID, "UID-00001")
+	assert.Equal(t, len(app.GetPlaceHolderTasks()), 2)
+	assert.Equal(t, len(app.GetAllocatedTasks()), 0)
 }
 
 func initAssumePodTest(binder *test.VolumeBinderMock) *Context {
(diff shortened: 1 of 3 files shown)


pkg/cache/context_test.go
```diff
@@ -2224,6 +2224,92 @@ func TestAssumePod_PodNotFound(t *testing.T) {
...
+		Spec: v1.PodSpec{SchedulerName: "yunikorn"},
+	}
+
+	// Add the ph pods first and then real driver pod at the last
+	context.AddPod(pod3)
+	context.AddPod(pod2)
+	context.AddPod(pod1)
+
+	app := context.getApplication("spark-app-01")
+	assert.Equal(t, app.originatingTask.taskID, "UID-00001")
```
pbacsko, line 2310: I believe it's best to see the core side as well. Eg. application object, its state, list of asks, etc.

manirajv06: We have simulated the scenario of processing pods in random order (after restart) directly unlike going through the full cycle as stated below:

creating the app, waiting for placeholder creation, YK restarts, processing pods in random order etc.

So, we can validate certain things only but not state etc because it won't be correct.

wilfred-s (approved):
LGTM

pkg/cache/context_test.go
pbacsko: OK I misunderstood. There's no real core here, I thought this test used a `MockScheduler`. So just ignore my comment.

pbacsko (approved):

pbacsko:
LGTM

PR #858, opened 2024-06-20
[YUNIKORN-2685] Use the newer WaitForCondition() in shim test

mean-world:
### What is this PR for?
shim completly use WaitForCondition function  from yunikorn-core
also remove E2E test about Dao.allocation.partition based on  [YUNIKORN-2593](https://github.com/apache/yunikorn-core/pull/858)

### What type of PR is it?
* [x] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2685

### Questions:


pbacsko:
cc @wilfred-s @craigcondit do we ever import stuff from the core directly? We do it for the deadlock detection, but apart from that, I haven't seen this practice anywhere.

chenyulin0719:
@mean-world Sorry I wasn't aware that there is a similar function in shim's utils.

For me, I would vote to align the implementation of WaitForCondition(), but keep two separate function in shim and core.
I don't think code reuse is necessary here.

@pbacsko, @chia7712 WDYT?

chia7712:
>  do we ever import stuff from the core directly? 

there are use case already. 

https://github.com/apache/yunikorn-k8shim/blob/24efbeda6800fabec17cf9e0474cebee0314bd6e/pkg/cache/task_test.go#L203
https://github.com/apache/yunikorn-k8shim/blob/24efbeda6800fabec17cf9e0474cebee0314bd6e/pkg/client/clients_test.go#L61
https://github.com/apache/yunikorn-k8shim/blob/24efbeda6800fabec17cf9e0474cebee0314bd6e/pkg/client/clients_test.go#L68

The core and k8shim are different repo, so it will be a issue in aligning the version, but reusing the utils methods is always good to me.

craigcondit:
We shouldn't share code between core and shim. The scheduler interface is meant to be the common point. There are exceptions for some low-level functionality like locks and logging, but that's out of necessity. Tightly coupling the two projects just creates friction and makes refactoring more difficult.

chia7712:
> The other examples just given should probably be fixed to not use core code.

that sounds good to me. Maybe we can have another jira to cleanup the usage of core code.

pbacsko:
We're on the same page. This PR can be closed and a new ticket should eliminate the existiting core imports where it's possible.

@mean-world do you create a new JIRA and close this one?

mean-world:
> We're on the same page. This PR can be closed and a new ticket should eliminate the existiting core imports where it's possible.
> 
> @mean-world please create a new JIRA and close this one.

I already create new JIRA
https://issues.apache.org/jira/browse/YUNIKORN-2690

PR #859, opened 2024-06-21
[YUNIKORN-2629] Adding a node can result in a deadlock

pbacsko:
### What is this PR for?
Fix deadlock problem by modifying locking in `cache.Context`.

Lock/unlock calls were removed where it doesn't seem to be necessary. In most cases, the state of the context is not modified at all. Only the scheduler cache is affected which has its own lock.

Testing done:
* `make test` multiple times
* `BenchmarkSchedulingThroughPut` with deadlock detector
* `BenchmarkSchedulingThroughPut` with race detector

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2629

### Questions:


pkg/cache/context.go
```diff
@@ -1053,25 +1014,28 @@ func (ctx *Context) getApplication(appID string) *Application {
 
 func (ctx *Context) RemoveApplication(appID string) error {
```
chia7712, line 1015: This method is used by testing only. Could we use `RemoveApplicationInternal` to replace it? It seems we can make `RemoveApplicationInternal` return `boolean` to indicate successful remove for testing.

chia7712 (commented):
@pbacsko thanks for this patch

chenyulin0719:
Based on my understanding. 
After this PR. One of the results is that the informer event handlers are no longer grab context lock when updating nodes or updating Pods. (NodeInformer/PodInformer go routine)

I'm wondering if there have any racing when informer events(Pods/Nodes) comes during below go routine running:

1. (main goroutine) InitializeState() -> I'm not quite confident.
2. (shim scheduler goroutine) schedule() -> Should be safe.
3. (dispatcher goroutine) -> Should be safe.

For the InitializeState(), I'm not quite confident for answering below question?
Q: What will happen if NodeInformer update NodeA when InitializeState() is registring NodeA?
Q: What will happen if PodInformer update PodA when InitializeState() is registring PodA?

Is it still be safe?

pkg/cache/context.go
wilfred-s: It needs a cleanup, might be better to do that via a seperate jira. That makes it clear we're cleaning up dead code.  We should keep the RemoveApplication name and remove the RemoveApplicationInternal name. It is not really internal as it gets called from outside the context.

wilfred-s:
> 1. (main goroutine) InitializeState() -> I'm not quite confident.

That should not be a problem.

> For the InitializeState(), I'm not quite confident to answer below questions: Q: What will happen if NodeInformer update NodeA when InitializeState() is registring NodeA? Q: What will happen if PodInformer update PodA when InitializeState() is registring PodA?
> 
> Is it still safe?

The informers are first synced without event handling on. That gives us the base list of pods and nodes. All these listed objects are then processed. The event handler are turned off until we have done all that work. The step 5 in the init turns them on. At that point we no longer have to worry about the init code. The cache lock will prevent the simultaneous changes.

craigcondit (approved):
+1 LGTM.

pkg/cache/context.go
wilfred-s: filed: [YUNIKORN-2782](https://issues.apache.org/jira/browse/YUNIKORN-2782)

wilfred-s (approved):
LGTM +1
The horrible fix needed in `(ctx *Context) RemoveApplication(appID string)` will disappear completely when we fix YUNIKORN-2782 as the function is not used

PR #861, opened 2024-06-24
[YUNIKORN-2658]add nolint:funlen to long functions to supress the lint warnings 

rich7420:
### What is this PR for?

To add nolint:funlen to long functions to supress the lint warnings, when we use `make lint` 
like this 
image

### What type of PR is it?

- [x] - Improvement

### Todos
-  Improve other warn message

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2658

in [Makefile](https://github.com/apache/yunikorn-k8shim/blob/master/Makefile#L294)  take off `--new-from-rev=$${headSHA}`
then run `make lint
`
NA
### Questions:

### DIFF
--- a/pkg/cache/application_test.go
+++ b/pkg/cache/application_test.go
@@ -559,6 +559,7 @@ func TestGetNonTerminatedTaskAlias(t *testing.T) {
 	task1 := NewTask(taskID1, app, context, pod1)
 	app.taskMap[taskID1] = task1
 	task1.sm.SetState(TaskStates().Pending)
+	//nolint:funlen
 	taskID2 := "task02"
 	task2 := NewTask(taskID2, app, context, pod2)
 	app.taskMap[taskID2] = task2
--- a/pkg/cache/application_test.go
+++ b/pkg/cache/application_test.go
@@ -750,6 +750,7 @@ func TestTryReserve(t *testing.T) {
 	assert.NilError(t, err, "placeholders are not created")
 }
 
+//nolint:funlen
 func TestTryReservePostRestart(t *testing.T) {
 	context := initContextForTest()
 	dispatcher.RegisterEventHandler("TestAppHandler", dispatcher.EventTypeApp, context.ApplicationEventHandler())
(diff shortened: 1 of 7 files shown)


pkg/cache/application_test.go
```diff
@@ -559,6 +559,7 @@ func TestGetNonTerminatedTaskAlias(t *testing.T) {
 	task1 := NewTask(taskID1, app, context, pod1)
 	app.taskMap[taskID1] = task1
 	task1.sm.SetState(TaskStates().Pending)
+	//nolint:funlen
```
chia7712, line 562: Could you please move this suppression to line#527?

chia7712 (commented):
@rich7420 thanks for patch. Could you please fix following warnings also?
```
pkg/cache/application_test.go:753: Function 'TestTryReservePostRestart' is too long (123 > 120) (funlen)
func TestTryReservePostRestart(t *testing.T) {
pkg/cache/context_test.go:1485: Function 'TestAddApplicationsWithTags' is too long (145 > 120) (funlen)
func TestAddApplicationsWithTags(t *testing.T) {

```

pkg/cache/application_test.go
rich7420: OK!!

pkg/cache/task_state.go
```diff
@@ -314,6 +314,7 @@ func TaskStates() *TStates {
 	return storeTaskStates
 }
 
+//nolint:funlen
 func newTaskState() *fsm.FSM {
```
chia7712, line 318: this is production code, so maybe we can do a bit refactor? for example, we can add two helper methods to generate `eventDesc` and `callbacks`. With those helpers, `newTaskState` can be simplified to following style
```go
func newTaskState() *fsm.FSM {
	states := TaskStates()
	return fsm.NewFSM(states.New, eventDesc(states), callbacks(states))
}

```

rich7420: Ok! no problem.

chia7712:
@rich7420 please take a look at failed tests?

rich7420:
OK, no problem

chia7712 (approved):
LGTM

PR #864, opened 2024-06-27
[YUNIKORN-2319] cache.Task: reference to old pod object is kept after update

pbacsko:
### What is this PR for?
Whenever a pod is updated, then replace it inside the `Task` as well so that the previous version can be garbage collected.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2319

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -305,20 +305,17 @@ func (ctx *Context) UpdatePod(_, newObj interface{}) {
 }
 
 func (ctx *Context) updateYuniKornPod(appID string, pod *v1.Pod) {
-	if app := ctx.getApplication(appID); app != nil {
-		if task, err := app.GetTask(string(pod.UID)); task != nil && err == nil {
+	var app *Application
+	taskID := string(pod.UID)
+	if app = ctx.getApplication(appID); app != nil {
+		if task, err := app.GetTask(taskID); task != nil && err == nil {
 			task.setTaskPod(pod)
 		}
 	}
 
 	// treat terminated pods like a remove
 	if utils.IsPodTerminated(pod) {
-		if taskMeta, ok := getTaskMetadata(pod); ok {
-			if app := ctx.getApplication(taskMeta.ApplicationID); app != nil {
-				ctx.notifyTaskComplete(taskMeta.ApplicationID, taskMeta.TaskID)
-			}
-		}
-
+		ctx.notifyTaskComplete(appID, taskID)
 		log.Log(log.ShimContext).Debug("Request to update terminated pod, removing from cache", zap.String("podName", pod.Name))
 		ctx.schedulerCache.RemovePod(pod)
 		return
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -202,7 +202,7 @@ func (ctx *Context) updateNodeInternal(node *v1.Node, register bool) {
 			if applicationID == "" {
 				ctx.updateForeignPod(pod)
 			} else {
-				ctx.updateYuniKornPod(pod)
+				ctx.updateYuniKornPod(applicationID, pod)
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


pkg/cache/context.go
```diff
@@ -296,14 +296,21 @@ func (ctx *Context) UpdatePod(_, newObj interface{}) {
...
 		ctx.updateForeignPod(pod)
 	} else {
-		ctx.updateYuniKornPod(pod)
+		ctx.updateYuniKornPod(applicationID, pod)
 	}
 }
 
-func (ctx *Context) updateYuniKornPod(pod *v1.Pod) {
+func (ctx *Context) updateYuniKornPod(appID string, pod *v1.Pod) {
+	if app := ctx.getApplication(appID); app != nil {
```
wilfred-s, line 308: We use the app in more than just this check. Retrieve outside of the an if and keep for reference.

pkg/cache/context.go
```diff
@@ -296,14 +296,21 @@ func (ctx *Context) UpdatePod(_, newObj interface{}) {
...
+		if task, err := app.GetTask(string(pod.UID)); task != nil && err == nil {
+			task.setTaskPod(pod)
+		}
+	}
+
 	// treat terminated pods like a remove
 	if utils.IsPodTerminated(pod) {
 		if taskMeta, ok := getTaskMetadata(pod); ok {
```
wilfred-s, line 316: We already have the app, or know it exists we need to clean up here now.
The `TaskID` in the metadata is the pod UID. The appID is passed into the function already.

wilfred-s: I also think we need to change the signature for `notifyTaskComplete(string, string)` to `notifyTaskComplete(*Application, string)` Probably better to use a separate jira for that as it flows through into `NotifyTaskComplete()` and some tests. The 2 tests have the application pointer already. It removes a number of extra getApplication() calls we really do not need.
Similar for `ensureAppAndTaskCreated()` which is only ever called from this function. Add a parameter to it to make it: `ensureAppAndTaskCreated(*v1.Pod, *Application)` and only execute application creation `if app == nil`. This can be either in this jira or in a separate one.

wilfred-s (changes requested):

pkg/cache/context.go
pbacsko: Created https://issues.apache.org/jira/browse/YUNIKORN-2724

pkg/cache/context.go
pbacsko: Done

pkg/cache/task.go
```diff
@@ -599,3 +601,9 @@ func (task *Task) failWithEvent(errorMessage, actionReason string) {
 	events.GetRecorder().Eventf(task.pod.DeepCopy(),
 		nil, v1.EventTypeWarning, actionReason, actionReason, errorMessage)
 }
+
+func (task *Task) setTaskPod(pod *v1.Pod) {
```
chia7712, line 605: Pardon me, why this PR does not adopt the cache solution from the discussion of https://issues.apache.org/jira/browse/YUNIKORN-2319?

pbacsko: Note sure what you mean? I think you're probably referring to the suggestion of removing the pod reference from Task itself entirely and retrieving from the scheduler cache. That's possible but it's far more work. This is way simpler. Also, as long as the task object exists in memory, I think it's better to keep the pod reference.

chia7712: > I think you're probably referring to the suggestion of removing the pod reference from Task itself entirely and retrieving it from the scheduler cache. That's possible but it's far more work. This is way simpler. Also, as long as the task object exists in memory, I think it's better to keep the pod reference.

Got it and agree to this is simpler

pbacsko:
@wilfred-s could you review this chage?

craigcondit (approved):
+1 LGTM.

PR #873, opened 2024-07-08
[YUNIKORN-2724] Improve the signature of methods notifyTaskComplete() and ensureAppAndTaskCreated()

ryankert01:
### What is this PR for?
From the review https://github.com/apache/yunikorn-k8shim/pull/864

Change notifyTaskComplete(string, string) to notifyTaskComplete(*Application, string). It removes a number of extra getApplication() calls we really do not need.
Similar for ensureAppAndTaskCreated() which is only ever called from this function. Add a parameter to it to make it: ensureAppAndTaskCreated(*v1.Pod, *Application) and only execute application creation if app == nil. 

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2724

```
make test
```
github ci

### Questions:


pbacsko:
@ryankert01 please rebase the PR with the latest master, thx.

pbacsko:
Nevermind, I updated the PR with master, please check if it's OK.

pkg/cache/context.go
```diff
@@ -315,7 +315,9 @@ func (ctx *Context) updateYuniKornPod(appID string, pod *v1.Pod) {
 
 	// treat terminated pods like a remove
 	if utils.IsPodTerminated(pod) {
-		ctx.notifyTaskComplete(appID, taskID)
+		if app != nil {
```
wilfred-s, line 318: Lets clean up a bit more: the line 310 we retrieve the app inside the if and we have a separate declaration of the var in 308. We can fold the two into one by retrieving the app outside the if:
```
app := ctx.getApplication(appID)
if app != nil {
```

pkg/cache/context.go
```diff
@@ -908,25 +910,23 @@ func (ctx *Context) StartPodAllocation(podKey string, nodeID string) bool {
...
-			zap.String("appID", appID),
-			zap.String("taskID", taskID))
-		ev := NewSimpleTaskEvent(appID, taskID, CompleteTask)
-		dispatcher.Dispatch(ev)
-		if app.GetApplicationState() == ApplicationStates().Resuming {
-			dispatcher.Dispatch(NewSimpleApplicationEvent(appID, AppTaskCompleted))
-		}
+	log.Log(log.ShimContext).Debug("release allocation",
+		zap.String("appID", app.applicationID),
+		zap.String("taskID", taskID))
```
wilfred-s, line 925: Two times similar log we only need/want one

wilfred-s (changes requested):
Missing the change for `ensureAppAndTaskCreated` as per the jira

brandboat:
gentle ping @ryankert01, could you please fix the conflicts ? Thanks.

pkg/cache/context.go
ryankert01: Done.

ryankert01:
Just finished the PR, ready for review now!

pkg/cache/context.go
```diff
@@ -295,41 +295,41 @@ func (ctx *Context) UpdatePod(_, newObj interface{}) {
...
+	if app != nil {
 		if task := app.GetTask(taskID); task != nil {
 			task.setTaskPod(pod)
 		}
 	}
 
 	// treat terminated pods like a remove
 	if utils.IsPodTerminated(pod) {
-		ctx.notifyTaskComplete(appID, taskID)
+		if app != nil {
```
brandboat, line 308: Perhaps we could move the condition `if app != nil ` back to `notifyTaskComplete` and simplify `deleteYuniKornPod`
from
```go
                 if app := ctx.GetApplication(taskMeta.ApplicationID); app != nil {
			ctx.notifyTaskComplete(taskMeta.ApplicationID, taskMeta.TaskID)
		}
```
to 
```go
		ctx.notifyTaskComplete(ctx.GetApplication(taskMeta.ApplicationID), taskMeta.TaskID)
```

WDYT ?

brandboat (commented):
Thanks for the patch.

pkg/cache/context.go
```diff
@@ -873,19 +873,23 @@ func (ctx *Context) StartPodAllocation(podKey string, nodeID string) bool {
 	return ctx.schedulerCache.StartPodAllocation(podKey, nodeID)
 }
 
-func (ctx *Context) notifyTaskComplete(appID, taskID string) {
+func (ctx *Context) NotifyTaskComplete(app *Application, taskID string) {
+	ctx.lock.Lock()
+	defer ctx.lock.Unlock()
```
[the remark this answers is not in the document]
ryankert01: Yes, we don't need lock here. Done.

pkg/cache/context.go
ryankert01: I'm with this change. That way, it change the behaviour of `notifyTaskComplete()` from assuming app is not nil to if nil occur it will return.

brandboat (approved):
LGTM, thanks

pkg/cache/context.go
chia7712: I guess `if app != nil {` can be removed now, right?

pkg/cache/context.go
```diff
@@ -873,19 +871,19 @@ func (ctx *Context) StartPodAllocation(podKey string, nodeID string) bool {
...
+		log.Log(log.ShimContext).Debug("In notifyTaskComplete but app is nil",
 			zap.String("taskID", taskID))
-		ev := NewSimpleTaskEvent(appID, taskID, CompleteTask)
-		dispatcher.Dispatch(ev)
-		if app.GetApplicationState() == ApplicationStates().Resuming {
-			dispatcher.Dispatch(NewSimpleApplicationEvent(appID, AppTaskCompleted))
-		}
+		return
+	}
+	log.Log(log.ShimContext).Debug("notifyTaskComplete and release allocation",
```
chia7712, line 880: Maybe we can keep origin message? `release allocation`

ryankert01: I change the message to "notifyTaskComplete and release allocation" because I combine 2 logs to make it only log once. Perhaps I can change it to " release allocation in notifyTaskComplete"?

pkg/cache/context.go
ryankert01: You're right! Done.

pkg/cache/context.go
chia7712: Got it. It is fine to keep current message :)

chia7712 (approved):
LGTM

PR #874, opened 2024-07-09
[YUNIKORN-2711] Skip setting the queue name to default queue in the shim

mitdesai:
### What is this PR for?
This change removes setting the default queue name from the admission controller. Instead default queue will be decided at the time of placement. (YUNIKORN-2703)

### What type of PR is it?
* [X] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2711

### Questions:
* [X] - It needs documentation.


mitdesai:
Related core changes to try and place the application in root.default are here: https://github.com/apache/yunikorn-core/pull/904

This change does take away the configuration for setting a custom default queue where the applications will be placed if no rule matches. This can be achieved by adding a fixed rule at the end of the placement rules with the desired queue name.

for example, if the placement rules are:
```
placementrules:
- provided
- tag
```
And the application placement cannot succeed from both these rules, we will try to place the application to place in 'root.default' before rejecting it.

If any other queue is desired as a default queue, we can update the placement rules to:
```
placementrules:
- provided
- tag
- fixed
  value: root.custom
```
Here, after provided and tag are unsuccessful, the app will be placed by the fixed rule as long as the acls are valid.

craigcondit (approved):
+1 LGTM.

PR #879, opened 2024-07-17
[YUNIKORN-2762] Improve util funtion's test coverage

SP12893678:
### What is this PR for?
Improve the following function unit test in util.go
-  IsPluginMode
- Convert2ConfigMap
- IsPodRunning
- GetNamespaceQuotaFromAnnotation (JSON Unmarshal error case)
- WaitForCondition
- GetCoreSchedulerConfigFromConfigMap (NotMapping file case)

### What type of PR is it?
* [x] - Test

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2762

### DIFF
--- a/pkg/common/utils/utils.go
+++ b/pkg/common/utils/utils.go
@@ -20,6 +20,7 @@ package utils
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"reflect"
 	"strconv"
@@ -44,6 +45,10 @@ const userInfoKey = siCommon.DomainYuniKorn + "user.info"
 const uniqueAutogenSuffix = "-uniqueautogen"
 
 var pluginMode bool
+var (
+	// ErrorTimeout returned if waiting for a condition times out
+	ErrorTimeout = errors.New("timeout waiting for condition")
+)
 
 func SetPluginMode(value bool) {
 	pluginMode = value
@@ -250,7 +255,7 @@ func WaitForCondition(eval func() bool, interval time.Duration, timeout time.Dur
 		}
 
 		if time.Now().After(deadline) {
-			return fmt.Errorf("timeout waiting for condition")
+			return ErrorTimeout
 		}
 
 		time.Sleep(interval)
--- a/pkg/common/utils/utils_test.go
+++ b/pkg/common/utils/utils_test.go
@@ -22,6 +22,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"fmt"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


pkg/common/utils/utils.go
```diff
@@ -44,6 +45,10 @@ const userInfoKey = siCommon.DomainYuniKorn + "user.info"
 const uniqueAutogenSuffix = "-uniqueautogen"
 
 var pluginMode bool
+var (
+	// ErrorTimeout returned if waiting for a condition times out
+	ErrorTimeout = errors.New("timeout waiting for condition")
```
chia7712, line 50: this can't be constant, so maybe we can focus on "error message" instead of `Error`. For example:
```go
const timeoutErrorMessage="timeout waiting for condition"
```

and then in testing we can check the error message instead of "error"

chia7712 (commented):
@SP12893678 thanks for this nice test

pkg/common/utils/utils.go
SP12893678: OK. I will use `ErrorContains `to check it.
In core repo, `WaitForCondition `unit test have implemented with the same usage.
I will create a issue to improve it.

SP12893678: Could have a look for this?
I think we should have a consistent way.

https://github.com/apache/yunikorn-core/pull/875#discussion_r1621576743

chia7712: ok, let's keep the consistency

pkg/common/utils/utils_test.go
```diff
@@ -1140,3 +1211,30 @@ func TestConvert2PriorityClass(t *testing.T) {
...
+	for _, test := range tests {
+		target = test.input
+		get := WaitForCondition(eval, test.timeout, test.interval)
+		if test.output == nil {
+			assert.NilError(t, get)
+		} else {
+			assert.Equal(t, get.Error(), test.output.Error())
+		}
+	}
+}
```
chenyulin0719, line 1240: The TestWaitForCondition() takes 5~6 secs (the second and forth test cases) , I guess it's accepted. But  I'm also thinking about whether we could use a smaller `interval` and `timout`, like 100 Millisecond and 200 Millisecond?

The only concern is I'm not sure whehter our CI env's goroutine could got CPU in time.
@chia7712 WDYT?

chenyulin0719 (commented):
A few comments.

pkg/common/utils/utils_test.go
chenyulin0719: Or, maybe we could removie this test? I'm not expecting any code change to this function in the future.

chia7712: I guess that test case is another sort of "consistent result"  https://github.com/apache/yunikorn-core/blob/21303191a9c9ee791de3371ddca973df51b7a755/pkg/common/utils_test.go#L239

Hence, I'm fine to keep this test case :)

chenyulin0719: OK, let's keep the test cases consistent with core.

pkg/common/utils/utils_test.go
```diff
@@ -1075,6 +1139,13 @@ func TestGetCoreSchedulerConfigFromConfigMap(t *testing.T) {
 	assert.Equal(t, "test", GetCoreSchedulerConfigFromConfigMap(cm))
 }
 
+func TestGetCoreSchedulerConfigNotMapping(t *testing.T) {
+	cm := map[string]string{
+		"unknow.yaml": "test",
+	}
+	assert.Equal(t, "", GetCoreSchedulerConfigFromConfigMap(cm))
+}
+
```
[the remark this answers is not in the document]
SP12893678: No problem

chenyulin0719 (approved):
+1 LGTM

chenyulin0719:
@SP12893678 Could you please fix the lint issue in CI.

chenyulin0719 (changes requested):
@SP12893678 Could you please fix the lint issue in CI.

SP12893678:
> @SP12893678 Could you please fix the lint issue in CI.

Sure. I have fixed lint problem currently. 
But, I think the funlen lint issue might not fully addressed in https://github.com/apache/yunikorn-k8shim/pull/861
@rich7420, Could you please check against?

chenyulin0719 (approved):
+1 LGTM

PR #892, opened 2024-08-15
[YUNIKORN-2807] Don't log 'task missing' as ERROR

craigcondit:
### What is this PR for?
When a task no longer exists but we have an event to process, we log very loudly. This is not uncommon, and happens frequently when pods are deleted. Log at DEBUG instead.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2807

### Questions:


chia7712 (approved):
LGTM

PR #896, opened 2024-08-18
[YUNIKORN-2817] Pods should be deleted immediately when they are plac…

zhuqi-lucas:
…eholder pods

### What is this PR for?
Pods should be deleted immediately when they are placeholder pods, for example when placeholder timeout, we clean up them and the shim side should finally delete those pods without delay.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-2817

### Questions:

### DIFF
--- a/pkg/client/kubeclient.go
+++ b/pkg/client/kubeclient.go
@@ -21,6 +21,7 @@ package client
 import (
 	"context"
 	"fmt"
+	"github.com/apache/yunikorn-k8shim/pkg/common/utils"
 
 	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
@@ -139,6 +140,9 @@ func (nc SchedulerKubeClient) Create(pod *v1.Pod) (*v1.Pod, error) {
 
 func (nc SchedulerKubeClient) Delete(pod *v1.Pod) error {
 	gracefulSeconds := int64(3)
+	if utils.GetPlaceholderFlagFromPodSpec(pod) {
+		gracefulSeconds = int64(0)
+	}
 	if err := nc.clientSet.CoreV1().Pods(pod.Namespace).Delete(context.Background(), pod.Name, apis.DeleteOptions{
 		GracePeriodSeconds: &gracefulSeconds,
 	}); err != nil {
--- a/pkg/client/kubeclient.go
+++ b/pkg/client/kubeclient.go
@@ -31,7 +31,6 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/retry"
 
-	"github.com/apache/yunikorn-k8shim/pkg/common/utils"
 	"github.com/apache/yunikorn-k8shim/pkg/conf"
 	"github.com/apache/yunikorn-k8shim/pkg/log"
 )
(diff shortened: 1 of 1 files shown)


craigcondit (changes requested):
This is unnecessary. We already create the placeholder pods with the timeout set to 0. There's no need to do it again here.

zhuqi-lucas:
Thanks @craigcondit for review, i think we may need to change both setting for pod spec and also the actually deleting settings, and the actually deleting settings will be executed. There are two timeout for pod, when we create the placeholder pod we set the pod spec timeout to zero now, but when we actually delete the pod, the timeout will be potentially override by the for API timeout setting:

For creating placeholder, we setting in this code:

https://github.com/apache/yunikorn-k8shim/blob/2278b3217c702ccb796e4d623bc7837625e5a4ec/pkg/cache/placeholder.go#L124


``` golang
// Optional duration in seconds the pod needs to terminate gracefully. May be decreased in delete request.
	// Value must be non-negative integer. The value zero indicates stop immediately via
	// the kill signal (no opportunity to shut down).
	// If this value is nil, the default grace period will be used instead.
	// The grace period is the duration in seconds after the processes running in the pod are sent
	// a termination signal and the time when the processes are forcibly halted with a kill signal.
	// Set this value longer than the expected cleanup time for your process.
	// Defaults to 30 seconds.
	// +optional
	TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds,omitempty" protobuf:"varint,4,opt,name=terminationGracePeriodSeconds"`
```

Actually pod deletion override code, we will finally call this for all pod deletion:
https://github.com/apache/yunikorn-k8shim/blob/2278b3217c702ccb796e4d623bc7837625e5a4ec/pkg/client/kubeclient.go#L143

``` golang
// The duration in seconds before the object should be deleted. Value must be non-negative integer.
	// The value zero indicates delete immediately. If this value is nil, the default grace period for the
	// specified type will be used.
	// Defaults to a per object value if not specified. zero means delete immediately.
	// +optional
	GracePeriodSeconds *int64 `json:"gracePeriodSeconds,omitempty" protobuf:"varint,1,opt,name=gracePeriodSeconds"`
```

craigcondit:
I wasn't aware we were setting 3 seconds currently. This is wrong; we shouldn't be overriding pod settings in deletion like that. The real fix here is to remove the 3 second override. Imagine if we delete a non-placeholder pod such as a database: Forcing non-graceful termination after 3 seconds could lead to data corruption. The override has no place in the generic Delete() call.

craigcondit:
This could also explain some performance issues seen in gang scheduling. If all placeholders are taking 3 seconds to be deleted that adds significant delays to gang scheduling jobs.

zhuqi-lucas:
> This could also explain some performance issues seen in gang scheduling. If all placeholders are taking 3 seconds to be deleted that adds significant delays to gang scheduling jobs.

Yeah, i agree, and i am in favor of  removing this override, it's more reasonable instead of adding new zero override. And final question for non placeholder pod, we default 3 seconds here, do we also need to remove it, or we only remove the placeholder pods override? If we not override the non-placeholder pods, it's default timeout should be 30s.

craigcondit:
When deleting, the default is whatever the pod specified and 30 seconds otherwise. We shouldn't change that logic; just remove the 3s override we have now.

zhuqi-lucas:
Thanks @craigcondit , reasonable suggestion and addressed in latest PR.

craigcondit (approved):
+1 LGTM.

PR #897, opened 2024-08-19
[YUNIKORN-2808] E2E test Verify_preemption_on_priority_queue test is flaky

manirajv06:
### What is this PR for?
E2E test Verify_preemption_on_priority_queue test is flaky

### What type of PR is it?
* [ ] - Bug Fix

### Todos
* [ ] - Task

### Questions:

### DIFF
--- a/test/e2e/preemption/preemption_test.go
+++ b/test/e2e/preemption/preemption_test.go
@@ -52,6 +52,8 @@ var sleepPodMemLimit int64
 var sleepPodMemLimit2 int64
 var taintKey = "e2e_test_preemption"
 var nodesToTaint []string
+var cantBeScheduledMesg = "The sleep pod %s can't be scheduled"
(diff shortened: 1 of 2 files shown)


manirajv06:
Preemption use single node for all these tests for simplicity. Failures happened only in Verify_preemption_on_priority_queue test. In this test, app submitted in low-priority queue has entered into reservation flow instead of going through preemption flow and eventually rejected because of its lower priority, followed by app waiting in high-priority queue not being able to complete the preemption because of node reserved  by app submitted in low-priority queue. This specific situation doesn't happen always in all runs. Now, modified a test a bit to trigger the preemption while the headroom checks itself for both the apps so that high priority app is able to get resources through preemption whereas low priority app is not able to.

craigcondit (approved):
+1 LGTM.

PR #908, opened 2024-09-05
[YUNIKORN-2852]Added test cases for group names in placement rule filter

rrajesh-cloudera:
This PR introduces positive and negative test cases for configuring the scheduler with group names in the placement rule filter. 

In the positive test case, a valid group name is used to configure the scheduler, and the test verifies that the configuration update is successful. 

In the negative test case, an invalid group name is used in an attempt to configure the scheduler. The test verifies that the configuration update fails as expected when an invalid group name is used in the placement rule filter. 

These tests ensure that the scheduler correctly handles group names in the placement rule filter and properly rejects invalid group names.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2852

### Questions:

### DIFF
--- a/test/e2e/configmap/configmap_test.go
+++ b/test/e2e/configmap/configmap_test.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"time"
 
+	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


pbacsko (approved):
LGTM

PR #914, opened 2024-09-11
[YUNIKORN-2874] Fix reproducible builds when a release is generated

craigcondit:
### What is this PR for?
Ensure that the parent directory (containing the core and SI) is mounted into the Docker build when generating reproducible builds in release mode.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2874

Verified manually by applying this patch to a release tarball. When .gitignore is not detected, we mount the parent directory into the Docker build environment rather than just the shim dir.

### Questions:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ else
 endif
 
 # Release build requires using parent dir as base for buildroot
-RELEASE_BUILD := $(test -f "$(BASE_DIR)/.gitignore" ; echo $?)
+RELEASE_BUILD := $(test -f "$(BASE_DIR)/.gitignore" ; echo $$?)
 ifeq ($(RELEASE_BUILD),1)
 	DOCKER_BUILDROOT := $(shell cd "$(BASE_DIR)/.." ; pwd)
 	DOCKER_SRCROOT := /buildroot/k8shim
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,6 @@ endif
 
 # Make sure we are in the same directory as the Makefile
 BASE_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
-PARENT_DIR := $(dir $(abspath $(BASE_DIR)/..))
 
 # Output directories
 OUTPUT=build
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -89,6 +90,16 @@ else
   REPRO :=
 endif
 
+# Release build requires using parent dir as base for buildroot
+RELEASE_BUILD := $(test -f "$(BASE_DIR)/.gitignore" ; echo $?)
```
pbacsko, line 94: I tried this locally but it didn't work. Spent some time debugging it, found out that this solves the problem:

```
RELEASE_BUILD := $(shell test -f "$(BASE_DIR)/.gitignore" ; echo $$?)
```

pbacsko (changes requested):
See comment

Makefile
craigcondit: Good catch, fixed.

Makefile
```diff
@@ -44,6 +44,7 @@ endif
 
 # Make sure we are in the same directory as the Makefile
 BASE_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+PARENT_DIR := $(dir $(abspath $(BASE_DIR)/..))
```
manirajv06, line 47: PARENT_DIR is set but not used later. Probably for DOCKER_BUILDROOT ?

manirajv06 (changes requested):
Is is safer to go one level up from incubator-yunikorn-k8shim directory?

Makefile
craigcondit: Removed in latest commit.

craigcondit:
> Is is safer to go one level up from incubator-yunikorn-k8shim directory?

We don't know for sure that a user has the different trees checked out in parallel directories. However, in the tarball build we do -- we check for this by the presence or absence of .gitignore (it won't be present in a tarball build). This way, builds work consistently if only k8shim is checked out, or when using a release tarball.

PR #915, opened 2024-09-14
[YUNIKORN-1977] E2E test for verifying user info with non kube-admin user

rrajesh-cloudera:
This PR adds an end-to-end (E2E) test that verifies pod deployment and user info using a non-`kube-admin` user. The test ensures that RBAC roles and permissions for non-admin users are correctly configured. Specifically, it verifies that user information can be retrieved through pod annotations.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1977

### Questions:

### DIFF
--- a/test/e2e/framework/helpers/k8s/k8s_utils.go
+++ b/test/e2e/framework/helpers/k8s/k8s_utils.go
@@ -17,14 +17,12 @@
 package k8s
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
@@ -50,6 +48,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	"k8s.io/client-go/tools/portforward"
 	"k8s.io/client-go/transport/spdy"
 	"k8s.io/client-go/util/retry"
@@ -156,6 +155,13 @@ func (k *KubeCtl) GetClient() *kubernetes.Clientset {
 func (k *KubeCtl) GetKubeConfig() (*rest.Config, error) {
 	if k.kubeConfig != nil {
 		return k.kubeConfig, nil
+	}
+	return nil, errors.New("kubeconfig is nil")
+}
+
+func (k *KubeCtl) GetKubeConfigNew() (*rest.TLSClientConfig, error) {
+	if k.kubeConfig != nil {
+		return &k.kubeConfig.TLSClientConfig, nil
 	}
 	return nil, errors.New("kubeconfig is nil")
 }
@@ -1091,6 +1097,58 @@ func (k *KubeCtl) CreateSecret(secret *v1.Secret, namespace string) (*v1.Secret,
 	return k.clientSet.CoreV1().Secrets(namespace).Create(context.TODO(), secret, metav1.CreateOptions{})
 }
 
+// SetClientFromConfig initializes a new Kubernetes clientset from the given config and sets it in kClient
+func (k *KubeCtl) SetClientFromConfig(newConf *rest.Config) error {
+	// Create a new Kubernetes clientset from the new configuration
+	clientset, err := kubernetes.NewForConfig(newConf)
+	if err != nil {
+		return err // Return the error if clientset creation fails
+	}
+
+	// Set the new clientset to the kClient object
+	k.clientSet = clientset
+	return nil
+}
+
+func WriteConfigToFile(config *rest.Config, kubeconfigPath string) error {
+	// Build the kubeconfig API object from the rest.Config
+	kubeConfig := &clientcmdapi.Config{
+		Clusters: map[string]*clientcmdapi.Cluster{
+			"default-cluster": {
+				Server:                   config.Host,
+				CertificateAuthorityData: config.CAData,
+				InsecureSkipTLSVerify:    config.Insecure,
+			},
+		},
+		AuthInfos: map[string]*clientcmdapi.AuthInfo{
+			"default-auth": {
+				Token: config.BearerToken,
+			},
+		},
+		Contexts: map[string]*clientcmdapi.Context{
+			"default-context": {
+				Cluster:  "default-cluster",
+				AuthInfo: "default-auth",
+			},
+		},
+		CurrentContext: "default-context",
+	}
+
+	// Ensure the directory where the file is being written exists
+	err := os.MkdirAll(filepath.Dir(kubeconfigPath), os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("failed to create directory for kubeconfig file: %v", err)
+	}
+
+	// Write the kubeconfig to the specified file
+	err = clientcmd.WriteToFile(*kubeConfig, kubeconfigPath)
+	if err != nil {
+		return fmt.Errorf("failed to write kubeconfig to file: %v", err)
+	}
+
+	return nil
+}
+
 func GetSecretObj(yamlPath string) (*v1.Secret, error) {
 	o, err := common.Yaml2Obj(yamlPath)
 	if err != nil {
@@ -1381,21 +1439,6 @@ func (k *KubeCtl) isNumJobPodsInDesiredState(jobName string, namespace string, n
 
 		return counter >= num, nil
 	}
-}
-
-func ApplyYamlWithKubectl(path, namespace string) error {
-	cmd := exec.Command("kubectl", "apply", "-f", path, "-n", namespace)
-	var stderr bytes.Buffer
-	cmd.Stderr = &stderr
-	// if err != nil, isn't represent yaml format error.
-	// it only represent the cmd.Run() fail.
-	err := cmd.Run()
-	// if yaml format error, errStr will show the detail
-	errStr := stderr.String()
-	if err != nil && errStr != "" {
-		return fmt.Errorf("apply fail with %s", errStr)
-	}
-	return nil
 }
 
 func (k *KubeCtl) GetNodes() (*v1.NodeList, error) {
@@ -1777,5 +1820,98 @@ func (k *KubeCtl) DeleteStorageClass(scName string) error {
 	if err != nil {
 		return err
 	}
+	return nil
+}
+
+func (k *KubeCtl) GetSecrets(namespace string) (*v1.SecretList, error) {
+	return k.clientSet.CoreV1().Secrets(namespace).List(context.TODO(), metav1.ListOptions{})
+}
+
+func (k *KubeCtl) GetSecretValue(namespace, secretName, key string) (string, error) {
+	var secret *v1.Secret
+	var err error
+
+	// Retry loop in case secret is not yet populated
+	for i := 0; i < 5; i++ {
+		secret, err = k.clientSet.CoreV1().Secrets(namespace).Get(context.TODO(), secretName, metav1.GetOptions{})
+		if err != nil {
+			return "", err
+		}
+
+		// Check if the secret contains data
+		if len(secret.Data) > 0 {
+			value, ok := secret.Data[key]
+			if !ok {
+				return "", fmt.Errorf("key %s not found in secret %s", key, secretName)
+			}
+			return string(value), nil
+		}
+
+		// Wait before retrying
+		time.Sleep(2 * time.Second)
+	}
+
+	return "", fmt.Errorf("secret %s has no data after retries", secretName)
+}
+
+func (k *KubeCtl) GetCurrentContext() (string, error) {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	configOverrides := &clientcmd.ConfigOverrides{}
+	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
+
+	rawConfig, err := kubeConfig.RawConfig()
+	if err != nil {
+		return "", err
+	}
+	return rawConfig.CurrentContext, nil
+}
+
+func (k *KubeCtl) GetCurrentCluster() (string, error) {
+	// get the current context cluster name
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	configOverrides := &clientcmd.ConfigOverrides{}
+	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
+	rawConfig, err := kubeConfig.RawConfig()
+	if err != nil {
+		return "", err
+	}
+	return rawConfig.Contexts[rawConfig.CurrentContext].Cluster, nil
+}
+
+func (k *KubeCtl) GetClusterCA() ([]byte, error) {
+	// Get the current clustername
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	configOverrides := &clientcmd.ConfigOverrides{}
+	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
+	rawConfig, err := kubeConfig.RawConfig()
+	if err != nil {
+		return nil, err
+	}
+	return rawConfig.Clusters[rawConfig.Contexts[rawConfig.CurrentContext].Cluster].CertificateAuthorityData, nil
+}
+
+func (k *KubeCtl) GetClusterServer() (string, error) {
+	// Get the current clustername
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	configOverrides := &clientcmd.ConfigOverrides{}
+	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
+	rawConfig, err := kubeConfig.RawConfig()
+	if err != nil {
+		return "", err
+	}
+	return rawConfig.Clusters[rawConfig.Contexts[rawConfig.CurrentContext].Cluster].Server, nil
+}
+
+func (k *KubeCtl) UpdateConfig(newConf *rest.Config) error {
+	_, err := clientcmd.BuildConfigFromFlags("", clientcmd.RecommendedHomeFile)
+	if err != nil {
+		return err
+	}
+	newClientset, err := kubernetes.NewForConfig(newConf)
+	if err != nil {
+		return err
+	}
+	k.clientSet = newClientset
+	k.kubeConfig = newConf
 	return nil
 }
--- a/test/e2e/framework/helpers/k8s/k8s_utils.go
+++ b/test/e2e/framework/helpers/k8s/k8s_utils.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
 	"go.uber.org/zap"
 	appsv1 "k8s.io/api/apps/v1"
@@ -48,7 +49,6 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
-	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	"k8s.io/client-go/tools/portforward"
 	"k8s.io/client-go/transport/spdy"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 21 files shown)


test/e2e/user_group_limit/user_group_limit_test.go
```diff
@@ -911,6 +925,200 @@ var _ = ginkgo.Describe("UserGroupLimit", func() {
...
+				}
+				return common.AddQueue(sc, constants.DefaultPartition, constants.RootQueue, configs.QueueConfig{Name: "sandbox2"})
+			})
+		})
+		// Backup the existing kubeconfig
+		oldKubeconfigPath := filepath.Join(os.Getenv("HOME"), ".kube", "config")
+		if _, err := os.Stat(oldKubeconfigPath); !os.IsNotExist(err) {
+			oldKubeconfigContent, err = os.ReadFile(oldKubeconfigPath)
+			gomega.Ω(err).NotTo(HaveOccurred())
+		}
```
ryankert01, line 977: It looks like backing up the original config file, but I don't see any code that modifies the original file. Could you point it out? or is it redundant?

ryankert01 (commented):
Hi, It looks great! having a simple question.

test/e2e/user_group_limit/user_group_limit_test.go
rrajesh-cloudera: Hi, @ryankert01  Thanks for Pointing out. I forgot to add the steps to restore oldkubeconfig after the test. I have updated the code for restoring oldkubeconfig.

ryankert01: @rrajesh-cloudera I mean it's dangerous to use backup-restore in case of the test fail which it'll not restore. Is it possible to use a temporary config file similar to what you just did around L1048? (An idea that might work would be modifying `KUBECONFIG` env variable, and using `defer` to finalize it to prevent not restoring when test fails.

rrajesh-cloudera: Hi @ryankert01  , The test code is not modifying the existing kubeconfig file; it’s creating a temporary kubeconfig file, using that file to set the context, and creating a pod. I can remove the part of the code where we are storing the old kubeconfig content and setting it back to the default kubeconfig location, as it’s creating unnecessary confusion.

test/e2e/user_group_limit/user_group_limit_test.go
```diff
@@ -911,6 +925,206 @@ var _ = ginkgo.Describe("UserGroupLimit", func() {
...
+		ginkgo.By("Creating Secret for the Service Account...")
+		_, err = clientset.CoreV1().Secrets(namespace).Create(context.TODO(), &v1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: secretName,
+				Annotations: map[string]string{
+					"kubernetes.io/service-account.name": serviceAccountName,
+				},
+			},
+			Type: v1.SecretTypeServiceAccountToken,
+		}, metav1.CreateOptions{})
```
pbacsko, line 1033: Please use `KubeCtl.CreateSecret()` if possible. If the annotation is necessary, create a new method with the name `CreateSecretWithAnnotation(secret *v1.Secret, namespace string, annotations map[string]string)` to avoid calling this directly.

test/e2e/user_group_limit/user_group_limit_test.go
```diff
@@ -1022,3 +1249,45 @@ func checkUsageWildcardGroups(testType TestType, name string, queuePath string,
...
+	kubeconfigContent := strings.ReplaceAll(kubeconfigTemplate, "${CURRENT_CONTEXT}", currentContext)
+	kubeconfigContent = strings.ReplaceAll(kubeconfigContent, "${CLUSTER_CA}", clusterCA)
+	kubeconfigContent = strings.ReplaceAll(kubeconfigContent, "${CLUSTER_SERVER}", clusterServer)
+	kubeconfigContent = strings.ReplaceAll(kubeconfigContent, "${USER_TOKEN_VALUE}", userTokenValue)
+
+	// Write the kubeconfig YAML to the file
+	err := os.WriteFile(path, []byte(kubeconfigContent), 0600)
+	gomega.Ω(err).NotTo(gomega.HaveOccurred())
+	return nil
+}
```
pbacsko, line 1283: All of this should NOT be necessary. Way too complicated to mess around with separate `kubectl` calls.

You can do this:
```
config, _ := kClient.GetKubeConfig() // handle error in real code

newConf := config.DeepCopy()  // copy existing config
newConf .TLSClientConfig.CertFile = ""  // remove cert file
newConf .TLSClientConfig.KeyFile = ""  // remove key file
newConf .BearerToken = "<base64Token>"  // set token that is retrieved in the test
_ = kClient.SetClientFromConfig(newConf)
```
After this point, `kClient` will use the token for authentication there's no need to delete/restore anything.

New method is necessary in `KubeCtl`:
```
func (k *KubeCtl) SetClientFromConfig(conf *rest.Config) error {
    k.kubeConfig = conf.DeepCopy()
    k.clientSet, err = kubernetes.NewForConfig(k.kubeConfig)   // creates new clientset 
    return err
} 
```

Also, try to retrieve the secret token using `KubeCtl`. We might need to create a new method for it, but again, it shouldn't involve `kubectl`:

```
func (k *KubeCtl) GetSecret(namespace string) (*v1.Secret, error) {
    return k.clientSet.CoreV1().Secrets(namespace).Get(context.TODO(), namespace, metav1.GetOptions{})
}
```

pbacsko (changes requested):
-1 I'm not a fan of the `kubectl` approach, it's completely unnecessary. It makes the test more complicated. It's possible to use the API.

test/e2e/user_group_limit/user_group_limit_test.go
rrajesh-cloudera: Sure @pbacsko will accommodate the requested changes.

test/e2e/user_group_limit/user_group_limit_test.go
ryankert01: My request is covered by Peter's, mark this as resolved.

pbacsko:
@rrajesh-cloudera please fix the merge conflicts in `user_group_limit_test.go`

rrajesh-cloudera:
> @rrajesh-cloudera please fix the merge conflicts in `user_group_limit_test.go`

Hi @pbacsko , resolved the Merge Conflict.

test/e2e/user_group_limit/user_group_limit_test.go
```diff
@@ -909,6 +923,190 @@ var _ = ginkgo.Describe("UserGroupLimit", func() {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: secretName,
+				Annotations: map[string]string{
+					"kubernetes.io/service-account.name": serviceAccountName,
+				},
+			},
+			Type: v1.SecretTypeServiceAccountToken,
+		}
+		time.Sleep(10 * time.Second)
+		_, err = kClient.CreateSecret(secret, namespace)
```
ryankert01, line 1023: Could you use `Eventually` in this case? time.Sleep can lead to flaky tests.

test/e2e/user_group_limit/user_group_limit_test.go
```diff
@@ -909,6 +923,190 @@ var _ = ginkgo.Describe("UserGroupLimit", func() {
...
+		gomega.Ω(err).NotTo(HaveOccurred())
+		time.Sleep(10 * time.Second)
```
ryankert01, line 1025: Could you use `Eventually` in this case? time.Sleep can lead to flaky tests.

test/e2e/user_group_limit/user_group_limit_test.go
```diff
@@ -909,6 +923,190 @@ var _ = ginkgo.Describe("UserGroupLimit", func() {
...
+		// Set the Kubernetes API server URL
+		newConf.BearerToken = userTokenValue
+		newConf.Username = ""
+		newConf.Password = ""
+		kubeconfigPath := filepath.Join(os.TempDir(), "test-user-config")
+		err = k8s.WriteConfigToFile(newConf, kubeconfigPath)
+		gomega.Ω(err).NotTo(gomega.HaveOccurred())
+		config, err = clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+		gomega.Ω(err).NotTo(HaveOccurred())
+		clientset, err = kubernetes.NewForConfig(config)
```
ryankert01, line 1048: Could we simplify this six line to:
```
clientset, err = kubernetes.NewForConfig(newConf)
```

test/e2e/user_group_limit/user_group_limit_test.go
```diff
@@ -909,6 +923,190 @@ var _ = ginkgo.Describe("UserGroupLimit", func() {
...
+				},
+			},
+		}
+		_, err = clientset.CoreV1().Pods(namespace).Create(context.TODO(), pod, metav1.CreateOptions{})
+		gomega.Ω(err).NotTo(HaveOccurred())
+		createdPod, err := clientset.CoreV1().Pods(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
+		gomega.Ω(err).NotTo(HaveOccurred())
+		ginkgo.By("Verifying User Info...")
+		userInfo, err := GetUserInfoFromPodAnnotation(createdPod)
+		gomega.Ω(err).NotTo(gomega.HaveOccurred())
```
ryankert01, line 1077:
```suggestion
		gomega.Ω(err).NotTo(HaveOccurred())
```

test/e2e/user_group_limit/user_group_limit_test.go
```diff
@@ -909,6 +923,190 @@ var _ = ginkgo.Describe("UserGroupLimit", func() {
...
+		// user info should contain the substring "system:serviceaccount:default:test-user-sa"
+		gomega.Ω(strings.Contains(fmt.Sprintf("%v", userInfo), "system:serviceaccount:default:test-user-sa")).To(gomega.BeTrue())
+		// Remove the new kubeconfig file
+		gomega.Ω(err).NotTo(gomega.HaveOccurred())
```
ryankert01, line 1081:
```suggestion
		gomega.Ω(err).NotTo(HaveOccurred())
```

test/e2e/user_group_limit/user_group_limit_test.go
```diff
@@ -909,6 +923,190 @@ var _ = ginkgo.Describe("UserGroupLimit", func() {
...
+		os.Unsetenv("KUBECONFIG")
+		err = kClient.SetClient()
+		gomega.Ω(err).NotTo(gomega.HaveOccurred())
```
ryankert01, line 1084:
```suggestion
		gomega.Ω(err).NotTo(HaveOccurred())
```

test/e2e/user_group_limit/user_group_limit_test.go
ryankert01: or could we further optimize it using TokenRequest API instead of manually create secret and wait until it pupulate.

ryankert01 (changes requested):
Thank you for your patch, some questions and nits left.

test/e2e/user_group_limit/user_group_limit_test.go
ryankert01:
```suggestion
		clientset, err = kubernetes.NewForConfig(newConf)
```

test/e2e/framework/helpers/k8s/k8s_utils.go
```diff
@@ -1091,6 +1097,58 @@ func (k *KubeCtl) CreateSecret(secret *v1.Secret, namespace string) (*v1.Secret,
...
+	}
+
+	// Write the kubeconfig to the specified file
+	err = clientcmd.WriteToFile(*kubeConfig, kubeconfigPath)
+	if err != nil {
+		return fmt.Errorf("failed to write kubeconfig to file: %v", err)
+	}
+
+	return nil
+}
```
pbacsko, line 1150: Do we really need this? I think we can get away with writing stuff to files.

test/e2e/framework/helpers/k8s/k8s_utils.go
```diff
@@ -1779,3 +1822,96 @@ func (k *KubeCtl) DeleteStorageClass(scName string) error {
...
+		if len(secret.Data) > 0 {
+			value, ok := secret.Data[key]
+			if !ok {
+				return "", fmt.Errorf("key %s not found in secret %s", key, secretName)
+			}
+			return string(value), nil
+		}
+
+		// Wait before retrying
+		time.Sleep(2 * time.Second)
```
pbacsko, line 1851: This retry loop & sleep should definitely not be here. A getter function should only be about retrieving something, if possible. If we have to wait for a value to show up, let's create a separate helper method for it.

test/e2e/framework/helpers/k8s/k8s_utils.go
```diff
@@ -1779,3 +1822,96 @@ func (k *KubeCtl) DeleteStorageClass(scName string) error {
...
+		return err
+	}
+	newClientset, err := kubernetes.NewForConfig(newConf)
+	if err != nil {
+		return err
+	}
+	k.clientSet = newClientset
+	k.kubeConfig = newConf
+	return nil
+}
```
pbacsko, line 1917: All these methods seem to be unused. Am I missing something?

test/e2e/user_group_limit/user_group_limit_test.go
pbacsko: +1 fixed sleeps are dangerous and can be time consuming. Let's avoid them if possible. Add helper poll method which repeatedly checks if the secret exists.

test/e2e/user_group_limit/user_group_limit_test.go
```diff
@@ -909,6 +923,190 @@ var _ = ginkgo.Describe("UserGroupLimit", func() {
...
+		createdPod, err := clientset.CoreV1().Pods(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
+		gomega.Ω(err).NotTo(HaveOccurred())
+		ginkgo.By("Verifying User Info...")
+		userInfo, err := GetUserInfoFromPodAnnotation(createdPod)
+		gomega.Ω(err).NotTo(gomega.HaveOccurred())
+		// user info should contain the substring "system:serviceaccount:default:test-user-sa"
+		gomega.Ω(strings.Contains(fmt.Sprintf("%v", userInfo), "system:serviceaccount:default:test-user-sa")).To(gomega.BeTrue())
+		// Remove the new kubeconfig file
+		gomega.Ω(err).NotTo(gomega.HaveOccurred())
+		os.Unsetenv("KUBECONFIG")
```
pbacsko, line 1082: I don't think we need to unset this variable, correct?

test/e2e/user_group_limit/user_group_limit_test.go
```diff
@@ -920,6 +1118,19 @@ var _ = ginkgo.Describe("UserGroupLimit", func() {
 	})
 })
 
+func GetUserInfoFromPodAnnotation(pod *v1.Pod) (interface{}, error) {
+	userInfo, ok := pod.Annotations[amCommon.UserInfoAnnotation]
+	if !ok {
+		return nil, fmt.Errorf("user info not found in pod annotation")
+	}
+	var userInfoObj interface{}
```
pbacsko, line 1126: We know the type of this object. It's `si.UserGroupInformation`, let's use that.

pbacsko (changes requested):
Some more comments from me as well.

test/e2e/user_group_limit/user_group_limit_test.go
rrajesh-cloudera: Updated the code with requested changes.

test/e2e/user_group_limit/user_group_limit_test.go
rrajesh-cloudera: Updated the code with requested changes.

test/e2e/user_group_limit/user_group_limit_test.go
rrajesh-cloudera: Tried the suggested solution but facing issue , It's not working with this test case.

test/e2e/framework/helpers/k8s/k8s_utils.go
rrajesh-cloudera: Removed

test/e2e/user_group_limit/user_group_limit_test.go
rrajesh-cloudera: Removed

test/e2e/user_group_limit/user_group_limit_test.go
rrajesh-cloudera: Updated the code with requested changes.

test/e2e/framework/helpers/k8s/k8s_utils.go
rrajesh-cloudera: Updated the code with requested changes.

test/e2e/framework/helpers/k8s/k8s_utils.go
rrajesh-cloudera: We need a temp file to write the kubeconfig and reload the kubeconfig of the temp file to apply the changes to Kubernetes clientset.
Tried without writing the file and copying the existing kubeconfig to newConf with bearertoken value set but that approach is not working well so that's why we needed to add the write config file to temp store the kubeconfig and reload for the k8s clientset.

test/e2e/user_group_limit/user_group_limit_test.go
```diff
@@ -909,40 +926,227 @@ var _ = ginkgo.Describe("UserGroupLimit", func() {
...
+		// Cleanup
+		ginkgo.By("Cleaning up resources...")
+		err = clientset.CoreV1().Pods(namespace).Delete(context.TODO(), podName, metav1.DeleteOptions{})
+		gomega.Ω(err).NotTo(HaveOccurred())
+		err = clientset.CoreV1().ServiceAccounts(namespace).Delete(context.TODO(), serviceAccountName, metav1.DeleteOptions{})
+		gomega.Ω(err).NotTo(HaveOccurred())
+		err = kClient.DeleteClusterRole("pod-creator-role")
+		gomega.Ω(err).NotTo(HaveOccurred())
+		err = kClient.DeleteClusterRoleBindings("pod-creator-role-binding")
+		gomega.Ω(err).NotTo(HaveOccurred())
```
pbacsko, line 1087: If the test fail, we never reach this part. This is supposed to be in a deferred section:

That is:
```
		defer func() {
				// Cleanup
				ginkgo.By("Cleaning up resources...")
				err = clientset.CoreV1().Pods(namespace).Delete(context.TODO(), podName, metav1.DeleteOptions{})
				gomega.Ω(err).NotTo(HaveOccurred())
				err = clientset.CoreV1().ServiceAccounts(namespace).Delete(context.TODO(), serviceAccountName, metav1.DeleteOptions{})
				gomega.Ω(err).NotTo(HaveOccurred())
				err = kClient.DeleteClusterRole("pod-creator-role")
				gomega.Ω(err).NotTo(HaveOccurred())
				err = kClient.DeleteClusterRoleBindings("pod-creator-role-binding")
				gomega.Ω(err).NotTo(HaveOccurred())
		}
		// Create Service Account
		ginkgo.By("Creating Service Account...")
		sa, err := kClient.CreateServiceAccount(serviceAccountName, namespace)
		gomega.Ω(err).NotTo(HaveOccurred())
```

test/e2e/framework/helpers/k8s/k8s_utils.go
pbacsko: You don't need to code the retry. We can take advantage of existing k8s functions.

To be consistent with existing code, let's use the following:

```
func (k *KubeCtl) GetSecretValue(namespace, secretName, key string) (string, error) {
	secret, err := k.GetSecret(namespace, secretName)
	if err != nil {
		return "", err
	}
	// Check if the key exists in the secret
	value, ok := secret.Data[key]
	if !ok {
		return "", fmt.Errorf("key %s not found in secret %s", key, secretName)
	}
	return string(value), nil
}

func (k *KubeCtl) GetSecret(namespace, secretName string) (*v1.Secret, error) {
	secret, err := k.clientSet.CoreV1().Secrets(namespace).Get(context.TODO(), secretName, metav1.GetOptions{})
	if err != nil {
		return nil, err
	}

	return secret, nil
}

func (k *KubeCtl) WaitForSecret(namespace, secretName string, timeout time.Duration) error {
	var cond wait.ConditionFunc
	cond = func() (done bool, err error) {
		secret, err := k.GetSecret(namespace, secretName)
		if err != nil {
			return false, err
		}
		if secret != nil {
			return false, nil
		}
		return true, nil
	}
	return wait.PollUntilContextTimeout(context.TODO(), time.Second, timeout, false, cond.WithContext())
}
```

First, you can `kClient.WaitForSecret()` to indicate that you're waiting for a secret to appear over the API. Then call `kClient.GetSecretValue()`. By having these 3 methods, all of them are usable indepdently of one another.

test/e2e/framework/helpers/k8s/k8s_utils.go
```diff
@@ -1779,3 +1755,83 @@ func (k *KubeCtl) DeleteStorageClass(scName string) error {
...
+		}
+
+		// Wait before retrying
+		time.Sleep(2 * time.Second)
+	}
+
+	return nil, fmt.Errorf("secret %s has no data after retries", secretName)
+}
+
+func WriteConfigToFile(config *rest.Config, kubeconfigPath string) error {
```
pbacsko, line 1800: I'm OK with this for the time being, let's not waste time. We can figure out later if needed.

pbacsko (changes requested):
Some more comments

pbacsko (approved):
LGTM

PR #916, opened 2024-09-27
[YUNIKORN-2884] Task fail with post allocated but the pod will keep p…

zhuqi-lucas:
…ending

### What is this PR for?
Application should also release task fail case, pod should be deleted, because TerminationType_STOPPED_BY_RM will also be called when task fail.
If we don't delete the pod, the pod will be stuck in pending state.

After apply this PR, i will add a new Jira for e2e testing, the test passed locally.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse//YUNIKORN-2884

### Questions:

### DIFF
--- a/pkg/cache/scheduler_callback.go
+++ b/pkg/cache/scheduler_callback.go
@@ -93,12 +93,11 @@ func (callback *AsyncRMCallback) UpdateAllocation(response *si.AllocationRespons
 		// update cache
 		callback.context.ForgetPod(release.GetAllocationKey())
 
-		// TerminationType 0 mean STOPPED_BY_RM
-		if release.TerminationType != si.TerminationType_STOPPED_BY_RM {
-			// send release app allocation to application states machine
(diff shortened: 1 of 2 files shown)


craigcondit:
I don't think this is the correct approach. We're arguably making things worse by terminating the pod, as this makes viewing events on it for diagnostic purposes near impossible. I think we should be looking into ways to recover from this issue instead.

zhuqi-lucas:
Thanks @craigcondit,
I see, i will do the POC and design for the perfect solution.

wilfred-s:
closing stale PR, approach should be fail the change and mark the pod as "new" and let it get scheduled again

PR #917, opened 2024-09-30
[YUNIKORN-2892] Log correct log for terminate type when releasing tas…

zhuqi-lucas:
…k from shim side

### What is this PR for?
Now we will log empty terminate type when releasing task from shim side, we should improve this to consistent with the real terminate type.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-289

### Questions:

### DIFF
--- a/pkg/cache/task.go
+++ b/pkg/cache/task.go
@@ -496,7 +496,7 @@ func (task *Task) beforeTaskCompleted() {
 func (task *Task) releaseAllocation() {
 	// We need to get the termination type from the task, when it is "" we default to the normal termination.
 	// We should log it correctly instead of "" in the logs.
-	terminationType := string(common.GetTerminationTypeFromString(task.terminationType))
+	terminationType := common.GetTerminationTypeFromString(task.terminationType)
 
 	// scheduler api might be nil in some tests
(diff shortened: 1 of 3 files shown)


pkg/cache/task.go
```diff
@@ -494,6 +494,10 @@ func (task *Task) beforeTaskCompleted() {
 
 // releaseAllocation sends the release request for the Allocation to the core.
 func (task *Task) releaseAllocation() {
+	// We need to get the termination type from the task, when it is "" we default to the normal termination.
+	// We should log it correctly instead of "" in the logs.
```
pbacsko, line 498: I think this comment can be removed, it's very obvious why we're doing this

zhuqi-lucas: Addressed in latest PR, thanks!

pbacsko (approved):
LGTM

craigcondit (approved):
+1 LGTM.

PR #922, opened 2024-10-09
[YUNIKORN-2844] Inject event recorder externally

pbacsko:
### What is this PR for?
Don't create the event recorder inside `GetRecorder()` because it has to call `IsTestMode()`. Afthis this PR, we can remove that method.
Instead, use a mock recorder by default, then inject the real one when YuniKorn starts.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2844

### Questions:

### DIFF
--- a/pkg/common/events/recorder.go
+++ b/pkg/common/events/recorder.go
@@ -19,43 +19,22 @@
 package events
 
 import (
-	"sync"
+	"sync/atomic"
 
-	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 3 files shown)


craigcondit (approved):
+1 LGTM.

PR #924, opened 2024-10-09
[YUNIKORN-2910] Fix data corruption due to insufficient shim context locking

craigcondit:
### What is this PR for?
Restore context locking that was removed as part of YUNIKORN-2629. The locks are necessary to prevent logical data corruption due to concurrent processing of both pod and node events.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2910

Ensure existing tests still pass and new deadlocks are not introduced. 

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -72,7 +72,7 @@ type Context struct {
 	pluginMode     bool                           // true if we are configured as a scheduler plugin
 	namespace      string                         // yunikorn namespace
 	configMaps     []*v1.ConfigMap                // cached yunikorn configmaps
-	lock           *locking.RWMutex               // lock
+	lock           *locking.RWMutex               // lock - used not only for context data but also to ensure that multiple event types are not executed concurrently
 	txnID          atomic.Uint64                  // transaction ID counter
 	klogger        klog.Logger
 }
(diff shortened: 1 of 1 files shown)


zhuqi-lucas (commented):
Do we have a good way to mock or reproduce those data race/corruption issues?

It will be very helpful if we can do it easily, for both e2e testing or unit test.

pbacsko (approved):
+1

pbacsko:
> Do we have a good way to mock or reproduce those data race/corruption issues?
> 
> It will be very helpful if we can do it easily, for both e2e testing or unit test.

zhuqi-lucas:
> > Do we have a good way to mock or reproduce those data race/corruption issues?
> > It will be very helpful if we can do it easily, for both e2e testing or unit test.
> 
> I don't think so. It's very difficult to repro things like that. Sometimes it's totally workload dependent.

Ok, i see.

shravan-achar:
I am working on making issues more reproducible. Will share updates more widely.

PR #928, opened 2024-10-17
[YUNIKORN-2933] Don't add duplicated taskGroup to app

XbaoWu:
### What is this PR for?
I think that before properly handling ph tasks with different resource specifications within the same task group, it is necessary to avoid having two task-group with the same name but different resource specifications in the app.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
Issue associated with this PR：https://issues.apache.org/jira/browse/YUNIKORN-2933

1. Submit a taskgroup with the same name, for example:
```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: gang-scheduling-job-example-1
  namespace: [namespace Name]
spec:
  completions: 2
  parallelism: 2
  template:
    metadata:
      labels:
        app: sleep
        applicationId: "gang-scheduling-job-example-1"
        queue: root.gang-example
      annotations:
        yunikorn.apache.org/task-group-name: task-group-example
        yunikorn.apache.org/task-groups: |-
          [{
              "name": "task-group-example",
              "minMember": 1,
              "minResource": {
                "cpu": "1",
                "memory": "50M"
              }
          },{
              "name": "task-group-example",
              "minMember": 2,
              "minResource": {
                "cpu": "1",
                "memory": "100M"
              }
          }]
    spec:
      schedulerName: yunikorn
      restartPolicy: Never
      containers:
        - name: sleep30
          image: "busybox:stable"
          imagePullPolicy: IfNotPresent
          command: ["sleep", "300"]
          resources:
            limits:
              cpu: "1"
              memory: "100M"
```
2. Observe whether only **2** ph Pods are created on K8S (`kubectl -n [namespace Name] get pod -w`)
```
NAME                                  READY   STATUS    RESTARTS   AGE
gang-scheduling-job-example-1-84ld5   0/1     Pending   0          13s
gang-scheduling-job-example-1-j9rpd   0/1     Pending   0          13s
tg-gang-scheduling-job-example--task-group-example-eu770s0adb   0/1     Pending   0          17s
tg-gang-scheduling-job-example--task-group-example-qaoxan96yt   0/1     Pending   0          17s
tg-gang-scheduling-job-example--task-group-example-eu770s0adb   0/1     Pending   0          18s
tg-gang-scheduling-job-example--task-group-example-qaoxan96yt   0/1     Pending   0          18s
tg-gang-scheduling-job-example--task-group-example-eu770s0adb   0/1     ContainerCreating   0          19s
tg-gang-scheduling-job-example--task-group-example-qaoxan96yt   0/1     ContainerCreating   0          19s
tg-gang-scheduling-job-example--task-group-example-eu770s0adb   0/1     Terminating         0          19s
tg-gang-scheduling-job-example--task-group-example-qaoxan96yt   0/1     Terminating         0          20s
tg-gang-scheduling-job-example--task-group-example-eu770s0adb   1/1     Terminating         0          20s
tg-gang-scheduling-job-example--task-group-example-qaoxan96yt   1/1     Terminating         0          21s
tg-gang-scheduling-job-example--task-group-example-eu770s0adb   0/1     Terminating         0          21s
gang-scheduling-job-example-1-84ld5                             0/1     Pending             0          49s
tg-gang-scheduling-job-example--task-group-example-eu770s0adb   0/1     Terminating         0          21s
tg-gang-scheduling-job-example--task-group-example-eu770s0adb   0/1     Terminating         0          21s
gang-scheduling-job-example-1-84ld5                             0/1     ContainerCreating   0          49s
tg-gang-scheduling-job-example--task-group-example-qaoxan96yt   0/1     Terminating         0          22s
gang-scheduling-job-example-1-j9rpd                             0/1     Pending             0          50s
tg-gang-scheduling-job-example--task-group-example-qaoxan96yt   0/1     Terminating         0          22s
tg-gang-scheduling-job-example--task-group-example-qaoxan96yt   0/1     Terminating         0          22s
gang-scheduling-job-example-1-j9rpd                             0/1     ContainerCreating   0          50s
gang-scheduling-job-example-1-84ld5                             1/1     Running             0          51s
gang-scheduling-job-example-1-j9rpd                             1/1     Running             0          52s
```

### Questions:

### DIFF
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -48,7 +48,7 @@ type Application struct {
 	groups                     []string
 	taskMap                    map[string]*Task
 	tags                       map[string]string
-	taskGroups                 []TaskGroup
+	taskGroups                 map[string]TaskGroup
 	taskGroupsDefinition       string
 	schedulingParamsDefinition string
 	placeholderOwnerReferences []metav1.OwnerReference
@@ -80,7 +80,7 @@ func NewApplication(appID, queueName, user string, groups []string, tags map[str
 		taskMap:                 taskMap,
 		tags:                    tags,
 		sm:                      newAppState(),
-		taskGroups:              make([]TaskGroup, 0),
+		taskGroups:              make(map[string]TaskGroup),
 		lock:                    &locking.RWMutex{},
 		schedulerAPI:            scheduler,
 		placeholderTimeoutInSec: 0,
@@ -163,12 +163,33 @@ func (app *Application) GetSchedulingParamsDefinition() string {
 	return app.schedulingParamsDefinition
 }
 
+// check if the task-groups is correct
+func (app *Application) checkTaskGroups(taskGroups []TaskGroup, pod *v1.Pod) {
+	tgs := make(map[string]TaskGroup)
+
+	for _, taskGroup := range taskGroups {
+		if _, exists := tgs[taskGroup.Name]; exists {
+			// for duplicated task-group, users will receive the event
+			log.Log(log.ShimCacheApplication).Warn("duplicate task-group within the task-groups",
+				zap.String("appID", app.applicationID),
+				zap.String("groupName", taskGroup.Name))
+			events.GetRecorder().Eventf(pod.DeepCopy(), nil, v1.EventTypeWarning, "GangScheduling",
+				"TaskGroupDuplicated", "Application %s has duplicated task-group %s", app.applicationID, taskGroup.Name)
+		} else {
+			tgs[taskGroup.Name] = taskGroup
+		}
+	}
+}
+
 func (app *Application) setTaskGroups(taskGroups []TaskGroup) {
 	app.lock.Lock()
 	defer app.lock.Unlock()
-	app.taskGroups = taskGroups
-	for _, taskGroup := range app.taskGroups {
-		app.placeholderAsk = common.Add(app.placeholderAsk, common.GetTGResource(taskGroup.MinResource, int64(taskGroup.MinMember)))
+	for _, taskGroup := range taskGroups {
+		// for duplicated task-group, will no longer be added to the app
+		if _, exists := app.taskGroups[taskGroup.Name]; !exists {
+			app.taskGroups[taskGroup.Name] = taskGroup
+			app.placeholderAsk = common.Add(app.placeholderAsk, common.GetTGResource(taskGroup.MinResource, int64(taskGroup.MinMember)))
+		}
 	}
 }
 
@@ -181,7 +202,15 @@ func (app *Application) getPlaceholderAsk() *si.Resource {
 func (app *Application) getTaskGroups() []TaskGroup {
 	app.lock.RLock()
 	defer app.lock.RUnlock()
-	return app.taskGroups
+
+	if len(app.taskGroups) > 0 {
+		taskGroups := make([]TaskGroup, 0, len(app.taskGroups))
+		for _, taskGroup := range app.taskGroups {
+			taskGroups = append(taskGroups, taskGroup)
+		}
+		return taskGroups
+	}
+	return nil
 }
 
 func (app *Application) setPlaceholderOwnerReferences(ref []metav1.OwnerReference) {
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -338,7 +338,7 @@ func (ctx *Context) ensureAppAndTaskCreated(pod *v1.Pod, app *Application) {
 		}
 		app = ctx.addApplication(&AddApplicationRequest{
 			Metadata: appMeta,
-		})
+		}, pod)
 	}
 
 	// get task metadata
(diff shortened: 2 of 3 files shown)


pkg/cache/application.go
```diff
@@ -80,7 +81,7 @@ func NewApplication(appID, queueName, user string, groups []string, tags map[str
 		taskMap:                 taskMap,
 		tags:                    tags,
 		sm:                      newAppState(),
-		taskGroups:              make([]TaskGroup, 0),
+		taskGroups:              taskGroups,
```
pbacsko, line 84: nit:
```
taskGroups: make(map[string]*TaskGroup]
```

pkg/cache/application.go
```diff
@@ -71,6 +71,7 @@ func (app *Application) String() string {
 
 func NewApplication(appID, queueName, user string, groups []string, tags map[string]string, scheduler api.SchedulerAPI) *Application {
 	taskMap := make(map[string]*Task)
+	taskGroups := make(map[string]*TaskGroup)
```
pbacsko, line 74: Does it have to be a pointer? Probably not.

pkg/cache/application.go
```diff
@@ -166,9 +167,16 @@ func (app *Application) GetSchedulingParamsDefinition() string {
 func (app *Application) setTaskGroups(taskGroups []TaskGroup) {
 	app.lock.Lock()
 	defer app.lock.Unlock()
-	app.taskGroups = taskGroups
-	for _, taskGroup := range app.taskGroups {
-		app.placeholderAsk = common.Add(app.placeholderAsk, common.GetTGResource(taskGroup.MinResource, int64(taskGroup.MinMember)))
+	for _, tg := range taskGroups {
+		taskGroup := tg
```
pbacsko, line 171: If the map is `string[TaskGroup]` then this is not needed I believe.

pkg/cache/application.go
```diff
@@ -166,9 +167,16 @@ func (app *Application) GetSchedulingParamsDefinition() string {
...
+		if _, exists := app.taskGroups[taskGroup.Name]; exists {
+			log.Log(log.ShimCacheApplication).Warn("duplicate task-group within the task-groups",
+				zap.String("appID", app.applicationID),
+				zap.String("groupName", taskGroup.Name))
```
pbacsko, line 175: Make sure you have an unit test which covers this branch.

pkg/cache/application.go
```diff
@@ -181,7 +189,14 @@ func (app *Application) getPlaceholderAsk() *si.Resource {
...
 	defer app.lock.RUnlock()
-	return app.taskGroups
+
+	taskGroups := make([]TaskGroup, 0)
+	if len(app.taskGroups) > 0 {
+		for _, taskGroup := range app.taskGroups {
+			taskGroups = append(taskGroups, *taskGroup)
+		}
+	}
+	return taskGroups
```
pbacsko, line 199: nit:
```
if len(app.taskGroups) > 0 {
  taskGroups := make([]TaskGroup, 0, len(app.taskGroups))
  for _, taskGroup := range app.taskGroups {
    taskGroups := append(taskGroups , *taskGroup) 
  }
  return taskGroups
}
return nil


```

pbacsko (changes requested):
Some nits

wilfred-s:
We cannot just eat the duplicates and not provide feedback to the customer via an event or something on the original pod.
See the comment in YUNIKORN-2517. Before making any code changes we need to define how we want to behave. 

Without specifying and documenting the behaviour I am against making this change.

wilfred-s (changes requested):
-1 without properly defining behaviour firts.

pkg/cache/application.go
XbaoWu: Thank you for your reply, I agree with you.

XbaoWu: > Does it have to be a pointer? Probably not.

Thank you for your reply, I agree with you.

pkg/cache/application.go
XbaoWu: > If the map is `string[TaskGroup]` then this is not needed I believe.

Yes, if the map is `string[TaskGroup]`, it really doesn 't need to be done here.

pkg/cache/application.go
XbaoWu: Well yeah, that 's a good addition. I will provide the unit test which covers this branch.

XbaoWu: > Make sure you have an unit test which covers this branch.

Well yeah, that 's a good addition. I will provide the unit test which covers this branch.

XbaoWu:
> We cannot just eat the duplicates and not provide feedback to the customer via an event or something on the original pod. See the comment in YUNIKORN-2517. Before making any code changes we need to define how we want to behave.
> 
> Without specifying and documenting the behaviour I am against making this change.
Thank you for your reply. In PR, I did ignore the description of our behavior in the document, which is really important. I will add this part of the description in the document and feedback to the submitter via an event on the original pod.

XbaoWu:
> We cannot just eat the duplicates and not provide feedback to the customer via an event or something on the original pod. See the comment in YUNIKORN-2517. Before making any code changes we need to define how we want to behave.
> 
> Without specifying and documenting the behaviour I am against making this change.
Thank you for your reply. In PR, I did ignore the description of our behavior in the document, which is really important. I will add this part of the description in the document and feedback to the submitter via an event on the original pod.

pbacsko:
> We cannot just eat the duplicates and not provide feedback to the customer via an event or something on the original pod. See the comment in YUNIKORN-2517. Before making any code changes we need to define how we want to behave.
> 
> Without specifying and documenting the behaviour I am against making this change.

+1 I forgot to say this. And also let's add some info on the docs site about how this is intended to work and warn the users that they should pay attention to the task group definition. We might even reject pods with incorrect taskgroup data.

XbaoWu:
According to YUNIKORN-2933 discussion, always taking the first task-group doesn 't seem like a good choice, so close this PR.

PR #929, opened 2024-10-18
[YUNIKORN-2879] [shim] yunikorn unschedulable pods pending forever

zj619:
* task postfail or rejected reschedule

### What is this PR for?
when task fail, yunikorn unschedulable pods pending forever,introduce retry

### What type of PR is it?
* [ ] - Bug Fix
* [√ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
Issue associated with this PR：https://issues.apache.org/jira/browse/YUNIKORN-2879

test cases
task fail to retry
image
after retry, bind successful
image

### Questions:

### DIFF
--- a/pkg/cache/task.go
+++ b/pkg/cache/task.go
@@ -40,17 +40,20 @@ import (
 )
 
 type Task struct {
-	taskID        string
-	alias         string
-	applicationID string
-	application   *Application
-	podStatus     v1.PodStatus // pod status, maintained separately for efficiency reasons
-	context       *Context
-	createTime    time.Time
-	placeholder   bool
-	originator    bool
-	sm            *fsm.FSM
-
+	taskID            string
+	alias             string
+	applicationID     string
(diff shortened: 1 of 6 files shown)


craigcondit:
I'm -1 on this whole approach. It's far too complex and introduces unnecessary per-pod configuration. We're trying to simplify the shim these days and this drastically complicates it.

pbacsko (changes requested):
-1 also.

@zhuqi-lucas is already working on sth like this, see [YUNIKORN-2804](https://issues.apache.org/jira/browse/YUNIKORN-2804).

This change alone is not even enough, we need to notify the core to cancel the allocation first and then the whole allocation cycle can restart again. We don't want to see an un-bindable pod on the core side to occupy resources.

PR #935, opened 2024-11-05
[YUNIKORN-2963] E2E Test for Foreign Pod Tracking

rrajesh-cloudera:
### What is this PR for?
This pull request introduces several end-to-end tests for verifying the tracking of foreign pods in the YuniKorn scheduler. The tests cover the following scenarios:
1. Verify foreign pod tracking fails for untracked or incorrectly mapped pods
2. Verify multiple foreign pods
3. Verify foreign pods on different nodes

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2963

### Questions:

### DIFF
--- a/test/e2e/foreign_pod/foreign_pod_test.go
+++ b/test/e2e/foreign_pod/foreign_pod_test.go
@@ -71,6 +71,119 @@ var _ = Describe("", func() {
 			Ω(foreignNodes[uid]).To(Equal(kubeNodes[uid]), "pod %s is tracked under incorrect node", uid)
 		}
 	})
+	It("Verify foreign pod tracking fails for untracked or incorrectly mapped pods", func() {
+		By("Retrieving foreign pods from kube-system")
+		kClient = k8s.KubeCtl{}
+		Ω(kClient.SetClient()).To(BeNil())
+		podList, err := kClient.GetPods(kubeSystem)
+		Ω(err).NotTo(gomega.HaveOccurred())
+
+		kubeUIDs := make(map[string]bool)
+		kubeNodes := make(map[string]string)
+		for _, pod := range podList.Items {
+			kubeUIDs[string(pod.UID)] = true
+			kubeNodes[string(pod.UID)] = pod.Spec.NodeName
+			fmt.Fprintf(ginkgo.GinkgoWriter, "pod: %s, uid: %s, node: %s\n", pod.Name, pod.UID, pod.Spec.NodeName)
+		}
+
+		// Simulate an error in foreign pod tracking
+		By("Simulating incorrect foreign allocation tracking")
+		var restClient yunikorn.RClient
+		nodes, err := restClient.GetNodes("default")
+		Ω(err).NotTo(gomega.HaveOccurred())
+
+		foreignAllocs := make(map[string]bool)
+		foreignNodes := make(map[string]string)
+		for _, n := range *nodes {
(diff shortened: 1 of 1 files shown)


rrajesh-cloudera:
Hi @pbacsko , Could you please take a look ?

pbacsko:
@rrajesh-cloudera I'm sorry to say but we already have e2e tests. Please see [tests](https://github.com/apache/yunikorn-k8shim/blob/master/test/e2e/foreign_pod/foreign_pod_test.go).

I'll take a look but I don't think we need more tests.

pbacsko (changes requested):
Ok. I've checked the tests. I don't see what extra they're performing besides the original `Verify foreign pod tracking` which I added 2 weeks ago. All tests just retrieve pods from `kube-system` and the assertions are slightly different, but that's it, it's basically the same in all.

Could you explain?

rrajesh-cloudera:
> @rrajesh-cloudera I'm sorry to say but we already have e2e tests. Please see [tests](https://github.com/apache/yunikorn-k8shim/blob/master/test/e2e/foreign_pod/foreign_pod_test.go). Looks like there was a communication issue :/
> 
> I'll take a look but I don't think we need more tests.
> 
> EDIT: oh, I can see you added the new tests to the same file. Anyway, I'm looking at them.

The First one is positive test case which you added , the 2nd is a negative test case for the first in which we are simulating an error where a pod from kube-system is either untracked or incorrectly mapped.
The Third case  Focuses on ensuring multiple pods from kube-system are all being tracked and 4th case Verifies that foreign pods are accurately tracked across different nodes.
Each case targets different potential issues, ensuring comprehensive coverage of the system’s functionality and robustness

test/e2e/foreign_pod/foreign_pod_test.go
```diff
@@ -71,6 +71,119 @@ var _ = Describe("", func() {
...
+		foreignNodes := make(map[string]string)
+		for _, n := range *nodes {
+			fmt.Fprintf(ginkgo.GinkgoWriter, "Checking node %s\n", n.NodeID)
+			if len(n.ForeignAllocations) > 0 {
+				for _, falloc := range n.ForeignAllocations {
+					// Introduce a mismatch to simulate an error
+					if string(falloc.AllocationKey) == "incorrect-uid" {
+						foreignAllocs[falloc.AllocationKey] = true
+						foreignNodes[falloc.AllocationKey] = "wrong-node"
+					} else {
```
pbacsko, line 105: I genuinely don't get this. I don't think this path is ever taken.

In all tests, you work with a fixed set of pods (allocations) from "kube-system". Those are system pods that are already running as they should be. You're not doing anything with those pods or with Yunikorn. The contents of `n.ForeignAllocations` will always be the same.

Unless there's some uncommitted/already-committed changes somewhere else which is not visible in the PR, this test (and the remaining ones) does not make any sense.

pbacsko:
> The First one is positive test case which you added , the 2nd is a negative test case for the first in which we are simulating an error where a pod from kube-system is either untracked or incorrectly mapped.
The Third case  Focuses on ensuring multiple pods from kube-system are all being tracked and 4th case Verifies that foreign pods are accurately tracked across different nodes.
Each case targets different potential issues, ensuring comprehensive coverage of the system’s functionality and robustness

This is just simply not the case.

PR #939, opened 2024-11-25
[YUNIKORN-2982] Send event when preemption occurs

rhh777:
### What is this PR for?
When a preemption occurs, send an event to be displayed in the victim's event list.

https://github.com/apache/yunikorn-core/pull/1000

### What type of PR is it?
* [x] - Improvement

### What is the Jira issue?
[YUNIKORN-2982](https://issues.apache.org/jira/browse/YUNIKORN-2982)

Trigger preemption


pbacsko (changes requested):
-1 for this approach.

If you use the `REQUEST` type as I explained in apache/yunikorn-core#1000, then the entire PR becomes unnecessary.

pbacsko:
@rhh777 do we still need this? Please close it if the changes in the core alone suffice.

rhh777:
close

PR #949, opened 2025-02-04
[YUNIKORN-2638] Simplify finalizeNodes and finalizePods

steinsgateted:
### What is this PR for?

- Change `map[types.UID]*v1.Pod` / `map[string]*v1.Node` to `map[types.UID]struct{}`.
Reduce the memory footprint of map, because struct{} is zero-size and does not occupy any additional space.

- Fixed map length to avoid reallocation of memory

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2638

### Questions:


chenyulin0719 (approved):
+1 LGTM

pkg/cache/context.go
```diff
@@ -1621,9 +1621,9 @@ func (ctx *Context) finalizeNodes(existingNodes []*v1.Node) error {
 	}
 
 	// convert the node list into a map
-	nodeMap := make(map[string]*v1.Node)
+	nodeMap := make(map[string]struct{}, len(nodes))
```
wilfred-s, line 1624: From a readability perspective I would prefer `nodeMap := make(map[string]bool, len(nodes))` and set `true` in 1626
Then at line 1631 the check simply becomes `if !nodeMap[node.Name] {`

Size (~5% more) and speed wise (not measurable unless the map has millions of entries) it does not make a difference.

chenyulin0719 (changes requested):
> From a readability perspective I would prefer nodeMap := make(map[string]bool, len(nodes)) and set true in 1626

I'm +1 on @wilfred-s's suggestion. I prefer readability over small performance enhancements too. @steinsgateted Could you make change for it?

pbacsko:
> I'm +1 on @wilfred-s's suggestion. I prefer readability over small performance enhancements too. @steinsgateted Could you make change for it?

Same here, I also vote for `map[string]bool`.

pbacsko:
@steinsgateted do you have time to finish this? If not, I can take it over and make the suggested change.

pkg/cache/context.go
craigcondit: Updated PR per suggestion.

pbacsko (approved):

PR #953, opened 2025-02-21
[YUNIKORN-2996] Build against Kubernetes 1.32.2

craigcondit:
### What is this PR for?
Update Kubernetes build dependency to v1.32.1.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2996

### Questions:


manirajv06:
looks good.

Can we look into code coverage warnings, focussing on the newly introduced methods?

pbacsko (approved):
LGTM

craigcondit:
> looks good.
> 
> Can we look into code coverage warnings, focussing on the newly introduced methods?

The only missing bits are mock code, stubs that should never be called, and a tiny bit of plugin init code that will always get hit during plugin mode startup in e2e tests.

PR #954, opened 2025-02-28
[YUNIKORN-3031] Add support for .spec.schedulingGates

pbacsko:
### What is this PR for?
Add support for `.spec.schedulingGates`. Pod is not scheduled as long as there is at least one gate defined.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3031

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -336,6 +336,7 @@ func (ctx *Context) updateYuniKornPod(appID string, oldPod, pod *v1.Pod) {
 		log.Log(log.ShimContext).Info("pod is waiting for scheduling gates", zap.String("name", pod.Name), zap.Strings("gates", gates))
 	}
 
+	// always call UpdatePod() first to make sure the pod instance is the latest in the cache
 	if ctx.schedulerCache.UpdatePod(pod) && !hasGates {
 		// pod was accepted; ensure the application and task objects have been created
 		ctx.ensureAppAndTaskCreated(pod, app)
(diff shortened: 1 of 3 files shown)


pkg/cache/context.go
```diff
@@ -317,7 +325,18 @@ func (ctx *Context) updateYuniKornPod(appID string, pod *v1.Pod) {
...
+			gates = append(gates, gate.Name)
+		}
+		events.GetRecorder().Eventf(pod.DeepCopy(), nil, v1.EventTypeNormal, "Scheduling", "Scheduling",
+			"waiting for scheduling gates: %s", strings.Join(gates, ","))
+		log.Log(log.ShimContext).Info("pod is waiting for scheduling gates", zap.String("name", pod.Name), zap.Strings("gates", gates))
+	}
+
+	if ctx.schedulerCache.UpdatePod(pod) && !hasGates {
 		// pod was accepted; ensure the application and task objects have been created
 		ctx.ensureAppAndTaskCreated(pod, app)
```
craigcondit, line 341: Might this be dangerous?. I could envision a case where the initiator pod of an app has scheduling gates set, but the other app pods do not. This could result in an unexpected pod being chosen as the application leader. Thoughts @pbacsko / @wilfred-s ?

wilfred-s: That would be really weird and broken setup. Even with the default scheduler the pods without the gates would be scheduled. If they all belong together you would get into a state that the main pod cannot be scheduled due to the gates but everything else is running on nodes. There is no way we can guarantee that it will not happen as we have no control over what the user does.
It is similar to the race issue we have seen when creating multiple placeholder pods. There is no specific order in which they come back to the scheduler. This is not different.

craigcondit (approved):
+1 LGTM.

PR #956, opened 2025-03-05
[YUNIKORN-2875] Trigger spark_py job on spark image to get rid of java version dependency

kaichiachen:
### What is this PR for?
If java version is 23, it will hit the blowing issue if run spark_submit because Java 23 will check SecurityManager capability and spark only support for 8, 11 and 17. See https://stackoverflow.com/questions/79016199/hadoop-installation-error-getsubject-is-supported-only-if-a-security-manager-i
```
25/03/03 20:40:30 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
25/03/03 20:40:30 INFO SparkKubernetesClientFactory: Auto-configuring K8S client using current context from users K8S config file
25/03/03 20:40:30 INFO KerberosConfDriverFeatureStep: You have not specified a krb5.conf file locally or via a ConfigMap. Make sure that you have the krb5.conf locally on the driver image.
25/03/03 20:40:30 INFO ShutdownHookManager: Shutdown hook called
25/03/03 20:40:30 INFO ShutdownHookManager: Deleting directory /private/var/folders/dt/cc9ffv8s2z5f7tr5hpcv3wvr0000gq/T/spark-6ebd3456-2aa8-41fd-9370-3b65b3a3584b
25/03/03 20:40:30 INFO ShutdownHookManager: Deleting directory /private/var/folders/dt/cc9ffv8s2z5f7tr5hpcv3wvr0000gq/T/spark-e9db98d6-c4be-4ae5-a173-b230e2103b94
```
Hence hit test error - spark queue name is not retrievable after 120 seconds
```
  Unexpected error:
      <context.deadlineExceededError>: 
      context deadline exceeded
      {}
  occurred
``` 

In this PR, I would like to trigger spark_py jobs from a spark pod to get rid of java dependency. The key point is how to update `spark_jobs.sh`
1. Deploy spark image with serviceAccountName(Get permission to deploy pods)
2. Copy template pod yaml to spark image
3. Deploy spark_py jobs from spark pod

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2875

Tests are all passed for `make kind-e2e` and `make e2e_test`

### Questions:


craigcondit (changes requested):
I'm -1 on this approach. Instead, we should find a way to make the test functional regardless of the java version installed on the host. Perhaps execute spark-submit inside the official Spark docker container?

chenyulin0719:
I think running spark-submit in an official Spark container is too cumbersome. We have to download the image for each test(In Github Action or any new environment). The total execution time depends on network condition. (Download + execute spark job)

To me, I'm +1 on checking java version(8, 11 or 17) before running sparlk-submit. It's good engough to print the reason for 
 Spark failure.

craigcondit:
> I think running spark-submit in an official Spark container is too cumbersome. We have to download the image for each test(In Github Action or any new environment). The total execution time depends on network condition. (Download + execute spark job)

Except that this already gets downloaded as part of running the test itself. We could simply create a starter pod using the same image that has an embedded spark-submit in it. It wouldn't be any more overhead than we already have.

craigcondit (approved):
+1  LGTM.

PR #957, opened 2025-03-05
[YUNIKORN-2850] Watch configmap only in yunikorn's deployed namespace

kousei47747:
### What is this PR for?
Currently, Yunikorn uses configmap informer to handle configuration hot reload.

However, In current implementation informer watches all namespaces even only need to watch namespace in which yunikorn is deployed. It causes in efficient behavior when sync and cache configmap states. If there is too many unrelated configmap in other namespace cause long recovery time to list and memory presure to handle configmap caches which is redundant.

So, If we could replace configmap informer to namespace restricted one, it would improve startup / recovery time and reduce memory usage.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2850


craigcondit (approved):
+1 LGTM.

PR #966, opened 2025-04-14
[YUNIKORN-2981] Support filtering schedulerName

kaichiachen:
### What is this PR for?
Implements scheduler name filtering in Yunikorn's admission controller to support environments with multiple schedulers 

Key changes:
- Added new configuration options:
  - `admissionController.filtering.overrideCustomSchedulers` 
- Modified admission controller logic to:
  - update pod scheduler to `yunikorn` forcely if `admissionController.filtering.overrideCustomSchedulers` is true or pod scheduler name is `default-scheduler`

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2981

1. Set up Yunikonr with
```
admissionController:
  filtering:
    overrideCustomSchedulers: true
```
2. Set up default-scheduler pods  - Verify pod scheduler turn to yunikorn
3. Set up pods with other scheduler name and turn on overrideCustomSchedulers - Verify pod scheduler turn to yunikorn
4. Set up pods with other scheduler name and turn off overrideCustomSchedulers - Verify pod scheduler wont change

### Questions:

### DIFF
--- a/pkg/admission/admission_controller.go
+++ b/pkg/admission/admission_controller.go
@@ -194,7 +194,14 @@ func (c *AdmissionController) processPod(req *admissionv1.AdmissionRequest, name
 		log.Log(log.Admission).Info("bypassing namespace", zap.String("namespace", namespace))
 		return admissionResponseBuilder(uid, true, "", nil)
 	}
-	patch = updateSchedulerName(patch)
+
+	// Only update scheduler name if it matches process list and doesn't match bypass list
+	if c.shouldProcessScheduler(pod.Spec.SchedulerName) {
+		patch = updateSchedulerName(patch)
+	} else {
+		log.Log(log.Admission).Info("bypassing scheduler name update",
+			zap.String("schedulerName", pod.Spec.SchedulerName))
+	}
 
 	if c.shouldLabelNamespace(namespace) {
 		patch = c.updateLabels(namespace, &pod, patch)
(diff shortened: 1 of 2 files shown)


wilfred-s:
We should only ever override the `default-scheduler` not any custom scheduler that has been set on a pod by default. If we have a custom scheduler we should let the pod pass and not make any changes. That check should be made *before* we do anything else.
We can then add an override of custom scheduler option. We can use that if we want to force everything in the cluster to be scheduled by YuniKorn based on namespaces etc. Adding the option to override some but not all of the schedulers passed in, is I think not going to give the correct user experience. 

So in processPod before checking user info:
* check the schedulername  is not `default-scheduler` or `yunikorn`
* check if custom scheduler override is set to `false`: return admission response to allow the pod without any changes

pkg/admission/admission_controller.go
```diff
@@ -194,7 +194,16 @@ func (c *AdmissionController) processPod(req *admissionv1.AdmissionRequest, name
...
+
+	// Only override default-scheduler unless override is enabled
+	if pod.Spec.SchedulerName == "default-scheduler" ||
+		(c.conf.GetOverrideCustomSchedulers() && pod.Spec.SchedulerName != constants.SchedulerName) {
+		patch = updateSchedulerName(patch)
+	} else {
+		log.Log(log.Admission).Info("bypassing scheduler name update",
+			zap.String("schedulerName", pod.Spec.SchedulerName),
+			zap.Bool("overrideCustomSchedulers", c.conf.GetOverrideCustomSchedulers()))
+	}
```
chenyulin0719, line 206: It seems that all pods without changing scheduler name are printing this log, which might not be necessary.

We can change it to only print the log when the scheduler name is actually modified.
Or perhaps we can remove the logging altogether, since we will log the patch later anyway

chenyulin0719 (commented):
Hi @kaichiachen, thanks for submitting the PR. One comment from me.

pkg/admission/admission_controller.go
kaichiachen: Fair enough. Only logs when updating scheduler name

wilfred-s (changes requested):
Two minor issues and then we're good to go.

pkg/admission/admission_controller.go
```diff
@@ -194,7 +194,16 @@ func (c *AdmissionController) processPod(req *admissionv1.AdmissionRequest, name
 		log.Log(log.Admission).Info("bypassing namespace", zap.String("namespace", namespace))
 		return admissionResponseBuilder(uid, true, "", nil)
 	}
-	patch = updateSchedulerName(patch)
+
+	// Only override default-scheduler unless override is enabled
+	if c.conf.GetOverrideCustomSchedulers() && pod.Spec.SchedulerName != constants.SchedulerName {
```
wilfred-s: See earlier comments: this is not correct, default behaviour should not change. can be turned off if wanted. 2nd part of the check is not needed as we preserve existing behaviour

wilfred-s (changes requested):

PR #978, opened 2025-07-18
[YUNIKORN-3103] Add comprehensive e2e replication test suite for YuniKorn scheduler

rrajesh-cloudera:
### What is this PR for?
- Add 13 comprehensive test cases covering various replication scenarios
- Test ReplicaSets, Deployments, StatefulSets, multi-container pods, node affinity, rolling updates, priority classes, mixed workloads, and stress testing
- Validate YuniKorn scheduler integration and resource allocation
- Include performance optimizations for faster CI/CD execution
- Add DeleteService method to k8s utils for test cleanup
- Optimize timeouts and resource requests for 60% faster execution
- Use parallel pod waiting and single lightweight image for efficiency

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [x] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3103

### Questions:


rrajesh-cloudera:
Could you please review the PR ?
@pbacsko @manirajv06

test/e2e/replication/replication_test.go
```diff
@@ -0,0 +1,1469 @@
...
+
+		ginkgo.By("Waiting for replica pods to be created and running")
+		var podList *v1.PodList
+		gomega.Eventually(func() bool {
+			podList, err = kClient.ListPods(dev, "app=replica-test")
+			if err != nil {
+				return false
+			}
+			return len(podList.Items) == int(replicaCount)
+		}, standardTimeout, pollInterval).Should(gomega.BeTrue(), "Should have 2 replica pods")
```
pbacsko, line 187: This code fragment is repeated a lot. We need helper functions inside `KubeCtl`, similar to `WaitForPodBySelectorRunning()`. We can create two methods:

`WaitForPodCountBySelectorRunning()` - wait for the number of pods + all of them should be in running state
`WaitForPodCountBySelector()` - wait for the number of pods (they don't have to be in Running state)

This way, a lot of code duplication inside `gomegy.Eventually()` can be removed and we can just switch to the appropriate `KubeCtl` method (which is what we do anyway).

I haven't tested these but this is a good starting point:
```
func (k *KubeCtl) WaitForPodCountBySelectorRunning(namespace string, selector string, count, timeout int) error {
	var podList *v1.PodList
	var err error

	err = wait.PollUntilContextTimeout(context.TODO(), time.Second, time.Duration(timeout)*time.Second, false,
		func(ctx context.Context) (done bool, err error) {
			podList, err = k.ListPods(namespace, selector)
			if err != nil {
				return false, err
			}
			return len(podList.Items) == count, nil
		})
	if err != nil {
		return err
	}
	
	for _, pod := range podList.Items {
		if err := k.WaitForPodRunning(namespace, pod.Name, time.Duration(timeout)*time.Second); err != nil {
			return err
		}
	}
	return nil
}

func (k *KubeCtl) WaitForPodCountBySelector(namespace string, selector string, count, timeout int) error {
	err := wait.PollUntilContextTimeout(context.TODO(), time.Second, time.Duration(timeout)*time.Second, false,
		func(ctx context.Context) (done bool, err error) {
			podList, err := k.ListPods(namespace, selector)
			if err != nil {
				return false, err
			}
			return len(podList.Items) == count, nil
		})
	if err != nil {
		return err
	}

	return nil
}
```

If you also need the podList, then just return it along with the error.

test/e2e/replication/replication_test.go
```diff
@@ -0,0 +1,1469 @@
...
+				return false
+			}
+			// Check all pods are running
+			for _, pod := range podList.Items {
+				if pod.Status.Phase != v1.PodRunning {
+					return false
+				}
+			}
+			return true
+		}, standardTimeout, pollInterval).Should(gomega.BeTrue(), "Should have 3 scaled running pods")
```
pbacsko, line 307: Replace with the suggested `KubeCtl.WaitFor...()` helper function

test/e2e/replication/replication_test.go
```diff
@@ -0,0 +1,1469 @@
...
+				return false
+			}
+			// Check all pods are running
+			for _, pod := range podList.Items {
+				if pod.Status.Phase != v1.PodRunning {
+					return false
+				}
+			}
+			return true
+		}, standardTimeout, pollInterval).Should(gomega.BeTrue(), "Should have 2 initial running pods")
```
pbacsko, line 384: Replace with the suggested `KubeCtl.WaitFor...()` helper function

test/e2e/replication/replication_test.go
```diff
@@ -0,0 +1,1469 @@
...
+				return false
+			}
+			// Check all pods are running
+			for _, pod := range podList.Items {
+				if pod.Status.Phase != v1.PodRunning {
+					return false
+				}
+			}
+			return true
+		}, standardTimeout, pollInterval).Should(gomega.BeTrue(), "Should have 2 pods after recovery")
```
pbacsko, line 407: Replace with the suggested helper function

test/e2e/replication/replication_test.go
```diff
@@ -0,0 +1,1469 @@
...
+										v1.ResourceCPU:    resource.MustParse("50m"),  // Reduced from 100m
+										v1.ResourceMemory: resource.MustParse("64Mi"), // Reduced from 128Mi
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		}
```
pbacsko, line 461: Replace with a replicaSet creator helper function.

test/e2e/replication/replication_test.go
```diff
@@ -0,0 +1,1469 @@
...
+										v1.ResourceCPU:    resource.MustParse("10m"),
+										v1.ResourceMemory: resource.MustParse("16Mi"),
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		}
```
pbacsko, line 546: Replace with a replicaSet creator helper function.

test/e2e/replication/replication_test.go
```diff
@@ -0,0 +1,1469 @@
...
+										v1.ResourceCPU:    resource.MustParse("10m"),  // Reduced from 25m
+										v1.ResourceMemory: resource.MustParse("16Mi"), // Reduced from 32Mi
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		}
```
pbacsko, line 661: Replace with a replicaSet creator helper function.

test/e2e/replication/replication_test.go
```diff
@@ -0,0 +1,1469 @@
...
+					return false
+				}
+				for _, containerStatus := range pod.Status.ContainerStatuses {
+					if !containerStatus.Ready {
+						return false
+					}
+				}
+			}
+			return true
+		}, longTimeout, pollInterval).Should(gomega.BeTrue(), "All multi-container pods should be ready")
```
pbacsko, line 700: Ok, this probably needs a different helper function than the previous ones. Regardless, this logic should not be here, but inside `KubeCtl`.

test/e2e/replication/replication_test.go
```diff
@@ -0,0 +1,1469 @@
...
+										v1.ResourceCPU:    resource.MustParse("10m"),
+										v1.ResourceMemory: resource.MustParse("16Mi"),
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		}
```
pbacsko, line 800: Replace with a replicaSet creator helper function.

test/e2e/replication/replication_test.go
```diff
@@ -0,0 +1,1469 @@
...
+										v1.ResourceCPU:    resource.MustParse("25m"),  // Reduced from 50m
+										v1.ResourceMemory: resource.MustParse("32Mi"), // Reduced from 64Mi
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		}
```
pbacsko, line 1114: Call replicaSet creator function.

test/e2e/replication/replication_test.go
```diff
@@ -0,0 +1,1469 @@
...
```
pbacsko, line 1201: Replace with a replicaSet creator function call.

pbacsko (changes requested):
Hey @rrajesh-cloudera thanks for the PR. My only concern is that the test logic is VERY LONG in the current form and it's difficult to read.
At least let's get rid of the most commonly repeated parts to reduce the size of the tests.

Please look at my suggestions:
1. Create helper methods which return a basic replica set, deployment and stateful set objects with all common fields populated.
2. If you need changes, just call the method ie. `getReplicaSetSpec()` then modify the fields in the testcase code.
3. Replace all `gomega.Eventually()` call with appropriate helper calls from `KubeCtl`. These will be new methods.

pbacsko (changes requested):
I added some further comments.

I don't think we need all tests here, especially the "nodeAffinity" and "stress" stuff. These can become complicated very quickly.
I also don't see tuch much value in the following: `Verify_Mixed_Workload_Replication`, `Verify_Replication_With_Priority_Classes`.

The checks inside `Verify_Resource_Allocation_Across_Replicas` and `Verify_Queue_Assignment_Consistency` can be merged to 
`Verify_ReplicaSet_Pod_Replication`.

The rest can stay. By making these changes, code will be further reduced, this making it a smaller maintenance burden.

PR #995, opened 2025-12-12
[YUNIKORN-3128] Rollback allocation on bind failure

jeffgao01:
### What is this PR for?

This PR is regarding https://issues.apache.org/jira/browse/YUNIKORN-3128

* Define BindFailed taskEventType
* Rollback allocation on bind-volumes-to-pod or bind-pod-to-node failure
* Modify the task state to new, for follow-up retry

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-3128

* unit test

### Questions:

### DIFF
--- a/pkg/cache/task.go
+++ b/pkg/cache/task.go
@@ -375,8 +375,13 @@ func (task *Task) postTaskAllocated() {
 				zap.String("podName", task.pod.Name),
 				zap.String("podUID", string(task.pod.UID)))
 			if err := task.context.bindPodVolumes(task.pod); err != nil {
-				log.Log(log.ShimCacheTask).Error("bind volumes to pod failed", zap.String("taskID", task.taskID), zap.Error(err))
-				task.failWithEvent(fmt.Sprintf("bind volumes to pod failed, name: %s, %s", task.alias, err.Error()), "PodVolumesBindFailure")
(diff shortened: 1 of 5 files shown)


pkg/cache/task.go
```diff
@@ -424,6 +434,18 @@ func (task *Task) beforeTaskAllocated(eventSrc string, allocationKey string, nod
 	}
 }
 
+func (task *Task) beforeTaskBindFail() {
+	task.releaseAllocation()
```
jeffgao01: > Looking at the change I think we're missing the part that needs to communicate the failure back to the core. The core already has assigned a node to the allocation. That also needs to be undone. Without that the core still thinks that the allocation is assigned to a node and will never retry it.

Actually, the current change is rolling back the allocation as a whole. 
But I synced with Wilfred, a more fine-grained approach is to partially rollback the allocation, to the extent that it can be retried (but the existing metadata/history is retained).

wilfred-s:
old pull request, new approach will be taken to fix this.

PR #1010, opened 2026-04-06
flag to disable k8s events

drewbailey:
Adds a flag to disable emitting kubernetes events. In environments under significant load, yunikorn can contribute to emitting millions of events which takes a toll on etcd.

Use it by setting:

service.disableKubernetesEvents: "true"

### What is this PR for?
A few sentences describing the overall goals of the pull request's commits.
First time? Check out the contributing guide - http://yunikorn.apache.org/community/how_to_contribute   

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:

### DIFF
--- a/pkg/common/events/recorder.go
+++ b/pkg/common/events/recorder.go
@@ -27,8 +27,7 @@ import (
 var eventRecorder atomic.Pointer[events.EventRecorder]
 
 func init() {
-	r := events.EventRecorder(NewMockedRecorder())
-	eventRecorder.Store(&r)
+	UseMockedRecorder()
 }
 
 func GetRecorder() events.EventRecorder {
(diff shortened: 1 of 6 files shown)


wilfred-s:
> The event types vary, failedmount, backoff, failedcreatepodsandbox. We are not relying on these events for debugging so having the option to disable them for performance is ideal for us. Rate limiting does seem like a good option for others

`failedmount` and `failedcreatepodsandbox` are kubelet events turning off the scheduler events does not prevent those from getting created.  
I have tried to find events that mention `backoff` or `back off` in our code but I have not been able to find anything. The generic K8s `backoff` event is linked to failing or crashing pods and not related to scheduling at all. Turning off scheduler events will not stop those from getting generated.
We can generate 27 events. Most of those are one off and can not be generated more than once for any pod. Example `PodBindSuccessful` when a pod is bound to a node. 

One set of events is not generated by the k8shim and is more "chatty" That is where I think the issue is located. In the core we have an event system which is hooked up to the k8shim for publishing for certain events. That could generate a much larger load of events most of them linked to quotas.
Can you try and turn off the event system by setting `event.trackingEnabled: "false"` in the config map?
See https://yunikorn.apache.org/docs/user_guide/service_config#eventtrackingenabled

pbacsko:
@drewbailey please give us the output of `kubectl events` or `kubectl describe pod` so we can see what events have been generated by Yunikorn, if any. As Wilfred pointed out, the only event-related issue has been observed with regards to unschedulable pods. 

Backoff: we recently introduced the application backoff feature, but that does not send any event, it just logs a single line to the console.

PR #1036, opened 2026-06-01
[YUNIKORN-3283] CVE fixes for golang.org/x/* packages

adityadtu5:
### What is this PR for?
CVE fixes for golang.org/x/* packages   

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3283

### Questions:


go.mod
```diff
@@ -104,17 +104,17 @@ require (
 	go.opentelemetry.io/otel/trace v1.43.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	go.yaml.in/yaml/v2 v2.4.3 // indirect
-	golang.org/x/crypto v0.49.0 // indirect
+	golang.org/x/crypto v0.51.0 // indirect
```
wilfred-s, line 107: This version is still vulnerable and has an override in the core also to update it. Override is missing here.

go.mod
```diff
@@ -138,8 +138,8 @@ require (
 )
 
 replace (
-	golang.org/x/mod => golang.org/x/mod v0.34.0
-	golang.org/x/net => golang.org/x/net v0.52.0
+	golang.org/x/mod => golang.org/x/mod v0.36.0
+	golang.org/x/net => golang.org/x/net v0.54.0
```
wilfred-s: incorrect override should be 0.55

wilfred-s (changes requested):

go.mod
adityadtu5: Added a replace for this with version v0.52.0. However note that require will still shows the minimum version which the direct module requires. However, during compilation, replace will override.

go.mod
adityadtu5: done, thanks

wilfred-s (approved):
LGTM

PR #1039, opened 2026-06-08
[YUNIKORN-3295] K8s 1.36 API changes in Shim

manirajv06:
### What is this PR for?

K8s 1.36.1 API changes in Shim
- DRA Manager has been used as is without any manipulations on top of it. So, config, resource claims, resource claim template, drivers settings etc at K8s level would be used as is.
- Updated Go version to 1.26.0 in go.mod

Used Informers `AddEventHandlersWithResyncPeriod` instead of `AddEventHandler` (without "resyncPeriod" option) as we were passing '0' which doesn't have any effect on resync process anyway
Used Streaming K8s API instead of API Machinery as it was deprecated
Used component-helpers utils instead of resource helper as it was deprecated
Refactored existing unit test mocks

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3295

### Questions:


wilfred-s:
> Pre-commit checks / e2e-tests (v1.33.7) (pull_request) Failing after 44m

The change is build on top of a commit that has not got #1040, besides that no test failures.

test/e2e/framework/helpers/k8s/k8s_utils.go
```diff
@@ -32,6 +32,8 @@ import (
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/client-go/transport/spdy"
+	"k8s.io/streaming/pkg/httpstream"
```
wilfred-s, line 36: filed a follow up jira for fixing imports in e2e tests: [YUNIKORN-3301](https://issues.apache.org/jira/browse/YUNIKORN-3301)

wilfred-s (approved):
LGTM +1

PR #1040, opened 2026-06-10
YUNIKORN-3298: Stabilize predicates cleanup health checks

venkateshwaracholan:
## What is this PR for?

Stabilize the predicates e2e cleanup path for YUNIKORN-3298.

The failures occur in `AfterEach` after namespace teardown:

```text
Name: Negative resources
Message: Nodes with negative resources: ["yk8s-worker"]
```

We don't yet know whether this is a transient reconciliation delay or an underlying scheduler issue. This change makes the cleanup path wait for scheduler health to recover instead of assuming it is immediately stable.

## What does this PR do?

* Wait for helper pods to fully terminate after deletion.
* Replace the immediate post-cleanup health check with polling that waits for:

  * non-negative node resources, and
  * scheduler health checks to pass.
* Continue to fail if health does not recover within 60 seconds.

## How should this be tested?

```bash
go test ./test/e2e/framework/helpers/yunikorn/... -count=10 -v
```

```bash
make kind-e2e E2E_TEST="--label-filter=TestPredicates"
```

Added tests:

* TestHasNegativeNodeResources
* TestWaitForSchedulerHealth
* TestWaitForSchedulerHealthNegativeNodes
* TestWaitForSchedulerHealthTimeout

Helper tests: Passed (10/10)

## What type of PR is this?

* Test Stabilization

## Jira

https://issues.apache.org/jira/browse/YUNIKORN-3298

### Questions:

### DIFF
--- a/test/e2e/framework/helpers/yunikorn/rest_api_utils.go
+++ b/test/e2e/framework/helpers/yunikorn/rest_api_utils.go
@@ -27,6 +27,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


wilfred-s:
kicked off the workflow, change looks good at first view.

wilfred-s (approved):
LGTM

pbacsko (approved):

PR #1043, opened 2026-06-19
[YUNIKORN-3300] Use feasible nodes returned from K8s PreFilter plugin…

manirajv06:
### Description

Implemented `Predicates Pre Filter` API in Predicate Manager to run the Predicate checks using K8s Pre filter plugins and return the feasible nodes with an error (if any) for the given pod. A new cycle state would be initiated for every pod and cached in `SchedulerCache` at the end as part of this API. An important info stored in cycle state is  Skipped Plugins. If K8s Prefilter returns Plugin status as `Skip`, then the same info would be persisted in Cycle state. When the same pod goes through `Predicates Filter` API later, cycle state would be fetched from the cache and used accordingly. If any plugin skipped earlier during pre filtering, then the same plugin should be skipped now during filtering as well. Cycle state stored in `SchedulerCache` would be cleared as part of pod removal process.

Modified existing tests based on these changes. Added more new tests as well.

### Type of change

- [ ] Bug Fix
- [X] Improvement
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3300

- [ ] The Jira ID is part of the title of this pull request.

### AI Tooling
If an AI tool was used:

### How has this been tested?
- [X] New unit tests were added to cover new or changed code paths.

### Questions:


pkg/plugin/predicates/predicate_manager.go
```diff
@@ -187,80 +169,87 @@ func (p *predicateManagerImpl) removePodFromNodeNoFail(node fwk.NodeInfo, pod *v
...
+			return fwk.AsStatus(errors.Join(fmt.Errorf("running PreFilter plugin %q: ", plugin), err)), plugin, map[string]struct{}{}
 		}
-		// Merge is nil safe and returns a new PreFilterResult result if mergedNodes was nil
-		mergedNodes = mergedNodes.Merge(nodes)
-		if !mergedNodes.AllNodes() && !mergedNodes.NodeNames.Has(node.Node().Name) {
-			return fwk.NewStatus(fwk.UnschedulableAndUnresolvable, "node not eligible"), plugin, skip
+		if nodes != nil {
+			for n := range nodes.NodeNames {
+				feasibleNodes[n] = struct{}{}
+			}
```
pbacsko: Merging logic looks incorrect.

In Kubernetes, PreFilterResult.Merge() intersects node sets. With union, if plugin A returns {n1,n2} and plugin B returns {n2,n3}, now it yields {n1,n2,n3} instead of {n2}.

If I'm right, then make sure that this is covered properly with an unit test.

pkg/cache/context.go
```diff
@@ -682,6 +683,28 @@ func (ctx *Context) EventsToRegister(queueingHintFn fwk.QueueingHintFn) []fwk.Cl
...
+	pod := ctx.schedulerCache.GetPod(name)
+	if pod == nil {
+		return map[string]struct{}{}, ErrorPodNotFound
+	}
+	// if pod exists in cache, try to run predicates
+	// need to lock cache here as predicates need a stable view into the cache
+	ctx.schedulerCache.LockForReads()
+	plugin, feasibleNodes, cycleState, err := ctx.predManager.PreFilter(pod, allocate)
+	ctx.schedulerCache.UnlockForReads()
+	ctx.schedulerCache.UpdateCycleState(pod, cycleState)
```
pbacsko: I think this should be collapsed under single write lock - that's better from the point of view of concurrency.

pkg/cache/context.go
```diff
@@ -682,6 +683,28 @@ func (ctx *Context) EventsToRegister(queueingHintFn fwk.QueueingHintFn) []fwk.Cl
 	return ctx.predManager.EventsToRegister(queueingHintFn)
 }
 
+// Prefilter evaluates given prefilter based predicates based on current context
+func (ctx *Context) Prefilter(name string, allocate bool) (map[string]struct{}, error) {
```
pbacsko: nit: `Prefilter` or `PreFilter` ?

pbacsko (changes requested):
found some minor things

pbacsko (changes requested):
@manirajv06 I asked a thorough review from Opus 4.8 (it went for 25 minutes) and it identified bugs and some less critical items to fix. Pay close attention to them, especially no#1. I believe no#4 about `go.mod` can be ignored. The review is about both shim & core changes.

[YUNIKORN-3300 Claude Opus 4.8 review.pdf](https://github.com/user-attachments/files/29718841/YUNIKORN-3300.Claude.Opus.4.8.review.pdf)

pkg/cache/context.go
manirajv06: Taken care

pkg/cache/context.go
manirajv06: Taken care.

But wondering why predicates manager requires stable view of the cache? Need to take a deep look at this exisitng comment

`// need to lock cache here as predicates need a stable view into the cache`

pkg/plugin/predicates/predicate_manager.go
```diff
@@ -187,80 +171,92 @@ func (p *predicateManagerImpl) removePodFromNodeNoFail(node fwk.NodeInfo, pod *v
...
 				zap.String("pluginName", plugin),
 				zap.String("pod", fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)),
 				zap.Error(err))
-			return fwk.AsStatus(errors.Join(fmt.Errorf("running PreFilter plugin %q: ", plugin), err)), plugin, skip
+			return fwk.AsStatus(errors.Join(fmt.Errorf("running PreFilter plugin %q: ", plugin), err)), map[string]*si.Empty{}
+		} else {
+			if mergedPreFilterResults == nil {
+				mergedPreFilterResults = &fwk.PreFilterResult{}
+			}
+			mergedPreFilterResults.Merge(nodes)
```
pbacsko: BUG:  `Merge()` has a return value which is discarded. It is a pure function which does not mutate anything.

Simply:
```
mergedPreFilterResults = mergedPreFilterResults.Merge(nodes)
```

pkg/cache/context.go
```diff
@@ -706,9 +739,14 @@ func (ctx *Context) IsPodFitNode(name, node string, allocate bool) error {
...
-	defer ctx.schedulerCache.UnlockForReads()
-	plugin, err := ctx.predManager.Predicates(pod, targetNode, allocate)
+	ctx.schedulerCache.LockForWrites()
+	defer ctx.schedulerCache.UnlockForWrites()
+	cycleState := ctx.schedulerCache.GetCycleState(pod)
+	if cycleState == nil {
+		return ErrorCycleStateNotFound
+	}
+	plugin, err := ctx.predManager.Filter(pod, targetNode, cycleState, allocate)
+	ctx.schedulerCache.DeleteCycleState(pod)
```
pbacsko: BUG: the real flow is one `PreFilter()`, then `Filter()` once per candidate node (core's `tryNodes()` calls `preAllocateConditions()` once, then `tryNode()` -> `Predicates()` per node). After the first node's `Filter()` the state is gone, so every subsequent node returns `ErrorCycleStateNotFound`, which core treats as a predicate failure.

pkg/cache/context.go
```diff
@@ -706,9 +739,14 @@ func (ctx *Context) IsPodFitNode(name, node string, allocate bool) error {
 		return ErrorNodeNotFound
 	}
 	// need to lock cache here as predicates need a stable view into the cache
-	ctx.schedulerCache.LockForReads()
-	defer ctx.schedulerCache.UnlockForReads()
-	plugin, err := ctx.predManager.Predicates(pod, targetNode, allocate)
+	ctx.schedulerCache.LockForWrites()
+	defer ctx.schedulerCache.UnlockForWrites()
```
pbacsko: I think removing `DeleteCycleState()` is sufficient. So these can go back to read locks.

pbacsko (changes requested):
This PR still looks problematic.

Another thing to point out that now since `PreFilter()` is a distinct, separate phase, we lose some information when a pod is unschedulable. We no longer print the plugin name, only a generic diagnostic message.

This is because `si.PreFilterPredicatesResponse` only carries feasible nodes and `Success`. 

I'd be happier with a more proper message instead of just:

```
a.LogAllocationFailure(common.ErrorPreFilterPredicate.Error(), allocate)          // "prefilter predicates failed"
podPredicateErrors[common.ErrorPreFilterPredicate.Error()]++
a.SendPredicatesFailedEvent(podPredicateErrors)
```

In fact, this might be just more confusing to YK users ("what the heck is prefilter?"). Plugin name carries extra information, displaying "prefilter" is only valuable to those who understand scheduling internals (btw the current diagnostics message could be improved as well to better match those produced by the default scheduler).

manirajv06:
> 
> In fact, this might be just more confusing to YK users ("what the heck is prefilter?"). Plugin name carries extra information, displaying "prefilter" is only valuable to those who understand scheduling internals (btw the current diagnostics message could be improved as well to better match those produced by the default scheduler).

We log the plugin details in shim side. Thought of keeping the info as self contained in shim and exposing only the required info to core side. Yes, not only "prefilter" sounds different and bit difficult to interpret by the normal users, plugin name too :)

We could file a follow up to track this. For now, we can modify the message slightly.

"Predicate checks failed. Unable to find the suitable node to run this pod" ?

pbacsko:
> We could file a follow up to track this. For now, we can modify the message slightly.

I'm OK with a follow-up.

pkg/cache/context.go
manirajv06: Why we do need the cycle state once we decide the node?  Once the node is decided, we don't even iterate the next node. Isn't it? There is no need for us to run the `Filter()` for next node in the iterator.

BTW, We were clearing the cycle state from cache during the pod removal. But now we had changed our mind and decided to delete the cycle state as and when we are done with node selection to reduce the memory footprint of the scheduler cache. It is as good as keeping the variable in local memory :)

pkg/plugin/predicates/predicate_manager.go
manirajv06: Taken care

pkg/cache/context.go
pbacsko: > Why we do need the cycle state once we decide the node?

Isn't this method called from `tryNodes()` where we're iterating though nodes?

At this point, we only know that the node is an element of a set of feasible nodes. The `Filter()` of a plugin can still fail for a given `targetNode` and you have to proceed to the next one. So we can't just delete the `CycleState` here. That must happen later, we we ran out of target nodes.

manirajv06: Sorry, I thought we were deleting the cycle state in case of successful Filter calls, but thats not the case. We should wrap the delete operation inside nil err check. Does that sounds good?

pbacsko:
I think we're getting there. This change is now logically correct, but we can still improve it.

Suggestions:
1. Move `CycleState` deletion from `IsPodFitNode()` to scheduler cache's `updatePod()`:
```
if utils.IsAssignedPod(pod) && !utils.IsPodTerminated(pod) {
	// pod is bound to a node now: the scheduling cycle is over, so drop the
	// cycle state populated during PreFilter (no-op if it was never set)
	delete(cache.podsCycleState, key)
	// assign to node
	nodeInfo, ok := cache.nodesMap[pod.Spec.NodeName]
	...
``` 
This is more robust and cleaner. It's an unambiguous lifecycle point, runs under the write lock already held by `UpdatePod()`, consistent with the existing `delete(cache.podsCycleState, key)` in the terminated branch.

2. Revert back to using read locks in `IsPodFitNode()`. The write lock is only there for `DeleteCycleState()`. This serializes every predicate check against informer-driven cache updates (node/pod add/update), which is a throughput regression versus master (which used `LockForReads()`), and it's inconsistent with `IsPodFitNodeViaPreemption()`, which correctly uses `LockForReads()`.
If cycle-state deletion is moved to the pod-assignment path in updatePod (see above), Filter no longer mutates the cache.

pbacsko:
[YUNIKORN-3300  - Confirmed coverage gaps.pdf](https://github.com/user-attachments/files/30456791/YUNIKORN-3300.-.Confirmed.coverage.gaps.pdf)
[pr1043-missing-tests.patch](https://github.com/user-attachments/files/30456814/pr1043-missing-tests.patch)

@manirajv06 I asked Opus to check missing tests and there are many scenarios which are not covered at all. Please see the attached PDF and patch file which contains the generated tests. Some of them are skipped because you haven't implemented the changes I requested above. The patch should apply cleanly to master / PR branch because it contains pure additions.

manirajv06:
> Suggestions:

These suggestion doesn't even considered our decision making (We don't want to keep it unnecessarily to reduce the memory footprint of the cache) went behind the change and doesn't help us.

You can go through the suggestions and share your views. 

In general, rather than we go and back forth on AI tooling based reviews and to make the review comments look concise and clear, I kindly request you to go through the suggestions given by the tool and cross check with the "human" conversations :) so that further questions/comments could be added on top of the earlier comments in a meaningful way.

manirajv06:
> to check missing tests 

I haven't checked the attachments. But I do already have a feel that test coverage should be lesser especially when compared to the unit tests added in Core side. That could be because of my understanding that lot of new mocks need to be developed, especially mocking the results of different K8s PreFilter and Filter plugins itself which is quite time consuming and can be developed further through the follow ups.

pbacsko:
> > Suggestions:
> 
> These suggestion doesn't even considered our decision making (We don't want to keep it unnecessarily to reduce the memory footprint of the cache) went behind the change and doesn't help us.
> 
> You can go through the suggestions and share your views.
> 
> In general, rather than we go and back forth on AI tooling based reviews and to make the review comments look concise and clear, I kindly request you to go through the suggestions given by the tool and cross check with the "human" conversations :) so that further questions/comments could be added on top of the earlier comments in a meaningful way.

I thoroughly review the AI output - what you see from me is a trimmed down version + my own thoughts.  It's not just a blind copy-paste, the real output is much longer and detailed. I've found it extremely useful - critical bugs were caught and were proven by generated unit tests. It's not just some random noise.
The missing tests was different - that was verbatim from Opus. I didn't see the point of writing it manually - a nice list is already there. Your tests doesn't have to match that exactly.

Both suggestions make perfect sense to me: CycleState deletion is best done in `updatePod()` so we don't need write locks in `IsPodFitNode()`. You can absolutely treat those as something that came directly from me (and it did - it was tooling usage under my supervision). If you disagree that's fine - but if there ever was a discussion, I did not see that.

pbacsko:
> > to check missing tests
> 
> I haven't checked the attachments. But I do already have a feel that test coverage should be lesser especially when compared to the unit tests added in Core side. That could be because of my understanding that lot of new mocks need to be developed, especially mocking the results of different K8s PreFilter and Filter plugins itself which is quite time consuming and can be developed further through the follow ups.

We can be reasonable about it. Obviously we don't need to go for a 100% coverage. But this change adds significant complexity to both modules. At least the problems that were detected should be covered, just in case.

manirajv06:
> I've found it extremely useful - critical bugs were caught and were proven by generated unit tests. 

I agree with the first part to larger extent and really found it very useful. It help us lot. But not completely happy with the generated unit tests always. Few examples are https://github.com/apache/yunikorn-core/pull/1108 and https://github.com/apache/yunikorn-core/pull/1107. 

Yes, AI tooling definitely helps us and may mature overtime for sure but requires a manual assessment on the suggestions, reviews etc to take up further human conversation over the PR in a concise manner for now.

> I didn't see the point of writing it manually - a nice list is already there. Your tests doesn't have to match that exactly.

Yes, I do agree. Just that it is consuming good amount of time to go over the heaps of code changes to decide what to choose and what not to choose.

> Both suggestions make perfect sense to me: CycleState deletion is best done in `updatePod()` so we don't need write locks in `IsPodFitNode()`. You can absolutely treat those as something that came directly from me (and it did - it was tooling usage under my supervision). If you disagree that's fine - but if there ever was a discussion, I did not see that.

Please see https://github.com/apache/yunikorn-k8shim/pull/1043#discussion_r3658637383.

manirajv06:
> At least the problems that were detected should be covered, just in case

Yes, it helps. We probably need to take it up in bits and pieces.

pbacsko:
> Yes, I do agree. Just that it is consuming good amount of time to go over the heaps of code changes to decide what to choose and what not to choose.

Ah I see your point... ok, next time I'll summarize all stuff if needed. No worries.

> Please see https://github.com/apache/yunikorn-k8shim/pull/1043#discussion_r3658637383.

Thanks, I missed this comment. I see the reason behind it - but I don't feel this optimization is worth it. Under normal circumstances, it goes back to the core and the pod will be bound in milliseconds so `updatePod()` will remove it quickly after that.

manirajv06:
> That could be because of my understanding that lot of new mocks need to be developed, especially mocking the results of different K8s PreFilter and Filter plugins itself which is quite time consuming and can be developed further through the follow ups.

1. Introduced Mocks for PreFilter & Filter Plugins. Used those mocks to improve test coverage for PreFilter and Filter predicate checks in PredicateManager. 

2. Delete cycle state even during preemption predicate checks and added the tests as well

manirajv06:
> We could file a follow up to track this. For now, we can modify the message slightly.
> 
> "Predicate checks failed. Unable to find the suitable node to run this pod" ?

Updated the message in Core PR.

pkg/cache/external/scheduler_cache.go
```diff
@@ -577,6 +590,24 @@ func (cache *SchedulerCache) updatePVCRefCounts(node *framework.NodeInfo, remove
...
+func (cache *SchedulerCache) GetCycleState(pod *v1.Pod) *framework.CycleState {
+	return cache.podsCycleState[string(pod.UID)]
+}
+
+func (cache *SchedulerCache) UpdateCycleState(pod *v1.Pod, cycleState *framework.CycleState) {
+	if pod == nil {
+		return
+	}
+	cache.podsCycleState[string(pod.UID)] = cycleState
+}
```
wilfred-s: We should not have an update, we create a CycleState, and set it in the cache. We retrieve it and then it is removed. The next time we need a clean new cycle state.
The updates are made directly in the cycle state object always using that same reference

pkg/cache/context.go
```diff
@@ -692,6 +699,38 @@ func (ctx *Context) EventsToRegister(queueingHintFn fwk.QueueingHintFn) []fwk.Cl
 	return ctx.predManager.EventsToRegister(queueingHintFn)
 }
 
+// PreFilter evaluates given prefilter based predicates based on current context
+func (ctx *Context) PreFilter(name string, allocate bool) *si.PreFilterPredicatesResponse {
```
wilfred-s, line 703: Here we can and must create a new cycle state. This cycle state can be stored for later retrieval

pkg/cache/context.go
```diff
@@ -692,6 +699,38 @@ func (ctx *Context) EventsToRegister(queueingHintFn fwk.QueueingHintFn) []fwk.Cl
...
+			Success:       false,
+		}
+	}
+	// if pod exists in cache, try to run predicates
+	// need to lock cache here as predicates need a stable view into the cache
+	ctx.schedulerCache.LockForWrites()
+	defer ctx.schedulerCache.UnlockForWrites()
+	feasibleNodes, cycleState, err := ctx.predManager.PreFilter(pod, allocate)
+	if err == nil {
+		ctx.schedulerCache.UpdateCycleState(pod, cycleState)
```
wilfred-s: Not `UpdateCycleState` should be `StoreCycleState`

pkg/cache/external/scheduler_cache.go
```diff
@@ -209,6 +219,7 @@ func (cache *SchedulerCache) removeNode(node *v1.Node) (*v1.Node, []*v1.Pod) {
 		key := string(pod.UID)
 		delete(cache.assignedPods, key)
 		delete(cache.assumedPods, key)
+		delete(cache.podsCycleState, key)
```
wilfred-s: There should never be a cycle state that needs a cleanup based on a node removal. The scheduling cycle cleans up its state and if a pod gets removed that handles all other cases. A pod running on a node MUST not have a cycle state left.
If the assume or bind fail the cleanup should be automatic.

wilfred-s (changes requested):
CycleState cannot be kept over scheduling attempts. `framework.CycleState` is thread safe and can be accessed without race issues.
That normally means that we can create one when we start executing the predicates. It can be kept until the assume is done but after that it is no longer usable. It also means as soon as we start the bind it can be retrieved and it should be removed to be no longer accessible in the next cycle.

After any failure at any stage we MUST delete the cycle state.

manirajv06:
Cycle state would be created for every pod during PreFilter for both reservation and allocation filters just before calling the plugins in Predicate Manager not in Context. Made changes to delete the cycle state in the following cases:

1. If Filter fails
2. If Assume Pod fails
3. Once Bind Pod Volume operation is over, in both success and failure cases.
4. Already bounded Pods ( without volumes requirements).

manirajv06:
> If Filter fails

We should not delete the state when the filter fails for the given pod and node as we need it while trying next node from the list. Since state is required for binding process, we shall let it be there around until we are done with the binding process.

go.mod
```diff
@@ -22,112 +22,120 @@ go 1.26.0
 
 require (
 	github.com/apache/yunikorn-core v0.0.0-20260810093905-7dc1287c839f
```
wilfred-s: This should pull in the latest commit that has YUNIKORN-3300 changes in the core, at least

wilfred-s (changes requested):

go.mod
manirajv06: Taken care. All tests are passing. Can you take a lookl?

pkg/cache/context.go
```diff
@@ -706,9 +745,13 @@ func (ctx *Context) IsPodFitNode(name, node string, allocate bool) error {
 		return ErrorNodeNotFound
 	}
 	// need to lock cache here as predicates need a stable view into the cache
-	ctx.schedulerCache.LockForReads()
-	defer ctx.schedulerCache.UnlockForReads()
-	plugin, err := ctx.predManager.Predicates(pod, targetNode, allocate)
+	ctx.schedulerCache.LockForWrites()
+	defer ctx.schedulerCache.UnlockForWrites()
+	cycleState := ctx.schedulerCache.GetCycleState(pod)
```
wilfred-s: Something to still think about. The `CycleState` can be accessed concurrently. A read lock on the cache should be enough as a cycle state must exist and multiple filters can write into it if needed.
gets this inline with the preemption code below.

manirajv06: Yes, read lock is enough. Also taken care of write locks in other places too.

pkg/plugin/predicates/predicate_manager.go
```diff
@@ -36,22 +37,24 @@ import (
 	apiConfig "k8s.io/kubernetes/pkg/scheduler/apis/config"
 	"k8s.io/kubernetes/pkg/scheduler/apis/config/scheme"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
-	"k8s.io/kubernetes/pkg/scheduler/framework/plugins"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/names"
 	fwruntime "k8s.io/kubernetes/pkg/scheduler/framework/runtime"
 	"k8s.io/kubernetes/pkg/scheduler/metrics"
 
 	"github.com/apache/yunikorn-k8shim/pkg/log"
+
```
wilfred-s: NIT remove empty line

pkg/plugin/predicates/predicate_manager.go
```diff
@@ -191,80 +174,90 @@ func (p *predicateManagerImpl) removePodFromNodeNoFail(node fwk.NodeInfo, pod *v
...
+	} else {
+		status, feasibleNodes = p.runPreFilterPlugins(ctx, cycleState, *p.reservationPreFilters, pod)
 	}
-
-	// Run "filter" plugins on node
-	status, plugin = p.runFilterPlugins(ctx, filters, state, pod, node, skip)
-	if !status.IsSuccess() {
-		return plugin, errors.New(status.Message())
+	if !status.IsSuccess() && !status.IsSkip() {
+		return map[string]*si.Empty{}, cycleState, errors.New(status.Message())
```
wilfred-s: if pre-filter fails we should not return a cycle state as we should never go to filter and we skip storing it

pkg/cache/context.go
```diff
@@ -692,6 +699,38 @@ func (ctx *Context) EventsToRegister(queueingHintFn fwk.QueueingHintFn) []fwk.Cl
...
+		return &si.PreFilterPredicatesResponse{
+			FeasibleNodes: make(map[string]*si.Empty),
+			Success:       false,
+		}
+	}
+	// if pod exists in cache, try to run predicates
+	// need to lock cache here as predicates need a stable view into the cache
+	ctx.schedulerCache.LockForWrites()
+	defer ctx.schedulerCache.UnlockForWrites()
+	feasibleNodes, cycleState, err := ctx.predManager.PreFilter(pod, allocate)
```
wilfred-s: Can we read lock here and oinly pick the write lock for storing the cycle state?
We can never run the pre-filter on the same pod in multiple threads, there is no chance of a race between unlock cache and store of the cycle state

wilfred-s (changes requested):
minor things

wilfred-s:
based on the e2e tests a possible deadlock has been introduced as tests fail on deadlock detection traps.

pkg/plugin/predicates/predicate_manager.go
manirajv06: Taken care

pkg/cache/context.go
manirajv06: Taken care

manirajv06:
> based on the e2e tests a possible deadlock has been introduced as tests fail on deadlock detection traps.

Taken care.

We were deleting the cycle state after running the preemption predicates successfully. It is not correct. We should not delete the cycle state in this stage similar to normal predicate checks and take care of this only after Assume, Bind operations etc.

chenyulin0719 (changes requested):
@manirajv06 could you please fix the merge conflict? thanks.

manirajv06:
Addressed conflicts. Runs are passing as well.

@wilfred-s Can you take a final look?

wilfred-s:
fixed the merge conflicts, letting the build run to check.

pkg/cache/context.go
```diff
@@ -692,6 +699,40 @@ func (ctx *Context) EventsToRegister(queueingHintFn fwk.QueueingHintFn) []fwk.Cl
...
+	}
+	// if pod exists in cache, try to run predicates
+	// need to lock cache here as predicates need a stable view into the cache
+	ctx.schedulerCache.LockForReads()
+	feasibleNodes, cycleState, err := ctx.predManager.PreFilter(pod, allocate)
+	ctx.schedulerCache.UnlockForReads()
+	if err == nil {
+		ctx.schedulerCache.LockForWrites()
+		ctx.schedulerCache.StoreCycleState(pod, cycleState)
+		ctx.schedulerCache.UnlockForWrites()
```
wilfred-s: should follow the same setup as for the delete. Lock inside the call which means we can remove the UnlockForWrites and LockForWrites calls.
Besides that we're good to go.

wilfred-s (changes requested):

manirajv06:
Fixed conflicts

manirajv06:
@wilfred-s Can you take a final look? As there are lot of PR's e2e tests are failing and not able to green build because of this.

tigerquoll:
Heads-up that this PR is currently blocking everything else: **master does not compile without it.**

The YUNIKORN-3435 dependency bump (`a2222d65`) pulled in a scheduler-interface that adds `PreFilterPredicates` to `api.ResourceManagerCallback`, but the shim-side implementation lands here, so a clean checkout of master fails to build:

```
pkg/cache/scheduler_callback.go:44:37: cannot use &AsyncRMCallback{} (value of type *AsyncRMCallback) as
api.ResourceManagerCallback value in variable declaration: *AsyncRMCallback does not implement
api.ResourceManagerCallback (missing method PreFilterPredicates)
```

`pkg/cache`, `pkg/cmd/shim` and `pkg/shim` all fail with `[build failed]`. Master's own CI has been red on this since 4 Sep — most recent run [33974617433](https://github.com/apache/yunikorn-k8shim/actions/runs/33974617433) — and every open PR inherits the failure on rebase, so unit-test results across the repo are currently uninformative.

Not asking to shortcut the review here; just flagging that the SI/core side went in ahead of the shim side, so this one is on the critical path.

manirajv06:
Yes, we are aware of this and working upon it. We are almost done. It won't take more than a day from now. Thanks.

@wilfred-s

wilfred-s (approved):
LGTM
thanks for working through this major change.

PR #1052, opened 2026-07-16
[YUNIKORN-3316] Shim changes for handling AssumePod errors

adityadtu5:
### Description
During AssumePod cache update if there are any failures task goes to failed state and the application is completed if the failed task is the last task for the application.

It would be great if instead of failing, Yunikorn can retry the task again (and possibly core will select different node).

As part of these changes following changes are done - 
1) Shim will send this termination type to core after exhausting all retries on assume pod failures. 
2) Task will be kept in scheduling state so that it can be retries again.
3) Task will clear node assignment, clear pvc bindings from cache and will remove itself from node (`forgetPod`).

~~E2E tests are part of this PR. To enable failure injection during assumepod, e2e test will create a configmap. Yunikorn code returns error if this configmap is present.~~

k8s-shim PR - 
core PR - https://github.com/apache/yunikorn-core/pull/1104
Scheduler Interface PR - https://github.com/apache/yunikorn-scheduler-interface/pull/162

### Type of change

- [x] Bug Fix

### Jira issue
 https://issues.apache.org/jira/browse/YUNIKORN-3316

### AI Tooling
If an AI tool was used:

### How has this been tested?
- [x] `make e2e_test` was run, and no failures reported.
- [x] new e2e tests have been added.

I have tested this with a custom binary in which I inject failures during volumebinder's assumepodvolumes call.

After failing for sometime and few retries, pod eventually gets scheduled. 
Events of a pod from testing
```
Events:
  Type     Reason             Age                From      Message
  ----     ------             ----               ----      -------
  Normal   Scheduling         27s                yunikorn  default/app-b-7bd9975549-4b85j is queued and waiting for allocation
  Warning  AssumePodFailed    22s (x2 over 24s)  yunikorn  Node assignment failed for default/app-b-7bd9975549-4b85j on node yunikorn-3128-worker, it will be retried
  Normal   Scheduled          20s                yunikorn  Successfully assigned default/app-b-7bd9975549-4b85j to node yunikorn-3128-worker
  Normal   PodBindSuccessful  20s                yunikorn  Pod default/app-b-7bd9975549-4b85j is successfully bound to node yunikorn-3128-worker
  Normal   Pulled             20s                kubelet   spec.containers{sleep-container}: Container image "registry.k8s.io/pause:3.9" already present on machine and can be accessed by the pod
  Normal   Created            20s                kubelet   spec.containers{sleep-container}: Container created
  Normal   Started            20s                kubelet   spec.containers{sleep-container}: Container started
```

### Questions:


pkg/cache/context.go
```diff
@@ -884,6 +888,13 @@ func (ctx *Context) AssumePod(name, node string) error {
 	return nil
 }
 
+func (ctx *Context) assumePodShouldInjectError() bool {
+	clientSet := ctx.apiProvider.GetAPIs().KubeClient.GetClientSet()
+	_, err := clientSet.CoreV1().ConfigMaps(ctx.namespace).Get(
+		context.TODO(), constants.FaultInjectConfigMapName, metav1.GetOptions{})
+	return err == nil
+}
+
```
pbacsko: We shouldn't be doing this just for the sake of a simulated `AssumePod()` failure. It just adds extra complexity and noise to the production code. Also, writing and maintaining such e2e tests are more difficult.

There are `MockScheduler`-based tests exactly for this purpose. It's very easy to simulate errors with the mocks and these tests are essentially run the whole scheduler without a real K8s cluster. Much easier to run, validate and debug. In fact, `TestAssumePodError()` is already there - just add the new assertions. Add new test cases separately if needed.

Edit: meant to address lines 891-897

pbacsko (changes requested):
My blunt take: remove the fault injection code from `cache.Context` and the e2e tests and perform them from a `MockScheduler`-based integration test. Less code, faster, easier to debug.

pkg/cache/context.go
adityadtu5: I know that it is an ugly hack. But MockScheduler can not simulate complete end to end tests. That is the reason, I have added this as the proof that the change is working.

adityadtu5: `TestAssumePodError` is already updated. All other unit tests are already in place.

pbacsko: > I know that it is an ugly hack. But MockScheduler can not simulate complete end to end tests. That is the reason, I have added this as the proof that the change is working.

Have you discussed this with @wilfred-s? What's his take on this approach?

It's perfectly OK to have this as an integration test. I had to do the same when I fixed YUNIKORN-2520. The goal is to verify the inner workings of shim+core when performing a scheduling cycle. I don't see an added value of an e2e tests, especially when it comes with such maintainbility costs.

adityadtu5: I have asked @wilfred-s to review and let me know his thoughts about it. 

IMHO, it is very valuable in proving that change is working end to end and gives immense confidence in shipping this change.

Anyways, I just have reverted that change. I can add it later if needed.

adityadtu5:
> My blunt take: remove the fault injection code from `cache.Context` and the e2e tests and perform them from a `MockScheduler`-based integration test. Less code, faster, easier to debug.

E2E tests are removed.

pkg/shim/scheduler_test.go
```diff
@@ -248,12 +249,33 @@ func TestAssumePodError(t *testing.T) {
...
 	cluster.AddPod(pod1)
 
-	// expect app to enter Completing state with allocation+ask removed
-	err = cluster.waitForApplicationStateInCore("app0001", partitionName, "Completing")
-	assert.NilError(t, err)
-	app := cluster.getApplicationFromCore("app0001", partitionName)
-	assert.Equal(t, 0, len(app.GetAllRequests()), "asks were not removed from the application")
-	assert.Equal(t, 0, len(app.GetAllAllocations()), "allocations were not removed from the application")
+	// wait for the application to be registered and its ask queued in the shim
+	cluster.waitAndAssertApplicationState(t, "app0001", cache.ApplicationStates().Accepted)
```
pbacsko: This change causes the test to be flaky - it keeps appearing on my local laptop. "Accepted" state is only observable for a bried amount of time. Sometimes it moves on to "Running" quickly and the part times out.

Just simply remove this line.

pbacsko (changes requested):
1. Fix the flaky test
2. **Follow-up:** don't fetch PodVolumes with FindPodVolumes before reverting. This object should be stored separately in Context or in the scheduler cache. It's not a problem right now, but if we want to change the rollback behavior to be in-line with the default scheduler, then we have to perform a revert in case of a bind failure (we don't do this atm). So if we do it to achieve K8s parity (which is desirable), then the revert operation can fail inside the volumeBinder's cache. 
3. **Follow-up:** we return nil from AssumePod if either the node/pod is not found. I just spotted this. This is a pre-existing problem, not related to this change.  Investigate this in a separate JIRA (low prio).

More details on 1)

Revert scenarios
|Scenario|Assume happened?|Re-fetch works?|
|---|---|---|
| All retries fail at AssumePodVolumes   | 	No (self-reverted)    | N/A |
| Assume succeeds, bind proceeds | 	Yes  | N/A  (no revert needed)|
| Assume succeeds, then core release -> ForgetPod  (mostly a non-issue for us)| 	Yes   | Risky, especially dynamic PVCs|
| Assume succeeds, bind fails, cleanup needed   | 	Yes   | Risky |

So the gap we have is "Assume succeeds, bind fails, cleanup needed". This is worth implementing as a follow-up, but it has the PodVolumes impact.

adityadtu5:
> 1. Fix the flaky test
> 2. **Follow-up:** don't fetch PodVolumes with FindPodVolumes before reverting. This object should be stored separately in Context or in the scheduler cache. It's not a problem right now, but if we want to change the rollback behavior to be in-line with the default scheduler, then we have to perform a revert in case of a bind failure (we don't do this atm). So if we do it to achieve K8s parity (which is desirable), then the revert operation can fail inside the volumeBinder's cache.
> 3. **Follow-up:** we return nil from AssumePod if either the node/pod is not found. I just spotted this. This is a pre-existing problem, not related to this change.  Investigate this in a separate JIRA (low prio).
> 
> More details on 1)
> 
> Revert scenarios
> 
> Scenario	Assume happened?	Re-fetch works?
> All retries fail at AssumePodVolumes	No (self-reverted)	N/A
> Assume succeeds, bind proceeds	Yes	N/A (no revert needed)
> Assume succeeds, then core release -> ForgetPod (mostly a non-issue for us)	Yes	Risky, especially dynamic PVCs
> Assume succeeds, bind fails, cleanup needed	Yes	Risky
> So the gap we have is "Assume succeeds, bind fails, cleanup needed". This is worth implementing as a follow-up, but it has the PodVolumes impact.

1) I tried running locally the test multiple times and have not faced any failures. The application state assertion is not very important for this flow, so I will remove it. 
2) There are jiras in place to handle volume and pod bind failures. See epic - YUNIKORN-2804 for more details. They will be done once this PR is merged as this PR is base for them. Saving podvolumes in scheduler sounds correct. Not sure how default scheduler does the cleanup. I will go through it and file jira for follow up. 
3) Yes, ideally it should return error if either pod or node is not found. I will file a jira to track that error handling.

pkg/shim/scheduler_test.go
adityadtu5: removed this assertion as it was not very important for this test

pbacsko:
> 2) We can not save the pod volumes into scheduler cache as revert needs to work even if yunikorn lose its state in between. 

The revert I'm talking about is the manipulation of the volumeBinder's own cache. That is re-populated if YK ever restarts. The problem is when YK operates normally - in that case, re-fetching a new PodVolumes can cause the revert to be a no-op in the volumeBinder's assume cache. That is a problem which need to be addressed (not in this JIRA).

adityadtu5:
> > 2. We can not save the pod volumes into scheduler cache as revert needs to work even if yunikorn lose its state in between.
> 
> The revert I'm talking about is the manipulation of the volumeBinder's own cache. That is re-populated if YK ever restarts. The problem is when YK operates normally - in that case, re-fetching a new PodVolumes can cause the revert to be a no-op in the volumeBinder's assume cache. That is a problem which need to be addressed (not in this JIRA).

Sorry I did not get it first, I have update my earlier comment.

pkg/cache/task.go
```diff
@@ -620,6 +620,58 @@ func (task *Task) failWithEvent(errorMessage, actionReason string) {
...
+// pending ask so it can be re-scheduled on a different node.
+// Must be called without holding the task lock.
+func (task *Task) rollbackOnAssumePodFailure(allocationKey, nodeID string) {
+	// Read fields needed for event posting and release request under read lock.
+	task.lock.RLock()
+	podCopy := task.pod.DeepCopy()
+	alias := task.alias
+	appID := task.applicationID
+	partition := task.application.partition
+	task.lock.RUnlock()
```
pbacsko: I think it is better from readability if you merge L629-634 with L642-L645 under a single write lock.

pkg/cache/task_test.go
```diff
@@ -930,3 +934,94 @@ func TestCheckPodMetadataBeforeScheduling(t *testing.T) {
...
+	pod := &v1.Pod{
+		TypeMeta:   metav1.TypeMeta{Kind: "Pod", APIVersion: "v1"},
+		ObjectMeta: metav1.ObjectMeta{Name: "rollback-pod", UID: "rollback-uid"},
+	}
+	task := NewTask(allocationKey, app, ctx, pod)
+	task.sm.SetState(TaskStates().Scheduling)
+	task.lock.Lock()
+	task.allocationKey = allocationKey
+	task.nodeName = nodeID
+	task.lock.Unlock()
```
pbacsko: This object only exists here at this point, no need to lock-unlock.

pkg/cache/context.go
```diff
@@ -897,6 +897,37 @@ func (ctx *Context) ForgetPod(name string) {
...
+	if pod == nil {
+		return
+	}
+	node := ctx.schedulerCache.GetNode(nodeID)
+	if node == nil {
+		return
+	}
+	podVolumeClaims, err := ctx.apiProvider.GetAPIs().VolumeBinder.GetPodVolumeClaims(ctx.klogger, pod)
+	if err != nil {
+		log.Log(log.ShimContext).Debug("RevertPodVolumeAssumptions: failed to get pod volume claims",
```
pbacsko: we better log it with an appropriate log level

pkg/cache/context.go
```diff
@@ -897,6 +897,37 @@ func (ctx *Context) ForgetPod(name string) {
...
+			zap.String("pod", podName), zap.Error(err))
+		return
+	}
+	podVolumes, _, err := ctx.apiProvider.GetAPIs().VolumeBinder.FindPodVolumes(ctx.klogger, pod, podVolumeClaims, node.Node())
+	if err != nil || podVolumes == nil {
+		log.Log(log.ShimContext).Debug("RevertPodVolumeAssumptions: failed to find pod volumes",
```
pbacsko: we better log it with an appropriate log level

pbacsko (changes requested):
found a couple of smaller things

pkg/cache/context.go
adityadtu5: changes log levels in the last commit

pkg/cache/context.go
adityadtu5: done

pkg/cache/task.go
adityadtu5: I think since DeepCopy is expensive operation, better to do it with read lock only.

pkg/cache/task_test.go
adityadtu5: thanks, done

pkg/cache/task.go
pbacsko: cc @wilfred-s what do you think?

wilfred-s: sounds good, one lock to do all the changes

adityadtu5: moved things into one lock

pbacsko (approved):

wilfred-s (approved):
LGTM

PR #1053, opened 2026-07-18
[YUNIKORN-3329] Reject task-group placeholder resources that overflow or are negative

thc1006:
### Description

`GetTGResource` (`pkg/common/resource.go`) computed the gang placeholder ask as `members * minResource` with unchecked `int64` accessors, and `setTaskGroups` aggregates the per-group results with the unchecked `common.Add`. Because `minMember` and `minResource` come unbounded from the `yunikorn.apache.org/task-groups` annotation, the ask could wrap to a negative value in three places: the `Quantity.MilliValue()`/`Value()` accessor, the `members * value` product, and the cross-task-group aggregate. That negative value is then sent to the core as `AddApplicationRequest.PlaceholderAsk`.

Rather than saturate (which would hide an invalid annotation as a near-infinite gang reservation), this rejects the annotation at the parse boundary. `GetTaskGroupsFromAnnotation` now returns an error when a `minResource` is negative, when a cpu/other quantity's int64 accessor would overflow, when `minMember * minResource` would overflow, or when the aggregate across task groups would overflow.

The aggregate is checked in the scheduler-interface canonical key-space that `GetTGResource` actually emits (an implicit `pods` count per task group, and `cpu` mapped to `siCommon.CPU`/`vcore`), with same-canonical-key collisions rejected, so the check matches the real placeholder ask rather than the raw annotation keys.

Behaviorally this is the existing invalid-annotation path (YUNIKORN-1089): the task-group definition is discarded and the application is scheduled without gang reservation rather than being rejected outright. The change only ensures a wrapped/negative placeholder ask is not sent to the core.

### Type of change

- [x] Bug Fix

### Jira issue

Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3329

### AI Tooling
If an AI tool was used:

### How has this been tested?

`TestGetTaskGroupsFromAnnotationValidatesResources` rejects a negative cpu, an int64-accessor overflow (cpu `MilliValue` and memory `Value`), a `minMember * minResource` product overflow, and an aggregate overflow, including the `cpu`/`vcore` canonical collision and the explicit/implicit `pods` collision, and accepts a normal task group. `TestValidatedTaskGroupsNeverProduceNegativePlaceholderAsk` is a caller-level property test: for every annotation the validator accepts, it rebuilds the placeholder ask the way `setTaskGroups` does (`common.Add` over `GetTGResource`) and asserts every resource is non-negative. Both fail before the change and pass after. `make test_all` passes (lint reports 0 issues, license checks OK, all unit tests pass); e2e was not run.

### Questions:

### DIFF
--- a/pkg/cache/utils.go
+++ b/pkg/cache/utils.go
@@ -21,11 +21,13 @@ package cache
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 
 	v1 "k8s.io/api/core/v1"
 
 	"github.com/apache/yunikorn-k8shim/pkg/common/constants"
 	"github.com/apache/yunikorn-k8shim/pkg/common/utils"
+	siCommon "github.com/apache/yunikorn-scheduler-interface/lib/go/common"
 )
 
 func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]TaskGroup, error) {
@@ -40,6 +42,7 @@ func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]TaskGroup, error) {
 		return nil, err
 	}
 	// json.Unmarshal won't return error if name or MinMember is empty, but will return error if MinResource is empty or error format.
+	totals := make(map[string]int64)
 	for _, taskGroup := range taskGroups {
 		if taskGroup.Name == "" {
 			return nil, fmt.Errorf("can't get taskGroup Name from pod annotation, %s",
@@ -57,6 +60,62 @@ func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]TaskGroup, error) {
 			return nil, fmt.Errorf("minMember cannot be negative, %s",
 				taskGroupInfo)
 		}
+		if err := validateTaskGroupResources(taskGroup, totals); err != nil {
+			return nil, err
+		}
 	}
 	return taskGroups, nil
+}
+
+// validateTaskGroupResources folds one task group's placeholder-ask contribution
+// into totals (the canonical-key aggregate across all groups) and rejects an ask
+// that cannot be a non-negative int64: a negative minResource, an int64 overflow
+// (the accessor, the minMember product, or the cross-group aggregate), or a
+// same-canonical-key collision. It mirrors GetTGResource, which emits minMember
+// "pods" plus minMember*minResource per resource ("cpu" canonicalizes to
+// siCommon.CPU, "vcore"); a collision (cpu with vcore, or an explicit "pods") is
+// rejected because GetTGResource would otherwise let one silently overwrite the
+// other (cpu vs vcore by map-iteration order, an explicit "pods" over the implicit count).
+func validateTaskGroupResources(taskGroup TaskGroup, totals map[string]int64) error {
+	members := int64(taskGroup.MinMember)
+	// seen holds the canonical keys this task group has already contributed, so a
+	// same-group collision is rejected instead of silently overwritten. The implicit
+	// "pods" count claims the "pods" key up front.
+	seen := map[string]bool{"pods": true}
+	if totals["pods"] > math.MaxInt64-members {
+		return fmt.Errorf("aggregate placeholder request for \"pods\" overflows int64 across taskGroups")
+	}
+	totals["pods"] += members
+
+	for resName, quantity := range taskGroup.MinResource {
+		if quantity.Sign() < 0 {
+			return fmt.Errorf("minResource %q in taskGroup %q cannot be negative", resName, taskGroup.Name)
+		}
+		cpu := resName == v1.ResourceCPU.String()
+		canonical := resName
+		milliConvert := int64(1)
+		if cpu {
+			canonical = siCommon.CPU
+			milliConvert = 1000
+		}
+		if seen[canonical] {
+			return fmt.Errorf("minResource %q in taskGroup %q collides with another resource under canonical key %q", resName, taskGroup.Name, canonical)
+		}
+		seen[canonical] = true
+		// Reject before the int64 accessor or the minMember product can overflow:
+		// members*value (in milli-units for cpu) must stay within int64.
+		if quantity.CmpInt64(math.MaxInt64/(members*milliConvert)) > 0 {
+			return fmt.Errorf("minResource %q in taskGroup %q overflows int64 when scaled by minMember %d", resName, taskGroup.Name, members)
+		}
+		value := quantity.Value()
+		if cpu {
+			value = quantity.MilliValue()
+		}
+		contribution := members * value
+		if totals[canonical] > math.MaxInt64-contribution {
+			return fmt.Errorf("aggregate placeholder request for %q overflows int64 across taskGroups", canonical)
+		}
+		totals[canonical] += contribution
+	}
+	return nil
 }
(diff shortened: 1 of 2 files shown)


chia7712:
> Pending. ASF Jira account requested via https://selfserve.apache.org/jira-account.html; will link the YUNIKORN issue once approved.

I have approved your request. Please file a jira ticket for this PR. Thanks!

thc1006:
> > Pending. ASF Jira account requested via https://selfserve.apache.org/jira-account.html; will link the YUNIKORN issue once approved.
> 
> I have approved your request. Please file a jira ticket for this PR. Thanks!

I just got my ID haha! ok I wil do it later !!! 
Get my breakfast first!!!

chia7712:
> Get my breakfast first!!!

sure. Food comes first. No rush!

pkg/cache/utils.go
```diff
@@ -58,5 +61,77 @@ func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]TaskGroup, error) {
 				taskGroupInfo)
 		}
 	}
+	if err := validateTaskGroupResources(taskGroups); err != nil {
+		return nil, err
+	}
```
wilfred-s: Instead of doing a new range can this not be included in the range above? Saves one loop
```
totals := make(map[string]int64)
for range taskGroups {
...
    err := validateTaskGroupResources(taskGroup, totals)
    if err != nil {
        return nil, err
    }
}
```
Check totals inside the validate call directly.

pkg/cache/utils.go
```diff
@@ -58,5 +61,77 @@ func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]TaskGroup, error) {
...
+// actually emit: each task group contributes an implicit "pods" count plus one
+// entry per minResource under its scheduler-interface canonical name ("cpu" becomes
+// siCommon.CPU, i.e. "vcore"), and the groups are summed. minMember and minResource
+// come unbounded from the pod annotation, so an unchecked int64 accessor, a
+// minMember*minResource product, or the aggregate across task groups can wrap to a
+// negative value that is then sent to the core as PlaceholderAsk. Because "cpu"
+// canonicalizes to the same key as "vcore" and an explicit "pods" collides with the
+// implicit count, the aggregate is computed in the canonical key-space with
+// same-key collisions rejected, so the check matches the real placeholder ask.
+func validateTaskGroupResources(taskGroups []TaskGroup) error {
```
wilfred-s: Change signature, see above

pkg/cache/utils.go
```diff
@@ -58,5 +61,77 @@ func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]TaskGroup, error) {
...
+	totals := make(map[string]int64)
+	for _, taskGroup := range taskGroups {
+		members := int64(taskGroup.MinMember)
+		// group mirrors the si.Resource GetTGResource emits for this task group.
+		group := map[string]int64{"pods": members}
```
wilfred-s: update totals directly with a check for overflow and return error on fail.

pkg/cache/utils.go
```diff
@@ -58,5 +61,77 @@ func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]TaskGroup, error) {
...
+		for resName, quantity := range taskGroup.MinResource {
+			value, err := checkedResourceValue(resName, quantity, taskGroup.Name)
```
wilfred-s: * check sign here first, error on fail
* make canonical for CPU 
* calculate max value based on type (cpu/vcore or not) and members:
```
milliConvert := 1
if cpu {
	milliConvert = 1000
}
maxVal := math.MaxInt64/(members*milliConvert)
if quantity.CmpInt64(maxVal) > 0 {
 return err...
}
```
* update totals directly with a check for overflow and return error on fail

pkg/cache/utils.go
```diff
@@ -58,5 +61,77 @@ func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]TaskGroup, error) {
...
+				return fmt.Errorf("minResource %q in taskGroup %q collides with canonical resource %q", resName, taskGroup.Name, canonical)
+			}
+			group[canonical] = members * value
+		}
+		for name, value := range group {
+			if totals[name] > math.MaxInt64-value {
+				return fmt.Errorf("aggregate placeholder request for %q overflows int64 across taskGroups", name)
+			}
+			totals[name] += value
+		}
```
wilfred-s: directly in the above loop

pkg/cache/utils.go
```diff
@@ -58,5 +61,77 @@ func GetTaskGroupsFromAnnotation(pod *v1.Pod) ([]TaskGroup, error) {
...
+	}
+	return nil
+}
+
+// checkedResourceValue returns the int64 value GetTGResource uses for a
+// (resourceName, quantity) pair, rejecting a negative or non-int64-representable
+// quantity. cpu is measured in milli-units (MilliValue), everything else in whole
+// units (Value), matching GetTGResource.
+func checkedResourceValue(resName string, quantity resource.Quantity, groupName string) (int64, error) {
```
wilfred-s: This can be done directly in the loop above

wilfred-s (changes requested):

thc1006:
Thanks, this reads a lot cleaner. I've pushed the rework:
- Validation now runs inside the existing task-group loop, no second pass.
- `validateTaskGroupResources(taskGroup, totals)` takes the shared `totals` and updates it directly.
- Per resource: sign check first, then the cpu->vcore canonical, then your `maxVal := math.MaxInt64/(members*milliConvert)` with `CmpInt64`. Dropped `checkedResourceValue` and `maxMilliCPU`.

One place I didn't fold in: I kept a small per-group `seen` set that rejects a same-canonical-key collision (cpu together with vcore, or an explicit `pods`). GetTGResource adds the implicit `pods` first and then iterates the map, so an explicit `pods` deterministically overwrites that count, and cpu vs vcore overwrite each other by map order. Rejecting felt safer than passing a silently-overwritten ask to the core, but I'm happy to drop it and let them sum if you'd rather.

Smaller note on the cpu bound: `maxVal` floor-divides by 1000, so it conservatively rejects cpu in the sub-milli remainder above roughly 9e15 cores (`MilliValue` rounds up). That's fail-safe and well past anything real, so I left your simpler bound, but I can make it an exact Quantity limit if you want it precise.

I also strengthened the property test to check each accepted ask against an exact big.Int recomputation instead of just the sign, which pins the validator's canonicalization to GetTGResource's.

thc1006:
@wilfred-s when you have a moment, could you take another look? The validation now runs inside the existing task-group loop, so the separate pass is gone. I also added a unit test for the pods aggregate-overflow guard. Thanks for the earlier review.

thc1006:
Hi @wilfred-s, I've reworked this along the lines you suggested. The aggregation now happens inside the single pass over the task groups: `GetTaskGroupsFromAnnotation` calls `validateTaskGroupResources` once per group, and that checks the sign and the overflow inline, updates the running totals directly, and returns an error on failure. The separate second range is gone. I also rebased onto current master. Could you take another look when you have a chance?

wilfred-s (approved):
LGTM, waiting on the CI to finish before committing.

PR #1060, opened 2026-08-09
[YUNIKORN-3356] Split Kubernetes clients by concern (writes/informers/events)

tigerquoll:
### What is this PR for?

Implements [YUNIKORN-3356](https://issues.apache.org/jira/browse/YUNIKORN-3356): separates the shim's event traffic from its scheduling traffic, gives every client its own rate-limit policy and a distinct User-Agent, and bounds event traffic under a storm. Previously `kubernetes.qps` applied to each of two general-purpose clientsets independently (effective ceiling 2× the configured value), an event burst shared a token bucket with informer relists, events could not be limited without limiting binds, and no client was identifiable server-side.

This PR has a follow up that provides a suggested APF configuration for the YuniKorn helm chart.

| Client | User-Agent | Serves | Client-side limiter |
|---|---|---|---|
| scheduler | `yunikorn-scheduler` | Bind, Create/Delete, status updates, configmap loads, volume binder, predicate handle, and both informer factories | none by default; `kubernetes.qps`/`kubernetes.burst` as opt-in cap (`< 0` = no limiter, `0` = client-go's 5/10) |
| events | `yunikorn-scheduler/events` | events/v1 broadcaster sink | Normal events shed above `kubernetes.eventQPS`/`kubernetes.eventBurst` (new keys, default 200/400); Warning events always sent; server-informed mute on 429 for all events |
| bootstrap | `yunikorn-bootstrap` | two startup ConfigMap GETs (both binaries) | client-go defaults (5/10) |
| admission controller | `yunikorn-admission-controller` | AC informers, webhook and secret writes | client-go defaults (5/10) |

### Design rationale

Server-side API Priority & Fairness (APF) governs every request on every supported Kubernetes version (default-on since 1.20, GA 1.29) and, unlike a client rate limit/token bucket, is fair across tenants and work-aware. We take advantage of that by:

1. **Having server side APF arbitrate everything that is sent.** Client-side limiting is retained only where it does something APF cannot.
2. **Not throttling must-complete traffic (binds, status) on the client side by default.** We should not try and guess the state of the server — we take heed of the 429 metadata dynamically suggesting a slowdown period. Informers share the scheduler client: WATCH bypasses the client-go limiter anyway and relists are rare, so they do not need a client of their own.
3. **Using client-side limiting only to avoid sending discardable traffic.** Events are the sole discardable class, and within them only the Normal ones: three per successfully scheduled pod (`Scheduling`, `Scheduled`, `PodBindSuccessful`) plus the core's informational records, spiking exactly with throughput. Dropping those is the one action YuniKorn can take to lower API server load that does not compromise on core scheduling functionality.

### How events are shed

A standard token bucket "delay on limit" implementation for the events clientset is not the right fit here - the events/v1 broadcaster spawns a goroutine per event, so delaying traffic converts traffic backlogs into unbounded goroutine growth with nothing ever dropped. Instead we configure a qps/burst bucket to manage event traffic, but we drop events rather than delay them if the event client is over budget.

Shedding is type-aware. **Warning events — the ones that explain why a pod is not running (`FailedScheduling`, `PodUnschedulable`, `TaskRejected`, `NodeRejected`, `ApplicationFailed`, the gang scheduling failures) — are never shed by the bucket and consume no token.** `PodUnschedulable` is retyped from Normal to Warning to match (kube-scheduler's `FailedScheduling` is a Warning). Only Normal events are shed above `eventQPS`.

`kubernetes.eventLevel` (`normal` | `warning`, default `normal`) mirrors the two Kubernetes event types and is applied at the recorder, so a suppressed event is never copied, cached or written. It is hot-reloadable like the log level.

Once APF is configured separately for events, rather than relying on a statically configured rate limit, we can rely on the server itself to tell us to back off (by sending 429 status codes with a suggested slow down period). In this case we mute all events for the nominated cool down period rather than delay them: client-go drops a server-rejected event itself after its Retry-After retries (`recordEvent`: "Server rejected event (will not retry!)"), so the mute only saves the round trips and the parked goroutines.

The event dropping is implemented inside the event sink object. The 429 status detection and removal is done via an http round-tripper added to the events clientset. Only the event path gets this treatment. For must-complete traffic, client-go's Retry-After retransmission behaviour is left in place and remains untouched.

### Behaviour changes

- `kubernetes.qps`/`kubernetes.burst` now apply to the scheduler client (scheduling actions and informers) as a single bucket; defaults change **1000/1000 → no limiter (`-1`)**. Deployments relying on the implicit 1000 cap should set it explicitly.
- Upgrade corner cases — what happens to a config already deployed under the old semantics:

  | Existing config | Old behaviour | New behaviour |
  |---|---|---|
  | `burst` set, `qps` unset | qps silently defaulted to 1000, so the burst value applied | no limiter; burst ignored¹ — set `kubernetes.qps` to restore one (burst then defaults to the same value) |
  | `qps: "0"` explicitly | client-go's 5/10 defaults | unchanged: client-go's 5/10 defaults |
  | `qps` set, `burst` unset | effective qps/1000 | qps/qps — same sustained rate, less bursty |
  | `qps` set, `burst: 0` | crashed client construction (startup crash-loop) | burst defaults to qps |

  ¹ burst is a bucket *size*, qps its *refill rate* — without a positive refill rate there is no limiter for the burst value to shape.

  For the new event keys the same convention applies to the rate: `eventQPS <= 0` disables shedding, while an unset `eventBurst` falls back to its own default (400) rather than to `eventQPS`.
- **`PodUnschedulable` and `TaskFailed` events are now type Warning** (were Normal); every other event site was cross-checked against the Kubernetes Normal/Warning semantics.
- **New key `kubernetes.eventLevel`** (`normal` | `warning`), hot-reloadable.
- **Admission controller**: both of its clients run on client-go's defaults (5/10) with the `yunikorn-admission-controller` user agent. Previously its informer client ran on the scheduler configuration defaults (1000/1000, the scheduler ConfigMap keys never actually applied to it) and its webhook client on client-go's 5/10.
- **All clients send distinct User-Agents** (`yunikorn-scheduler`, `yunikorn-scheduler/events`, `yunikorn-bootstrap`, `yunikorn-admission-controller`) for apiserver-side attribution.

### Type of change

- Improvement

### Jira issue

Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3356

- `make test` (18 packages) and `make lint` clean; both binaries build.
- Unit: qps/burst normalisation table; `newRestConfig` plumbing; sink shed/pass-through/mute paths with a fake clock, including Warning events bypassing the bucket and being muted; recorder level filter over both levels × both types; `PodUnschedulable` and `TaskFailed` asserted as Warning; `eventLevel` parsing with fallback; transport strip/record (httptest, asserting a single request in ~10ms vs 11 in 10s); integration-level crash-loop regression through `kubernetes.NewForConfig`.
- Live on kind (K8s v1.36.1) + KWOK (500 nodes), taken on the previous revision of this PR (four-client layout; the event path and the scheduler write path are unchanged in mechanism since):
  - defaults: 3,000-pod bind burst at ~1,900 binds/s;
  - `qps: 100` without burst: boots with a working 100/100 limiter (crash-looped before the fix); cap enforces ~134 binds/s;
  - event storm: 9,805 events shed with one throttled warning; steady-state delivery lossless;
  - APF squeeze (2,394 rejections served): server escalated Retry-After to 31–32s, mute engaged (`delaySource: transport`), 869 events shed in one window, no goroutine accumulation (vs ~360 parked pre-transport).

Generated by Author with assistance from Claude Code.

### Questions:

- [x] The change needs documentation, a pull request for apache/yunikorn-site repository will be created.
- [x] There is breaking changes for older versions: jira is tagged with `release-notes` label.

### DIFF
--- a/pkg/admission/webhook_manager.go
+++ b/pkg/admission/webhook_manager.go
@@ -86,22 +86,11 @@ type webhookManagerImpl struct {
 
 // NewWebhookManager is used to create a new webhook manager
 func NewWebhookManager(conf *conf.AdmissionControllerConf) (WebhookManager, error) {
-	kubeconfig, err := client.CreateRestConfig(conf.GetKubeConfig())
-	if err != nil {
-		log.Log(log.AdmissionWebhook).Error("Unable to create kubernetes config", zap.Error(err))
-		return nil, err
-	}
-	// the webhook and secret writes must be attributed to the admission controller
-	kubeconfig.UserAgent = client.UserAgent(client.UserAgentAdmissionController)
-	// no client side rate limiting: leaving the QPS at 0 would mean the client-go defaults
-	// of 5 QPS / 10 burst
-	kubeconfig.QPS = -1
-	clientset, err := kubernetes.NewForConfig(kubeconfig)
-	if err != nil {
-		log.Log(log.AdmissionWebhook).Error("Unable to create kubernetes clientset", zap.Error(err))
-		return nil, err
-	}
-	return newWebhookManagerImpl(conf, clientset), nil
+	// the webhook and secret writes must be attributed to the admission controller; the
+	// admission controller never populates the scheduler configuration, so the client runs
+	// on the defaults: no client side rate limiting
+	kubeClient := client.NewKubeClientWithUserAgent(conf.GetKubeConfig(), client.UserAgentAdmissionController)
+	return newWebhookManagerImpl(conf, kubeClient.GetClientSet()), nil
 }
 
 func newWebhookManagerImpl(conf *conf.AdmissionControllerConf, clientset kubernetes.Interface) *webhookManagerImpl {
--- a/pkg/admission/webhook_manager.go
+++ b/pkg/admission/webhook_manager.go
@@ -91,6 +91,11 @@ func NewWebhookManager(conf *conf.AdmissionControllerConf) (WebhookManager, erro
 		log.Log(log.AdmissionWebhook).Error("Unable to create kubernetes config", zap.Error(err))
 		return nil, err
 	}
+	// the webhook and secret writes must be attributed to the admission controller
+	kubeconfig.UserAgent = client.UserAgent(client.UserAgentAdmissionController)
+	// no client side rate limiting: leaving the QPS at 0 would mean the client-go defaults
+	// of 5 QPS / 10 burst
+	kubeconfig.QPS = -1
 	clientset, err := kubernetes.NewForConfig(kubeconfig)
 	if err != nil {
 		log.Log(log.AdmissionWebhook).Error("Unable to create kubernetes clientset", zap.Error(err))
--- a/pkg/admission/webhook_manager.go
+++ b/pkg/admission/webhook_manager.go
@@ -85,12 +85,9 @@ type webhookManagerImpl struct {
 }
 
 // NewWebhookManager is used to create a new webhook manager
-func NewWebhookManager(conf *conf.AdmissionControllerConf) (WebhookManager, error) {
-	// the webhook and secret writes must be attributed to the admission controller; the
-	// admission controller never populates the scheduler configuration, so the client runs
-	// on the defaults: no client side rate limiting
-	kubeClient := client.NewKubeClientWithUserAgent(conf.GetKubeConfig(), client.UserAgentAdmissionController)
-	return newWebhookManagerImpl(conf, kubeClient.GetClientSet()), nil
+func NewWebhookManager(conf *conf.AdmissionControllerConf) WebhookManager {
+	kubeClient := client.NewAdmissionControllerKubeClient(conf.GetKubeConfig())
+	return newWebhookManagerImpl(conf, kubeClient.GetClientSet())
 }
 
 func newWebhookManagerImpl(conf *conf.AdmissionControllerConf, clientset kubernetes.Interface) *webhookManagerImpl {
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -1235,7 +1235,7 @@ func (ctx *Context) HandleContainerStateUpdate(request *si.UpdateContainerSchedu
 					Message: request.Reason,
 				}) {
 				events.GetRecorder().Eventf(task.GetTaskPod().DeepCopy(), nil,
-					v1.EventTypeNormal, "PodUnschedulable", "PodUnschedulable",
+					v1.EventTypeWarning, "PodUnschedulable", "PodUnschedulable",
 					"Task %s is skipped from scheduling because the queue quota has been exceed", task.alias)
 			}
 		case si.UpdateContainerSchedulingStateRequest_FAILED:
@@ -1249,7 +1249,7 @@ func (ctx *Context) HandleContainerStateUpdate(request *si.UpdateContainerSchedu
 					Message: request.Reason,
 				}) {
 				events.GetRecorder().Eventf(task.GetTaskPod().DeepCopy(), nil,
-					v1.EventTypeNormal, "PodUnschedulable", "PodUnschedulable",
+					v1.EventTypeWarning, "PodUnschedulable", "PodUnschedulable",
 					"Task %s is pending for the requested resources become available", task.alias)
 			}
 		default:
--- a/pkg/client/apifactory.go
+++ b/pkg/client/apifactory.go
@@ -23,7 +23,6 @@ import (
 
 	"go.uber.org/zap"
 	"k8s.io/client-go/informers"
-	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/feature"
@@ -90,11 +89,17 @@ type APIFactory struct {
 	lock     *locking.RWMutex
 }
 
-// NewAPIFactory creates the clients shared by the shim. The clientset backing informerFactory
-// is passed in so that the namespaced factory created here shares it: both only run informers
-// so they need the same unlimited client and the same attribution.
-func NewAPIFactory(scheduler api.SchedulerAPI, informerClientSet kubernetes.Interface, informerFactory informers.SharedInformerFactory, configs *conf.SchedulerConf, testMode bool) (*APIFactory, error) {
+// NewAPIFactory creates the clients shared by the shim.
+func NewAPIFactory(scheduler api.SchedulerAPI, configs *conf.SchedulerConf, testMode bool) (*APIFactory, error) {
 	kubeClient := NewKubeClient(configs.KubeConfig)
+
+	// all informers, cluster wide and namespaced, share one clientset: both only run
+	// informers so they need the same unlimited client and the same attribution.
+	// resync is disabled (period 0): the shim relies on the watch stream for updates; a
+	// periodic resync only replays the local cache into the handlers, which at scale is
+	// pure no-op churn
+	informerClientSet := NewInformerClientSet(configs.KubeConfig)
+	informerFactory := informers.NewSharedInformerFactory(informerClientSet, 0)
 	namespaceInformerFactory := informers.NewSharedInformerFactoryWithOptions(informerClientSet, 0, informers.WithNamespace(configs.Namespace))
 	// init informers
 	// volume informers are also used to get the Listers for the predicates
--- a/pkg/client/apifactory.go
+++ b/pkg/client/apifactory.go
@@ -23,6 +23,7 @@ import (
 
 	"go.uber.org/zap"
 	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/feature"
@@ -89,9 +90,12 @@ type APIFactory struct {
 	lock     *locking.RWMutex
 }
 
-func NewAPIFactory(scheduler api.SchedulerAPI, informerFactory informers.SharedInformerFactory, configs *conf.SchedulerConf, testMode bool) (*APIFactory, error) {
+// NewAPIFactory creates the clients shared by the shim. The clientset backing informerFactory
+// is passed in so that the namespaced factory created here shares it: both only run informers
+// so they need the same unlimited client and the same attribution.
+func NewAPIFactory(scheduler api.SchedulerAPI, informerClientSet kubernetes.Interface, informerFactory informers.SharedInformerFactory, configs *conf.SchedulerConf, testMode bool) (*APIFactory, error) {
 	kubeClient := NewKubeClient(configs.KubeConfig)
-	namespaceInformerFactory := informers.NewSharedInformerFactoryWithOptions(kubeClient.GetClientSet(), 0, informers.WithNamespace(configs.Namespace))
+	namespaceInformerFactory := informers.NewSharedInformerFactoryWithOptions(informerClientSet, 0, informers.WithNamespace(configs.Namespace))
 	// init informers
 	// volume informers are also used to get the Listers for the predicates
 	podInformer := informerFactory.Core().V1().Pods()
--- a/pkg/client/apifactory.go
+++ b/pkg/client/apifactory.go
@@ -93,14 +93,11 @@ type APIFactory struct {
 func NewAPIFactory(scheduler api.SchedulerAPI, configs *conf.SchedulerConf, testMode bool) (*APIFactory, error) {
 	kubeClient := NewKubeClient(configs.KubeConfig)
 
-	// all informers, cluster wide and namespaced, share one clientset: both only run
-	// informers so they need the same unlimited client and the same attribution.
 	// resync is disabled (period 0): the shim relies on the watch stream for updates; a
 	// periodic resync only replays the local cache into the handlers, which at scale is
 	// pure no-op churn
-	informerClientSet := NewInformerClientSet(configs.KubeConfig)
-	informerFactory := informers.NewSharedInformerFactory(informerClientSet, 0)
-	namespaceInformerFactory := informers.NewSharedInformerFactoryWithOptions(informerClientSet, 0, informers.WithNamespace(configs.Namespace))
+	informerFactory := informers.NewSharedInformerFactory(kubeClient.GetClientSet(), 0)
+	namespaceInformerFactory := informers.NewSharedInformerFactoryWithOptions(kubeClient.GetClientSet(), 0, informers.WithNamespace(configs.Namespace))
 	// init informers
 	// volume informers are also used to get the Listers for the predicates
 	podInformer := informerFactory.Core().V1().Pods()
--- a/pkg/client/eventsink.go
+++ b/pkg/client/eventsink.go
@@ -1,0 +1,228 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package client
+
+import (
+	"context"
+	"sync/atomic"
+	"time"
+
+	"go.uber.org/zap"
+	eventsv1 "k8s.io/api/events/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/tools/events"
+	"k8s.io/client-go/util/flowcontrol"
+	"k8s.io/utils/clock"
+
+	"github.com/apache/yunikorn-k8shim/pkg/conf"
+	"github.com/apache/yunikorn-k8shim/pkg/log"
+)
+
+const (
+	// the shortest time between two logs of the number of shed events
+	eventShedLogInterval = 30 * time.Second
+	// the longest the events are muted after the server asked us to back off
+	maxEventMute = 60 * time.Second
+	// the mute used when the server throttles without saying for how long
+	defaultEventMute = time.Second
+)
+
+// rateLimitedEventSink sheds events which exceed the configured rate instead of sending them.
+// The broadcaster writes every event from its own goroutine, so a rate limiter on the events
+// client would only pace those writes: the goroutines pile up for as long as a storm lasts.
+// Events are discardable, dropping them here bounds a storm instead of delaying it.
+// The events are also muted for as long as the server asks us to back off: priority and
+// fairness rejects with a 429 which carries the time to wait for.
+type rateLimitedEventSink struct {
+	inner     events.EventSink
+	limiter   flowcontrol.RateLimiter
+	rateLimit string
+	clock     clock.PassiveClock
+	// the delay advertised by the server, recorded by the transport which removes the header
+	hints *muteHintHolder
+
+	// number of events shed since they were last logged
+	shed atomic.Int64
+	// time the shed events were last logged, as unix nanoseconds
+	lastLogged atomic.Int64
+	// time the mute was last logged, as unix nanoseconds
+	lastMuteLogged atomic.Int64
+	// time until which no event is sent, as unix nanoseconds
+	muteUntil atomic.Int64
+}
+
+// NewRateLimitedEventSink wraps an event sink and sheds the events which exceed the given
+// rate. A qps <= 0 disables shedding, every event is passed on to the wrapped sink.
+func NewRateLimitedEventSink(inner events.EventSink, qps, burst int) events.EventSink {
+	return newRateLimitedEventSink(inner, qps, burst, clock.RealClock{}, nil)
+}
+
+// newRateLimitedEventSink allows the clock and the delays recorded by the transport to be
+// passed in for testing
+func newRateLimitedEventSink(inner events.EventSink, qps, burst int, clk clock.PassiveClock, hints *muteHintHolder) *rateLimitedEventSink {
+	limitQPS, limitBurst, rateLimit := rateLimitPolicy(userAgentEvents, qps, burst)
+
+	sink := &rateLimitedEventSink{
+		inner:     inner,
+		rateLimit: rateLimit,
+		clock:     clk,
+		hints:     hints,
+	}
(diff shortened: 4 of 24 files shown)


pkg/admission/webhook_manager.go
```diff
@@ -91,6 +91,11 @@ func NewWebhookManager(conf *conf.AdmissionControllerConf) (WebhookManager, erro
 		log.Log(log.AdmissionWebhook).Error("Unable to create kubernetes config", zap.Error(err))
 		return nil, err
 	}
+	// the webhook and secret writes must be attributed to the admission controller
+	kubeconfig.UserAgent = client.UserAgent(client.UserAgentAdmissionController)
+	// no client side rate limiting: leaving the QPS at 0 would mean the client-go defaults
+	// of 5 QPS / 10 burst
+	kubeconfig.QPS = -1
```
manirajv06: Can we use newly introduced constructors in `kubeclient.go` to create the `clientset` even here as well? If not available, can we create one so that all client creation happens in single place.

manirajv06: I can see one available for this and used already in https://github.com/apache/yunikorn-k8shim/pull/1060/changes#diff-d757b0e6523c6adc5df3a6423e547aca3497226a451c9fe4bd07ec7051b7b508R63

pkg/shim/scheduler.go
```diff
@@ -64,12 +64,12 @@ var (
 )
 
 func NewShimScheduler(scheduler api.SchedulerAPI, configs *conf.SchedulerConf, bootstrapConfigMaps []*v1.ConfigMap) *KubernetesShim {
-	kubeClient := client.NewKubeClient(configs.KubeConfig)
-
+	// all informers, cluster wide and namespaced, share one client
+	informerClientSet := client.NewInformerClientSet(configs.KubeConfig)
 	// we have disabled re-sync to keep ourselves up-to-date
-	informerFactory := informers.NewSharedInformerFactory(kubeClient.GetClientSet(), 0)
+	informerFactory := informers.NewSharedInformerFactory(informerClientSet, 0)
```
manirajv06: Can we move this into `NewAPIFactory` as well?

That way we need to pass only `configs` to `NewAPIFactory`

pkg/client/apifactory.go
```diff
@@ -89,9 +90,12 @@ type APIFactory struct {
 	lock     *locking.RWMutex
 }
 
-func NewAPIFactory(scheduler api.SchedulerAPI, informerFactory informers.SharedInformerFactory, configs *conf.SchedulerConf, testMode bool) (*APIFactory, error) {
+// NewAPIFactory creates the clients shared by the shim. The clientset backing informerFactory
+// is passed in so that the namespaced factory created here shares it: both only run informers
+// so they need the same unlimited client and the same attribution.
+func NewAPIFactory(scheduler api.SchedulerAPI, informerClientSet kubernetes.Interface, informerFactory informers.SharedInformerFactory, configs *conf.SchedulerConf, testMode bool) (*APIFactory, error) {
 	kubeClient := NewKubeClient(configs.KubeConfig)
```
manirajv06, line 94: Where is this being used? Looks like for volume binding and other operations too. Can you take a look at this? Should we re-use the one meant for `writes` instead of this one?

manirajv06: > Should we re-use the one meant for `writes` instead of this one?

I see the changes in `interfaces.go`. Yes, we are using the one meant for `writes` only.

However, I can see that this has been passed to `Clients` and being used to load config maps in `Context.go#LoadConfigMaps`. Is this correct? Are we supposed to use the meant for `reads`? Do we need to use the one being used here  https://github.com/apache/yunikorn-k8shim/pull/1060/changes#diff-49e2cf379c065c43983fe806d7739d9a0947f7e21f22f29f22d905cc34d1bb39R32?

pkg/conf/schedulerconf.go
```diff
@@ -91,8 +93,10 @@ const (
 	DefaultOperatorPlugins                 = "general"
 	DefaultDisableGangScheduling           = false
 	DefaultEnableConfigHotRefresh          = true
-	DefaultKubeQPS                         = 1000
-	DefaultKubeBurst                       = 1000
+	DefaultKubeQPS                         = 0 // client side write limiting is opt-in: <= 0 means no limiter
```
manirajv06: '0' would mean the client-go defaults of 5 QPS / 10 burst not really "no limiter".  Is my understanding correct? Should we make this as -1 if we don't want rate limiter at all for all except for events?

manirajv06 (changes requested):
I reviewed the code for all clients except for `events`. Will look at that too.

pkg/client/eventsink.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// muted returns true while the server asked us to stop sending events
+func (s *rateLimitedEventSink) muted() bool {
+	return s.clock.Now().UnixNano() < s.muteUntil.Load()
+}
+
+// logShedEvents logs the number of events shed since the last time they were logged, at most
+// once every eventShedLogInterval. The first drop is always logged.
+func (s *rateLimitedEventSink) logShedEvents(muted bool) {
```
manirajv06, line 218: We do already have `NewRateLimitedLogger` to make our life easier in core. Probably, we need one for shim as well

pkg/admission/webhook_manager.go
tigerquoll: Agreed, that's an inconsistency this PR introduced — will fix.

pkg/client/eventsink.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// checkThrottled mutes the events for as long as the server asked us to back off. Priority
+// and fairness rejects with a 429 and advertises a delay which grows while it keeps dropping
+// requests, following it is cheaper than having every event rejected.
+func (s *rateLimitedEventSink) checkThrottled(err error) {
+	if err == nil || !errors.IsTooManyRequests(err) {
+		return
+	}
+	delay, source := s.muteDelay(err)
```
manirajv06, line 154: Since above `errors.IsTooManyRequests` ensures return except for 429, I don't see any value adde because of  `muteDelay`. We need only value (seconds) stored in hint which can be directly fetched from the hint itself and use the same below. I think `muteDelay` can be removed unless there are other compelling reason.

pkg/shim/scheduler.go
tigerquoll: Nice idea - that will make things even tidier. Will do.

manirajv06 (changes requested):

pkg/client/apifactory.go
tigerquoll: I think the writes client is the right one here, and it's what master already does: this PR doesn't touch LoadConfigMaps, it just gives the client it uses a name and a policy. There's no "reads" client to switch to - the informers clientset exists only to back the informer factories and isn't exposed in Clients, deliberately, so that list/watch traffic stays isolated and attributable. The bootstrap client wouldn't be right either: it exists to run before the configuration is loaded, so it builds from the default kubeconfig path rather than the configured one, and its user agent is meant to mark startup traffic only.

So the writes client carries the must-complete, direct request/response traffic. The name is perhaps a little loose: apart from the configmap loads - two GETs, once per process, at registration — everything on it is either a mutation (pods, plus PVCs via the volume binder) or a read in service of one, like the pod GETs inside the update retry loops. How about I add a comment on Clients spelling out what each client carries?

pkg/conf/schedulerconf.go
tigerquoll: rateLimitPolicy translates anything <= 0 into QPS = -1  so we don't get the 5/10 defaults, but your query has highlighted that this is not exactly clear.  Setting directly to -1 will make things clearer - I will do that.

pkg/client/eventsink.go
tigerquoll: The shed log can't use a rate-limited logger, unfortunately - it needs to aggregate, not just throttle. Every dropped event increments a counter, and the call that claims the interval logs the accumulated total (`shed.Swap(0)`), so the "`9,805 events shed`" line in the description is exact, not sampled.

pkg/client/eventsink.go
tigerquoll: 1. Server rejects with 429 + Retry-After: 31.
2. Transport records 31 in the hint holder, deletes the header, passes the 429 up.
3. client-go's retry layer consults checkWait: status is 429, but retryAfterSeconds(resp) now finds nothing → no internal sleep-retry. This is the only thing the strip changes -verified in with_retry.go earlier: client-go only wait-retries a 429/5xx that tells it how long to wait.
4. client-go transforms the 429 into a StatusError carrying code 429 - the status code was never touched - and returns it as the error from inner.Create.
5. The sink's checkThrottled(err) fires, and IsTooManyRequests(err) is how it recognises what just happened.

So we move the hint out of the header (but keep the status code), to prevent the go-client automatically retrying, but still return the 429 status code to the caller (The event sink) to allow the appropriate action to trigger.

pkg/shim/scheduler.go
tigerquoll: Nice idea - that made things even tidier. Done in the latest push: both informer arguments are gone, the informer clientset and both factories are created inside `NewAPIFactory`, and the signature keeps just `scheduler` and `testMode` alongside `configs`.

pkg/client/apifactory.go
manirajv06: I am trying to introduce a new concern "configmaps" or something similar to handle the config map GET calls. 
With this, we could have 4 concerns 1. writes 2. informers 3. events 4. configmaps or reads. This new client could be separated out from "writes" client operations which should be meant only to handle create, update, delete of pods/pvcs. New client meant for "configmaps" concern should handle these places 1. Two GET calls during bootstrap 2. Two GET calls during registration. That way, all reads are routed through this new client and could be configured different than the writes. Currently, we might have these 4 GET calls but new cases for reads in future also could use this new client.

BTW, I think interfaces#GetConfigs(), interfaces#Get() can be removed as I don't see any usage in production code. If required in mocks only, then methods could be called directly without these wrappers.

manirajv06 (changes requested):

pkg/client/apifactory.go
tigerquoll: I think the value here is the distinguishing the "must-complete, direct request/response traffic" from anything else.  Splitting that out further to read v write doesn't gain us anything - functionality they would have the same priority and traffic profile, and would be treated the same.

manirajv06 (approved):
looks good

manirajv06:
@pbacsko @wilfred-s I am good. Do you want to take a look? Otherwise, I can take it forward.

pkg/admission/webhook_manager.go
```diff
@@ -86,17 +86,11 @@ type webhookManagerImpl struct {
...
-		return nil, err
-	}
-	clientset, err := kubernetes.NewForConfig(kubeconfig)
-	if err != nil {
-		log.Log(log.AdmissionWebhook).Error("Unable to create kubernetes clientset", zap.Error(err))
-		return nil, err
-	}
-	return newWebhookManagerImpl(conf, clientset), nil
+	// the webhook and secret writes must be attributed to the admission controller; the
+	// admission controller never populates the scheduler configuration, so the client runs
```
wilfred-s: Incorrect it reads the configmap and loads the settings from `main()`. It runs with the same QPS and burst as the scheduler. 1000/1000 which is too high

pkg/admission/webhook_manager.go
```diff
@@ -86,17 +86,11 @@ type webhookManagerImpl struct {
...
+	// on the defaults: no client side rate limiting
```
wilfred-s: The webhook should run with the default 5/10 or slightly higher which should be more than enough. (3 informers and some adhoc connections on startup and cert change)
All admission requests are REST calls made from the API server to the controller and do not use a kubeclient using unlimited is not OK

pkg/client/apifactory.go
```diff
@@ -89,9 +89,18 @@ type APIFactory struct {
...
 }
 
-func NewAPIFactory(scheduler api.SchedulerAPI, informerFactory informers.SharedInformerFactory, configs *conf.SchedulerConf, testMode bool) (*APIFactory, error) {
+// NewAPIFactory creates the clients shared by the shim.
+func NewAPIFactory(scheduler api.SchedulerAPI, configs *conf.SchedulerConf, testMode bool) (*APIFactory, error) {
 	kubeClient := NewKubeClient(configs.KubeConfig)
-	namespaceInformerFactory := informers.NewSharedInformerFactoryWithOptions(kubeClient.GetClientSet(), 0, informers.WithNamespace(configs.Namespace))
+
+	// all informers, cluster wide and namespaced, share one clientset: both only run
+	// informers so they need the same unlimited client and the same attribution.
```
wilfred-s: That is not needed and also not correct. Informers do not do cycle through connections or create new write/read requests on an ongoing basis. They only do a single list on startup (1 connection) and then have a single watch on the object (1 connection). The list and watch are successive and do not run at the same time.
Watch is a push from the API on a long lived single API connection. There are ~20 informers so we do not need unlimited. This client can be easily shared with the normal scheduler actions without impact.

pkg/client/apifactory.go
wilfred-s: There should only be one split in the k8shim: events and scheduling actions. The admission controller gets it own client already.

pkg/client/eventsink.go
wilfred-s, line 1: I am not for dropping random events. Not all events are the same. Some you really want to get pushed others are not that important. Limiting at the point of creation for events that are of low impact is a better solution.
Even setting a verbosity level for events similar to the log level would be acceptable.

pkg/client/kubeclient.go
```diff
@@ -30,44 +30,127 @@ import (
...
+	userAgentBootstrap = "yunikorn-bootstrap"
+	userAgentWrites    = "yunikorn-scheduler/writes"
+	userAgentInformers = "yunikorn-scheduler/informers"
+	userAgentEvents    = "yunikorn-scheduler/events"
+)
+
+// UserAgent appends the build version to the concern, e.g. "yunikorn-scheduler/writes
+// (1.7.0)". The version is only set in release builds, it is left out when empty.
+func UserAgent(concern string) string {
+	return userAgentWithVersion(concern, conf.GetBuildInfoMap()["buildVersion"])
```
wilfred-s: Not sure if we want/should add build version to the agent string. It is not required from a debugging perspective.

pkg/client/kubeclient.go
```diff
@@ -30,44 +30,127 @@ import (
...
 
-func newSchedulerKubeClient(kc string) SchedulerKubeClient {
-	schedulerConf := conf.GetSchedulerConf()
+// rateLimitPolicy turns the configured qps and burst into the values to set on a REST config
+// and a description of the resulting limit for logging.
+// A qps <= 0 disables client side rate limiting: this is expressed as a negative QPS, which
+// client-go treats as "create no rate limiter", while a QPS of 0 silently falls back to its
+// own defaults of 5 QPS / 10 burst (see RESTClientFor and NewForConfigAndClient in
+// client-go). A QPS set without a burst is rejected by client-go, so the burst defaults to
+// the configured qps.
```
wilfred-s: We should leave this to the K8s go-client and allow a QPS and burst set to 0 to use the defaults

Much simpler solution for the value check would be:
```
config.QPS = float32(max(-1, schedulerConf.KubeQPS))
config.Burst = max(max(0, schedulerConf.KubeBurst), schedulerConf.KubeQPS)
```
They come back from the config as int32 values

wilfred-s (changes requested):

pkg/admission/webhook_manager.go
tigerquoll: You're right on the outcome, and the comment was misleading either way, so it's gone.

For the record the values didn't come from the configmap: nothing in the admission controller calls `conf.UpdateConfigMaps`, so `NewKubeClient` read the *defaults* of the scheduler configuration — 1000/1000 for the informer client in `main.go` — while the webhook client, built straight from `CreateRestConfig`, ran on client-go's 5/10. Both now run on the client-go defaults, see the next thread.

pkg/admission/webhook_manager.go
tigerquoll: Done. Both admission controller clients — the informers in `main.go` and the webhook/secret writes in `NewWebhookManager` — now leave QPS and burst at 0, i.e. client-go's 5/10, through `client.NewAdmissionControllerKubeClient`. The bootstrap client is on the same defaults, which is what it had on master.

pkg/client/apifactory.go
tigerquoll: Agreed. With the events off the shared bucket, the relist-versus-event-storm coupling that the informer client was meant to break is gone, and a separate clientset only bought a distinct user agent in the audit log. The informer clientset is removed; both informer factories are back on the scheduler client.

pkg/client/apifactory.go
tigerquoll: Done — the layout is now the two clients you describe: `yunikorn-scheduler` for the scheduling actions and the informers, and `yunikorn-scheduler/events` for the events, plus the pre-existing bootstrap and admission controller clients, both on client-go defaults.

pkg/client/eventsink.go
tigerquoll: Agreed that not all events are the same; the bucket approach was type blind - moved to a new prioritisation system:

- **Warning events are never shed by the bucket.** `FailedScheduling`, `TaskRejected`, `NodeRejected`, `ApplicationFailed`, the gang scheduling failures — everything that explains why a pod is not running — always goes out and does not consume a token.
- **`PodUnschedulable` becomes a Warning.** It was typed Normal, which would have put it in the sheddable class; kube-scheduler's equivalent, `FailedScheduling`, is a Warning too.
- **Only Normal events are shed above `eventQPS`.** Counting what the shim emits, the volume is three Normal events per *successfully* scheduled pod (`Scheduling`, `Scheduled`, `PodBindSuccessful`) plus the core's `Informational` records. At the ~1,900 binds/s measured on the KWOK rig that is ~5,700 events/s saying "this worked", and that is what a storm sheds.
- **`kubernetes.eventLevel`** (`normal` | `warning`, default `normal`) is the verbosity knob you suggested. It is applied at the recorder, so a suppressed event is never copied, cached or written, and it is hot-reloadable like the log level.

The server-side mute still applies to every type. Once APF is rejecting events, client-go drops them itself after its Retry-After retries (`recordEvent` treats every `StatusError` as "Server rejected event (will not retry!)"), so muting only saves the round trips and the parked goroutines; it does not lose an event that would otherwise have landed.

pkg/client/kubeclient.go
tigerquoll: Removed; the user agents are the bare concern strings.

pkg/client/kubeclient.go
tigerquoll: Done. `rateLimitPolicy` and its warnings are gone in favour of the normalisation you suggested, with one difference: burst is the configured burst when set, otherwise the qps - `max(burst, qps)` would silently raise an explicit `burst < qps`. `0/0` is left to client-go's 5/10 as you asked, so `qps: "0"` now behaves exactly as it did before this PR; the defaults stay at `-1`. The compatibility table in the description is updated to match.

One thing I'd like to confirm with you: the scheduler client itself still defaults to no client-side limiter (`-1`), on the measurements in the description - the 1000/1000 bucket clipped the bind burst, and at `qps: 50` the limiter was the throughput ceiling at 52 binds/s. If you'd rather keep a default cap on the scheduler client, say so and I'll set one.

tigerquoll:
@wilfred-s @manirajv06 review comments addressed in 628fe320 — the layout is now the two clients (scheduler + events), admission controller and bootstrap on client-go defaults, QPS/burst left to client-go, no version in the user agent, and event shedding is type-aware (Warning events are never shed, `PodUnschedulable` retyped to Warning, new `kubernetes.eventLevel`). Replies on each thread; description updated. PTAL.

pkg/conf/schedulerconf.go
```diff
@@ -97,7 +98,14 @@ const (
...
 	DefaultKubeEventQPS                    = 200 // events are discardable and limited by default: <= 0 means no limiter
 	DefaultKubeEventBurst                  = 400
+	DefaultKubeEventLevel                  = EventLevelNormal
 	DefaultAMFilteringGenerateUniqueAppIds = false
+
+	// event levels, the value of CMKubeEventLevel: all events, the Warning events only, or
+	// no events at all
+	EventLevelNormal  = "normal"
+	EventLevelWarning = "warning"
+	EventLevelNone    = "none"
```
manirajv06: Does K8s has "none" level? I can see only the first two. 1. Normal 2. Warning. Do we need this?

pkg/cache/context.go
```diff
@@ -1249,7 +1249,7 @@ func (ctx *Context) HandleContainerStateUpdate(request *si.UpdateContainerSchedu
 					Message: request.Reason,
 				}) {
 				events.GetRecorder().Eventf(task.GetTaskPod().DeepCopy(), nil,
-					v1.EventTypeNormal, "PodUnschedulable", "PodUnschedulable",
+					v1.EventTypeWarning, "PodUnschedulable", "PodUnschedulable",
```
manirajv06, line 1252: Can we do cross check to ensure correct event type has been used across all places?

For ex, task#beforeTaskFail uses normal.

manirajv06 (changes requested):

pkg/conf/schedulerconf.go
tigerquoll: Got mixed up with k8s audit level.  Will remove.

pkg/cache/context.go
tigerquoll: Done, every `Eventf` in `pkg/cache` cross-checked against the Kubernetes semantics (Normal = expected progress, Warning = something the user needs to look at):

- `TaskFailed` (`beforeTaskFail`) retyped Normal → Warning. Both paths into it (`postTaskRejected`, `failWithEvent`) already emit a Warning for the cause; the terminal notification now matches.
- `PodUnschedulable` is Warning at both emit sites (`HandleContainerStateUpdate` SKIPPED and FAILED) and nothing else emits that reason; a unit test now asserts the type for both, and for `TaskFailed`.
- Kept as Normal: `Scheduling`, `Scheduled`, `PodBindSuccessful`, `TaskCompleted`, `NodeAccepted`, `NodeDeleted`, and the gang progress events (`TaskGroupMatch`, `CreatingPlaceholders`, `PlaceholderAllocated`, `GangReservationComplete`). Already Warning: `FailedScheduling`, `TaskRejected`, `NodeRejected`, `ApplicationFailed`, `TaskGroupsError`, `GangSchedulingFailed`, the duplicate-metadata `Scheduling` warning, and everything routed through `failWithEvent`.
- Out of scope here: the core records forwarded by `PublishEvents` are all published as Normal `Informational` regardless of what the core is reporting (e.g. placeholder timeouts). Pre-existing and needs an SI-level change to carry a type, so I have left it for a follow-up.

tigerquoll:
@manirajv06 both threads addressed in 62c82bda: the `none` event level is gone and `TaskFailed` is retyped to Warning, with the full cross-check in the thread. PTAL.

PR #1061, opened 2026-08-11
[YUNIKORN-3357] Add goroutine leak detection to tests via goleak

tigerquoll:
### What is this PR for?

Shim half of YUNIKORN-3357 (core PR: apache/yunikorn-core#1124). Adds goroutine leak detection via [uber-go/goleak](https://github.com/uber-go/goleak) to all 18 test packages under `pkg/`, through a shared `pkg/common/leakcheck` helper mirroring the core design.

- **Scoped exemptions**: only the genuinely multi-package entry (DRA resource-slice tracker) is shared; the 17 `pkg/shim`-specific entries are passed only from that package's `TestMain` via `leakcheck.ShimSchedulerOptions()`, so the other 17 packages run with a 1-entry list and the check binds tightly there
- **No shim-owned exemptions remain.** All 17 `pkg/shim` entries are inherited from yunikorn-core; nothing this repo leaks is carved out
- **Two-line test fix included**: `defer apiProvider.Stop()` at two `pkg/cache` call sites, which deletes the exemption those tests caused
- `test/e2e` ginkgo suites deliberately not instrumented (live-cluster suites that leave informers running by design; documented in the package doc)
- No production code modified; goleak was already a transitive dependency (`go.sum` unchanged) and becomes a direct test-only one

### Findings surfaced (separate JIRAs)

Suspected production defects, all reproduced and documented in the exemption comments. **The burn-down is complete on the shim side**: all four shim-owned exemptions this PR opened with are now gone, because the defects they documented were fixed.

1. ~~`KubernetesShim.Stop()` stops only one of the two `doScheduling` loops~~ — **fixed on master** by YUNIKORN-3367 (#1074); exemption removed
2. ~~`Stop()` after a failed `Run()` is a no-op~~ — **fixed on master** by the same change (YUNIKORN-3368, duplicate of 3367); exemption removed
3. ~~`AsyncRMCallback.UpdateAllocation`'s AssumePod retry (30-step backoff) runs on the RM proxy event loop with no context/stop channel~~ — **fixed on master** by YUNIKORN-3369 (#1078); exemption removed in this PR
4. yunikorn-core services are not restartable in-process (`EventSystemImpl.Stop()` is one-way), so `MockScheduler.stop()` cannot call `coreContext.StopAll()` without introducing an intermittent failure at `-count>1`. The 17 inherited core-service exemptions burn down after a core-side fix (YUNIKORN-3370)

One of those 17, the YUNIKORN-3365 shutdown wedge, is the only intermittent entry: it did not reproduce in five consecutive runs against the current core pin. It is kept, and now says so in the comment, because 3365 is open and a leak that surfaces once in CI costs more than an entry that costs nothing.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-3357

Rebased onto master (`fa7b6add`) and re-verified.

**CI on this PR is red for a reason that is not this PR: master itself does not compile.** The YUNIKORN-3435 SI bump (`a2222d65`) added `PreFilterPredicates` to `api.ResourceManagerCallback` before #1043 implements it in the shim, so `pkg/cache` fails to build on master too — see master's own run [33974617433](https://github.com/apache/yunikorn-k8shim/actions/runs/33974617433). The results below were produced with a local throwaway stub of that one method (not committed, and not part of the diff), mirroring core's own mock callback:

- `go test ./pkg/... -race -tags deadlock -count=1`: all 18 instrumented packages ok, no data races
- `go test ./pkg/shim/ -tags deadlock -count=3`: ok (the racy package)
- `golangci-lint run`, `make license-check`, `go vet`, `gofmt`, `go mod tidy`: clean
- Every exemption re-probed against the current core pin by deleting it and confirming the result: the YUNIKORN-3369 entry now passes without it (hence its removal), and every remaining entry still fires except the intermittent 3365 one noted above
- Ground truth cross-checked by stripping the whole list and reading goleak's report: no core frame was renamed by the dependency bump this time

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,6 +30,7 @@ require (
 	github.com/onsi/gomega v1.40.0
 	github.com/prometheus/client_golang v1.24.1
 	github.com/sasha-s/go-deadlock v0.3.9
+	go.uber.org/goleak v1.3.0
 	go.uber.org/zap v1.28.0
 	go.yaml.in/yaml/v3 v3.0.5
 	gotest.tools/v3 v3.5.2
--- a/pkg/admission/conf/main_test.go
+++ b/pkg/admission/conf/main_test.go
@@ -1,0 +1,30 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package conf
+
+import (
+	"testing"
+
+	"github.com/apache/yunikorn-k8shim/pkg/common/leakcheck"
+)
+
+// TestMain fails the package if a test leaks a goroutine.
+func TestMain(m *testing.M) {
+	leakcheck.VerifyTestMain(m)
+}
--- a/pkg/admission/main_test.go
+++ b/pkg/admission/main_test.go
@@ -1,0 +1,30 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package admission
+
+import (
+	"testing"
+
+	"github.com/apache/yunikorn-k8shim/pkg/common/leakcheck"
+)
+
+// TestMain fails the package if a test leaks a goroutine.
+func TestMain(m *testing.M) {
+	leakcheck.VerifyTestMain(m)
+}
--- a/pkg/admission/metadata/main_test.go
+++ b/pkg/admission/metadata/main_test.go
@@ -1,0 +1,30 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package metadata
+
+import (
+	"testing"
+
+	"github.com/apache/yunikorn-k8shim/pkg/common/leakcheck"
+)
+
+// TestMain fails the package if a test leaks a goroutine.
+func TestMain(m *testing.M) {
+	leakcheck.VerifyTestMain(m)
+}
--- a/pkg/admission/pki/main_test.go
+++ b/pkg/admission/pki/main_test.go
@@ -1,0 +1,30 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package pki
+
+import (
+	"testing"
+
+	"github.com/apache/yunikorn-k8shim/pkg/common/leakcheck"
+)
+
+// TestMain fails the package if a test leaks a goroutine.
+func TestMain(m *testing.M) {
+	leakcheck.VerifyTestMain(m)
+}
--- a/pkg/cache/context_test.go
+++ b/pkg/cache/context_test.go
@@ -1886,6 +1886,7 @@ func TestGetExistingAllocation(t *testing.T) {
 func TestInitializeState(t *testing.T) {
 	context, apiProvider := initContextAndAPIProviderForTest()
 	apiProvider.RunEventHandler()
+	defer apiProvider.Stop()
 	pcLister, ok := apiProvider.GetAPIs().PriorityClassInformer.Lister().(*test.MockPriorityClassLister)
 	assert.Assert(t, ok, "unable to get mock priority class lister")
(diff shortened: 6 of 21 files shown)


chenyulin0719 (approved):
+1 LGTM, same reason in the core PR. https://github.com/apache/yunikorn-core/pull/1124#pullrequestreview-4909965796

manirajv06:
I prefer to do this as "one time" activity using appropriate tools as this is something not required to run every time as we don't introduce leaks in every change that we make. I would prefer running at the end of the "code commit" process like codecov etc to ensure there is no alarming signal, probably somewhere after or before the e2e tests or somewhere through a new option in `make` command. I had come across https://www.gopherguides.com/articles/golang-goroutine-leak-profile, a new feature introduced in go1.2.7 to find the leaks. This can be hooked up into anywhere in the build process.

Please share your thoughts.

tigerquoll:
> I prefer to do this as "one time" activity using appropriate tools as this is something not required to run every time as we don't introduce leaks in every change that we make. I would prefer running at the end of the "code commit" process like codecov etc to ensure there is no alarming signal, probably somewhere after or before the e2e tests or somewhere through a new option in `make` command. I had come across https://www.gopherguides.com/articles/golang-goroutine-leak-profile, a new feature introduced in go1.2.7 to find the leaks. This can be hooked up into anywhere in the build process.
> 
> Please share your thoughts.

The pprof leak profile and goleak actually detect different classes of leak, so I'd like to use both rather than pick one. The runtime profile only reports goroutines blocked on sync primitives that no runnable goroutine can reach - by design it doesn't flag running loops, sleeping goroutines, or goroutines blocked on reachable channels (the article's own limitations section). All four of the production defects this PR surfaced are in that second category (an unstopped wait.Until loop, event loops on package-singleton channels, a timer-backoff retry), so a build-stage profile pass would have reported nothing here. Goleak's test-exit check is what catches lifecycle bugs like these, and it attributes the failure to the exact package and PR that introduces it, at roughly zero cost while green.

Where the profile is genuinely stronger is live processes - which is exactly the gap this PR documents: the e2e suites are deliberately not goleak-instrumented. We're on Go 1.26, so it's available behind GOEXPERIMENT=goroutineleakprofile (per the article, on by default from 1.27) - the one wrinkle is that's a build-time flag, so the scheduler image used in e2e needs building with it until we're on 1.27. I'd be happy to file a follow-up to hook /debug/pprof/goroutineleak into the e2e/nightly stage as you describe - the combination covers both leak classes at both detection points: goleak catches unstopped-service bugs at the PR that introduces them, the runtime profile catches provably-stuck goroutines in real integrated runs.

wilfred-s:
A number of the carveouts will also not be fixed anytime soon. Some of them are in the base K8s code. K8s runs go routine leak tests regularly and fixes what shows up and what the deem to be a leak. The fact that the mentioned leaks are there means they are not considered fix worthy or cannot be fixed without backwards compatibility impact. 

I would also prefer to fix the ones we can fix easily as part of this test introduction. There should not be any exception from our code that could be fixed.

manirajv06:
> I'd be happy to file a follow-up to

Ok. A more improvised version of detecting the leaks from the same folks https://www.uber.com/us/en/blog/leakprof-featherlight-in-production-goroutine-leak-detection/ had come much later after the goleak project. Also, https://rednafi.com/shards/2026/06/go-goroutine-leak-profile/ is a good read. Please take a look. It should help us in doing the follow ups. 

As long as leaks found by goleak is not going to be subset of the list found by leakprofile, we are good. Otherwise, we would need only one way of doing this.

tigerquoll:
> I prefer to do this as "one time" activity using appropriate tools as this is something not required to run every time as we don't introduce leaks in every change that we make. I would prefer running at the end of the "code commit" process like codecov etc to ensure there is no alarming signal, probably somewhere after or before the e2e tests or somewhere through a new option in `make` command. I had come across https://www.gopherguides.com/articles/golang-goroutine-leak-profile, a new feature introduced in go1.2.7 to find the leaks. This can be hooked up into anywhere in the build process.
> 
> Please share your thoughts.

My preference would to keep the noted issues as a burndown chart,  moving them to a number of JIRAs, and have them available to work through

tigerquoll:
I've registered the issues as JIRAs, and change the code documentation to simply point to the JIRAs.

manirajv06 (changes requested):
I can see the new jira's being filed for these issues. Thanks.

I can see your changes on [Core PR](https://github.com/apache/yunikorn-core/pull/1124) based on the review comments. Can we start incorporating the same for this PR as well?

tigerquoll:
> I can see the new jira's being filed for these issues. Thanks.
> 
> I can see your changes on [Core PR](https://github.com/apache/yunikorn-core/pull/1124) based on the review comments. Can we start incorporating the same for this PR as well?

@manirajv06  The mentioned JIRA covers both Core and k8Shim - the "testing discovered" links point to various single issue JIRAs that are scoped to the relevant components (core/k8shim).  The PR has already been updated to strip explanatory comments from the PR and place them in these connected JIRAs

tigerquoll:
Rebased onto master and pushed. Two things changed since the last review round.

**Three shim-owned exemptions are gone.** YUNIKORN-3367 (#1074) landed on 25 Aug — it closes the shim stop channel and always releases the dispatcher and placeholder manager. That kills the exemptions for `PlaceholderManager.Start.func1`, `dispatcher.Start.func1` and `wait.BackoffUntilWithContext` (YUNIKORN-3368 was closed as a duplicate of 3367). Shim-owned carve-outs go from four to one; the survivor is YUNIKORN-3369, still open.

@wilfred-s — this is the burn-down, one cycle in. The exemptions this PR opened with are being deleted by the fixes they documented, and the deletion is mechanical precisely because each entry names its JIRA.

**Two core exemptions had silently stopped matching.** The yunikorn-core bump on master renamed `EventSystemImpl.StartServiceWithPublisher.func2` to `.func1` and unexported the publisher (`EventPublisher.StartService` → `eventPublisher.start`), so both entries had become no-ops and `pkg/shim` was red. Realigned, and I recorded a caveat in `leakcheck.go` that a dependency bump can rename a frame out from under an exemption — now demonstrated rather than hypothetical.

I also re-probed every remaining exemption by deleting it and confirming the check then fails, so none is dead weight: the DRA tracker entry (YUNIKORN-3371), the AssumePod entry (YUNIKORN-3369), and all 17 inherited core-service entries.

Verification on the rebased branch:

- `go test ./pkg/... -tags deadlock`: 18/18 ok
- `go test ./pkg/... -race -tags deadlock`: 18/18 ok, no data races
- `go test ./pkg/shim/ -tags deadlock -count=3`: ok
- golangci-lint, `make license-check`, `go vet`, `gofmt`: clean
- `go mod tidy`: no-op, so no `go.sum` change is needed on the shim side

@manirajv06 — your review is still marked changes-requested. On mirroring the core PR: as in my earlier reply, YUNIKORN-3357 spans both repos and the per-component defects live in their own linked JIRAs, and the explanatory essays have already been stripped from both PRs in favour of `See YUNIKORN-####` references. Is there something further you'd like changed here, or can this go back for another look?

tigerquoll:
Rebased onto master (`fa7b6add`) and pushed. One thing changed since the last round, and one thing about master needs flagging.

**The shim-side burn-down is now complete.** YUNIKORN-3369 (#1078) landed on 1 Sep and aborts the AssumePod retry loop on shutdown, which retires the last exemption whose cause was in shim code. This PR opened with four shim-owned carve-outs; 3367 removed three and 3369 removes the last. `shimOwnedOptions()` is now empty, so a new shim-owned leak fails the build instead of getting an entry.

@wilfred-s — this answers your point directly: *"There should not be any exception from our code that could be fixed."* There now isn't one. All 17 remaining `pkg/shim` entries come from yunikorn-core service goroutines the shim starts in-process and cannot stop in tests, and they burn down with YUNIKORN-3370.

@manirajv06 — the JIRA-referencing rework you asked for is in and has now had two cycles of evidence behind it: the exemptions the comments pointed at are the ones that actually got fixed.

**Master does not currently compile,** which is why CI here is red. The YUNIKORN-3435 SI bump (`a2222d65`) added `PreFilterPredicates` to `api.ResourceManagerCallback` before #1043 implements it in the shim:

```
pkg/cache/scheduler_callback.go:44:37: cannot use &AsyncRMCallback{} ... does not implement
api.ResourceManagerCallback (missing method PreFilterPredicates)
```

That reproduces on a clean checkout of master, and master's own CI has failed on it since 4 Sep — run [33974617433](https://github.com/apache/yunikorn-k8shim/actions/runs/33974617433). Nothing in this PR touches production code, so this PR cannot be the cause. I verified the change locally with a throwaway stub of that one method (not committed, not in the diff): all 18 instrumented packages green under `-race -tags deadlock`, plus `golangci-lint`, `go vet`, `gofmt`, `make license-check` and `go mod tidy` clean.

While re-verifying I also re-probed every exemption against the current core pin by deleting it and checking the result. Two things worth recording: no core frame was renamed by this dependency bump (unlike the previous one), and the YUNIKORN-3365 entry is the only one that no longer reproduces — five consecutive runs without it were green. I have kept it, and said so in the comment, because 3365 is still open and the wedge is a shutdown race; a leak that surfaces once in CI costs more than an entry that costs nothing. Happy to drop it if you would rather the list carry only entries that fire on demand.

PR #1062, opened 2026-08-12
[YUNIKORN-2884] handling volume and pod binding failure retries

adityadtu5:
### Description
- Update yunikorn-core version for new tidwall btree changes
- Added rollback and retry for both pod binding and volume binding errors
- Removed some redundant code in context.go

### Type of change

- [ ] Bug Fix
- [x] Improvement
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-2884

- [ ] I have created a Jira issue for this pull request.

### AI Tooling
If an AI tool was used:

### How has this been tested?

### Questions:

### DIFF
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -42,7 +42,6 @@ import (
 	fwk "k8s.io/kube-scheduler/framework"
 	"k8s.io/kubernetes/pkg/features"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/dynamicresources"
-	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/volumebinding"
 	"k8s.io/kubernetes/pkg/scheduler/util/assumecache"
 
 	schedulercache "github.com/apache/yunikorn-k8shim/pkg/cache/external"
@@ -797,14 +796,6 @@ func (ctx *Context) bindPodVolumes(pod *v1.Pod) error {
 					zap.String("nodeName", assumedPod.Spec.NodeName),
 					zap.Error(err))
 				return err
-			}
-			if volumes.StaticBindings == nil {
-				// convert nil to empty array
-				volumes.StaticBindings = make([]*volumebinding.BindingInfo, 0)
-			}
-			if volumes.DynamicProvisions == nil {
-				// convert nil to empty array
-				volumes.DynamicProvisions = make([]*volumebinding.DynamicProvision, 0)
 			}
 			err = ctx.apiProvider.GetAPIs().VolumeBinder.BindPodVolumes(context.Background(), assumedPod, volumes)
 			if err != nil {
--- a/pkg/cache/context.go
+++ b/pkg/cache/context.go
@@ -42,6 +42,7 @@ import (
 	fwk "k8s.io/kube-scheduler/framework"
 	"k8s.io/kubernetes/pkg/features"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/dynamicresources"
+	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/volumebinding"
 	"k8s.io/kubernetes/pkg/scheduler/util/assumecache"
 
 	schedulercache "github.com/apache/yunikorn-k8shim/pkg/cache/external"
@@ -796,6 +797,14 @@ func (ctx *Context) bindPodVolumes(pod *v1.Pod) error {
 					zap.String("nodeName", assumedPod.Spec.NodeName),
 					zap.Error(err))
 				return err
+			}
+			if volumes.StaticBindings == nil {
+				// convert nil to empty array
+				volumes.StaticBindings = make([]*volumebinding.BindingInfo, 0)
+			}
+			if volumes.DynamicProvisions == nil {
+				// convert nil to empty array
+				volumes.DynamicProvisions = make([]*volumebinding.DynamicProvision, 0)
 			}
 			err = ctx.apiProvider.GetAPIs().VolumeBinder.BindPodVolumes(context.Background(), assumedPod, volumes)
 			if err != nil {
--- a/pkg/cache/task.go
+++ b/pkg/cache/task.go
@@ -357,17 +357,14 @@ func (task *Task) postTaskPending() {
 // the allocation is rolled back to a pending ask so the core can re-schedule the task
 // on a different node. On success we move the task to the next state BOUND.
 func (task *Task) postTaskAllocated() {
-	go func() {
-		// Snapshot the fields needed for binding without holding the lock during the
-		// potentially slow bind calls. This avoids blocking other goroutines and lets the
-		// rollback/reschedule path (which acquires the lock) run when binding fails.
-		task.lock.RLock()
-		pod := task.pod
-		alias := task.alias
-		nodeName := task.nodeName
-		allocationKey := task.allocationKey
-		task.lock.RUnlock()
+	// Snapshot the fields needed for binding before launching the goroutine without re-acquiring the lock.
+	// We already hold task.lock (via task.handle()) during state transitions.
+	pod := task.pod
+	alias := task.alias
+	nodeName := task.nodeName
+	allocationKey := task.allocationKey
 
+	go func(pod *v1.Pod, alias, nodeName, allocationKey string) {
 		// post a message to indicate the pod gets its allocation
 		events.GetRecorder().Eventf(pod.DeepCopy(),
 			nil, v1.EventTypeNormal, "Scheduled", "Scheduled",
@@ -417,7 +414,7 @@ func (task *Task) postTaskAllocated() {
 		events.GetRecorder().Eventf(pod.DeepCopy(), nil,
 			v1.EventTypeNormal, "PodBindSuccessful", "PodBindSuccessful",
 			"Pod %s is successfully bound to node %s", alias, nodeName)
-	}()
+	}(pod, alias, nodeName, allocationKey)
 }
 
 // beforeTaskAllocated is called before handling the TaskAllocated event.
@@ -713,8 +710,8 @@ func (task *Task) rollbackAllocation(allocationKey, nodeID, eventReason, eventMs
 }
 
 // rescheduleOnBindFailure is called when volume or pod binding fails after all retries.
-// It moves the task from Allocated back to Scheduling and rolls the allocation back to a
-// pending ask so the core can re-schedule it on a different node.
+// Move the task back to Scheduling before rolling back the allocation, so a
+// subsequent TaskAllocated event from the core (which requires Scheduling) is accepted.
 // Must be called without holding the task lock.
 func (task *Task) rescheduleOnBindFailure(allocationKey, nodeID, eventReason, eventMsg string) {
 	// Move the task back to Scheduling before releasing to the core, so the re-delivered
--- a/pkg/cache/task.go
+++ b/pkg/cache/task.go
@@ -46,6 +46,7 @@ import (
 var bindPodBackoff = wait.Backoff{
 	Steps:    30,
 	Duration: time.Second,
+	Factor:   2,
 	Cap:      30 * time.Second,
 }
 
@@ -365,11 +366,6 @@ func (task *Task) postTaskAllocated() {
 	allocationKey := task.allocationKey
 
 	go func(pod *v1.Pod, alias, nodeName, allocationKey string) {
-		// post a message to indicate the pod gets its allocation
-		events.GetRecorder().Eventf(pod.DeepCopy(),
-			nil, v1.EventTypeNormal, "Scheduled", "Scheduled",
-			"Successfully assigned %s to node %s", alias, nodeName)
-
 		// before binding pod to node, first bind volumes to pod
 		log.Log(log.ShimCacheTask).Debug("bind pod volumes",
 			zap.String("podName", pod.Name),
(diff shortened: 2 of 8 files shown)


pkg/cache/context.go
```diff
@@ -798,14 +797,6 @@ func (ctx *Context) bindPodVolumes(pod *v1.Pod) error {
...
 				return err
 			}
-			if volumes.StaticBindings == nil {
-				// convert nil to empty array
-				volumes.StaticBindings = make([]*volumebinding.BindingInfo, 0)
-			}
-			if volumes.DynamicProvisions == nil {
-				// convert nil to empty array
-				volumes.DynamicProvisions = make([]*volumebinding.DynamicProvision, 0)
-			}
```
pbacsko, line 808: Why was this removed? I think it's a regression and it causes an e2e test to fail.

pkg/cache/task.go
```diff
@@ -343,45 +353,70 @@ func (task *Task) postTaskPending() {
...
-		defer task.lock.Unlock()
+		// Snapshot the fields needed for binding without holding the lock during the
+		// potentially slow bind calls. This avoids blocking other goroutines and lets the
+		// rollback/reschedule path (which acquires the lock) run when binding fails.
+		task.lock.RLock()
+		pod := task.pod
+		alias := task.alias
+		nodeName := task.nodeName
+		allocationKey := task.allocationKey
+		task.lock.RUnlock()
```
pbacsko: Copies can be obtained before `go func() { ... }` because lock is hold inside `postTaskAllocated()`. Not a big deal, but we can save a cycle of locking:

```
func (task *Task) postTaskAllocated() {
  // safe to copy, we have the mutex here
  pod := task.pod  (OR task.pod.DeepCopy() which might be even better)
  alias := task.alias
  nodeName := task.nodeName
  
  go  func(_, _, _) {  // new signature
  ...
  }(pod, alias, nodeName)
```

pkg/cache/task.go
```diff
@@ -669,6 +712,22 @@ func (task *Task) rollbackOnAssumePodFailure(allocationKey, nodeID string) {
 		zap.String("allocationKey", allocationKey))
 }
 
+// rescheduleOnBindFailure is called when volume or pod binding fails after all retries.
+// It moves the task from Allocated back to Scheduling and rolls the allocation back to a
+// pending ask so the core can re-schedule it on a different node.
+// Must be called without holding the task lock.
+func (task *Task) rescheduleOnBindFailure(allocationKey, nodeID, eventReason, eventMsg string) {
+	// Move the task back to Scheduling before releasing to the core, so the re-delivered
+	// allocation (valid only from the Scheduling state) is accepted by the state machine.
```
pbacsko, line 727: Misleading comment, a better one:

```
// Move the task back to Scheduling before rolling back the allocation, so a
// subsequent TaskAllocated event from the core (which requires Scheduling) is accepted.
```

pbacsko (changes requested):

pkg/cache/context.go
adityadtu5: Earlier I though it is redundant. But after checking these failing tests, it seems that volume binder plugin returns error in case if staticBindings or DynamicProvisions are nil. Adding them back.

pkg/cache/task.go
adityadtu5: thanks, done.

pkg/cache/task.go
adityadtu5: thanks, made these changes.

pkg/cache/task.go
```diff
@@ -343,46 +353,68 @@ func (task *Task) postTaskPending() {
...
+	nodeName := task.nodeName
+	allocationKey := task.allocationKey
 
+	go func(pod *v1.Pod, alias, nodeName, allocationKey string) {
 		// post a message to indicate the pod gets its allocation
-		events.GetRecorder().Eventf(task.pod.DeepCopy(),
+		events.GetRecorder().Eventf(pod.DeepCopy(),
 			nil, v1.EventTypeNormal, "Scheduled", "Scheduled",
-			"Successfully assigned %s to node %s", task.alias, task.nodeName)
+			"Successfully assigned %s to node %s", alias, nodeName)
```
wilfred-s: This is not really successful as yet, the attempt is being made but that might fail.
This should be moved to the point we have done the bind and we log it in line L407

pkg/cache/task.go
```diff
@@ -343,46 +353,68 @@ func (task *Task) postTaskPending() {
...
-		dispatcher.Dispatch(NewBindTaskEvent(task.applicationID, task.taskID))
-		events.GetRecorder().Eventf(task.pod.DeepCopy(), nil,
-			v1.EventTypeNormal, "PodBindSuccessful", "PodBindSuccessful",
-			"Pod %s is successfully bound to node %s", task.alias, task.nodeName)
+		log.Log(log.ShimCacheTask).Info("successfully bound pod", zap.String("podName", pod.Name))
 
+		task.lock.Lock()
 		task.schedulingState = TaskSchedAllocated
-	}()
+		task.lock.Unlock()
```
wilfred-s: This lock needs to be taken at the start of the go routine as we do not want anything to manipulate the task while running. It has to be a write lock also.
The removal of the lock outside the routine is correct but we need the lock inside the go routine as the first thing. The go routine will be blocked until the `handle()` returns as expected

pkg/cache/task.go
```diff
@@ -625,11 +657,20 @@ func (task *Task) failWithEvent(errorMessage, actionReason string) {
 // pending ask so it can be re-scheduled on a different node.
 // Must be called without holding the task lock.
 func (task *Task) rollbackOnAssumePodFailure(allocationKey, nodeID string) {
-	// Read fields needed for event posting and release request.
+	task.lock.RLock()
+	alias := task.alias
+	task.lock.RUnlock()
```
wilfred-s: Please check locking as this is called from an unlocked location. The callee `rollbackAllocation` needs a write lock to be held.

pkg/cache/task.go
```diff
@@ -625,11 +657,20 @@ func (task *Task) failWithEvent(errorMessage, actionReason string) {
...
+	task.rollbackAllocation(allocationKey, nodeID, "AssumePodFailed",
+		fmt.Sprintf("Node assignment failed for %s on node %s, it will be retried", alias, nodeID))
+}
+
+// rollbackAllocation resets the task allocation state and notifies the core to move
+// the allocation back to a pending ask so it can be re-scheduled on a different node.
+// Must be called without holding the task lock.
+func (task *Task) rollbackAllocation(allocationKey, nodeID, eventReason, eventMsg string) {
```
wilfred-s: This is now called from the locked go routine, no locking here.
Comment is incorrect, code should not lock.

pkg/cache/task.go
```diff
@@ -669,6 +709,22 @@ func (task *Task) rollbackOnAssumePodFailure(allocationKey, nodeID string) {
...
 }
 
+// rescheduleOnBindFailure is called when volume or pod binding fails after all retries.
+// Move the task back to Scheduling before rolling back the allocation, so a
+// subsequent TaskAllocated event from the core (which requires Scheduling) is accepted.
+// Must be called without holding the task lock.
+func (task *Task) rescheduleOnBindFailure(allocationKey, nodeID, eventReason, eventMsg string) {
+	// Move the task back to Scheduling before releasing to the core, so the re-delivered
+	// allocation (valid only from the Scheduling state) is accepted by the state machine.
+	if err := task.handle(NewRescheduleTaskEvent(task.applicationID, task.taskID)); err != nil {
```
wilfred-s: You must hold the task lock to do the (volume)binding. The calling go routine has the write lock. The handle will block as it needs a read lock.
The pattern used in the k8shim is not good for this. You either need to dispatch an event or simply change the state back via `task.sm.SetState(states.Scheduling)`. Preference is a direct state setting as it has less overhead than an event dispatch.

pkg/cache/task_state.go
```diff
@@ -47,10 +47,11 @@ const (
 	TaskFail
 	KillTask
 	TaskKilled
+	TaskRescheduling
```
wilfred-s: See above, direct state setting is preferred.

wilfred-s (changes requested):

pkg/cache/task_state.go
wilfred-s, line 1: should not be needed: changing the state directly instead of an event.

pkg/cache/task.go
```diff
@@ -39,6 +41,14 @@ import (
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
 )
 
+// bindPodBackoff controls the retry schedule for binding pod volumes and binding the
+// pod to a node. Declared as a package variable so tests can shorten it.
+var bindPodBackoff = wait.Backoff{
```
pbacsko: I realized if the intention is to cap the total amount of time spent in backoff, then this setting is not correct.

Cap suggests a total-time limit; it isn't one. Future readers may add Factor: 2 thinking total time stays ~30s, but they'd get much longer waits (1s + 2s + 4s + … up to 30s per sleep).

If the goal is a hard total deadline, wait.Backoff doesn't provide that directly. We might as well remove it: from here and from the existing backoff setting near AssumePod.

adityadtu5: You are right, currently it will retry till total `Steps` with the fixed delay of 1 second(as factor is set to 0, default value).
I think goal here is to retry with exponential backoff with 30s max delay `duration`. So I will add the `Factor` of 2. 

I will rerun e2e tests and push this change.

pkg/cache/task.go
adityadtu5: This is write lock and we only take it once all the bindings are success. This is to transition task to allocated. Let me know if I am missing something

pkg/cache/task.go
adityadtu5: Locking seems correct. It reads values while holding read lock and calls `rollbackAllocation` with those values. `rollBackAllocation` takes write lock as it updates some values on task.

pkg/cache/task.go
adityadtu5: Which go routine you are refering to? I checked and don't see any place where it is called while holding task lock.

pkg/cache/task.go
adityadtu5: agree, will move this down.

pkg/cache/task.go
adityadtu5: I think it is better to directly update the state as this all is happening on shim side only. Making this change.

pkg/cache/task.go
wilfred-s: While we are binding the task we do not want anything else to manipulate the task. If we do not lock the task could change while the volume bind or bind is running. Bind is a long running process.
It stops the go routine from starting until the handle call has returned.

We MUST have a write lock at the start of this go routine. That flows through the whole change as it fundamentally changes the locking for this call tree.

pkg/cache/task.go
wilfred-s: The go routine in `postTaskAllocated` must hold the write lock for the task while executing.
It was there before this change.

pkg/cache/task.go
wilfred-s: The go routine in `postTaskAllocated` must hold the write lock for the task while executing.
It was there before this change.

pkg/cache/task.go
adityadtu5: as discussed, releasing task lock after it is not necessary. This will prevent deadlock from other subsequent functions which take lock on `context` object.

pkg/cache/task.go
```diff
@@ -39,6 +41,15 @@ import (
...
 )
 
+// retryBackoff controls the retry schedule for binding pod volumes and binding the
+// pod to a node. Declared as a package variable so tests can shorten it.
+var retryBackoff = wait.Backoff{
+	Steps:    8, // total time is 1 + 2 + 4 + 8 + 16 + 30 + 30 + 30 = 121 seconds
+	Duration: time.Second,
+	Factor:   2,
+	Cap:      30 * time.Second,
+}
```
pbacsko, line 51: So total time = 120s is intentional?

adityadtu5: yes, this is as per internal discussion.
cc @wilfred-s

pkg/cache/task.go
```diff
@@ -343,46 +354,73 @@ func (task *Task) postTaskPending() {
...
-			log.Log(log.ShimCacheTask).Error("bind volumes to pod failed", zap.String("taskID", task.taskID), zap.Error(err))
-			task.failWithEvent(fmt.Sprintf("bind volumes to pod failed, name: %s, %s", task.alias, err.Error()), "PodVolumesBindFailure")
+			zap.String("podName", pod.Name),
+			zap.String("podUID", string(pod.UID)))
+		if err := retry.OnError(retryBackoff, func(err error) bool {
+			log.Log(log.ShimCacheTask).Error("bind volumes to pod failed, retrying",
+				zap.String("taskID", task.taskID), zap.Error(err))
+			return true
+		}, func() error {
+			return task.context.bindPodVolumes(pod)
```
wilfred-s, line 387: This has a timeout already which is 10 minutes. If we retry for that kind of failure the process would take 8*10m + 2m for the retry here, adding up to almost 1 1/2 hours. That is too much we need to retry API calls not the wait for the provisioning.

Worried about the backoff/retry on top of the long timeout.

wilfred-s (changes requested):

pkg/cache/task.go
wilfred-s: If we leave it at 30 steps as it was before we are adding another 22 * 30 seconds == 11 minutes. 8 retries with the backoff for API errors should be good.

pkg/cache/task.go
adityadtu5: removed retry on volumebindtimeout related steps

wilfred-s:
kicked off workflow

wilfred-s (approved):
LGTM

PR #1064, opened 2026-08-15
[YUNIKORN-3359] Run e2e tests with deadlock detection enabled

PoiBlackTea:
### Description
Enable deadlock detection when deploying YuniKorn for e2e tests by passing `--set deadlockDetection.enabled=true` during the Helm chart installation in `scripts/run-e2e-tests.sh`.

### Type of change

- [ ] Bug Fix
- [ ] Improvement
- [X] Feature
- [ ] Refactoring
- [ ] Documentation

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3359

- [ ] I have created a Jira issue for this pull request.

### AI Tooling
If an AI tool was used:
- [X] The PR includes the phrase "Generated by Antigravity IDE", where Antigravity IDE is the name of the AI tool used.
- [X] My use of AI contributions follows the ASF legal policy.

### How has this been tested?
- [ ] New unit tests were added to cover new or changed code paths.

### Questions:

### Screenshots or other details
This PR depends on the Helm chart changes in https://github.com/apache/yunikorn-release/pull/237 which introduce the deadlockDetection configuration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -590,7 +590,7 @@ test:
 	@echo "running unit tests"
 	@mkdir -p "$(OUTPUT)"
 	"$(GO)" clean -testcache
-	"$(GO)" test ./pkg/... -cover -race -tags deadlock -coverprofile="$(OUTPUT)/coverage.txt" -covermode=atomic
+	"$(GO)" test ./pkg/... -cover -race -coverprofile="$(OUTPUT)/coverage.txt" -covermode=atomic
 	"$(GO)" vet "$(REPO)"...
 
 # Run benchmarks
--- a/scripts/run-e2e-tests.sh
+++ b/scripts/run-e2e-tests.sh
@@ -168,7 +168,8 @@ function install_cluster() {
     --set admissionController.image.pullPolicy=IfNotPresent \
     --set web.image.repository=local/yunikorn \
     --set web.image.tag="${WEBTEST_IMAGE}" \
-    --set web.image.pullPolicy=IfNotPresent
+    --set web.image.pullPolicy=IfNotPresent \
+    --set deadlockDetection.enabled=true
   exit_on_error "failed to install yunikorn"
   "${KUBECTL}" wait --for=condition=available --timeout=300s deployment/yunikorn-scheduler -n yunikorn
   exit_on_error "failed to wait for yunikorn scheduler deployment being deployed"
--- a/scripts/run-e2e-tests.sh
+++ b/scripts/run-e2e-tests.sh
@@ -169,7 +169,9 @@ function install_cluster() {
     --set web.image.repository=local/yunikorn \
     --set web.image.tag="${WEBTEST_IMAGE}" \
     --set web.image.pullPolicy=IfNotPresent \
-    --set deadlockDetection.enabled=true
+    --set deadlockDetection.enabled=true \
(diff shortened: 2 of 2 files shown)


wilfred-s:
Should we not have the equivalent of the settings for the normal tests:
```
test: export DEADLOCK_DETECTION_ENABLED = true
test: export DEADLOCK_TIMEOUT_SECONDS = 10
test: export DEADLOCK_EXIT = true
```
Also the build tag `deadlock` is unknown to me and other reviewers I checked with. We traced it back to PR #4, a real old one, but it does not trigger anything as part of the build or running any tests. We should clean it up also.

PoiBlackTea:
> Should we not have the equivalent of the settings for the normal tests:
> 
> ```
> test: export DEADLOCK_DETECTION_ENABLED = true
> test: export DEADLOCK_TIMEOUT_SECONDS = 10
> test: export DEADLOCK_EXIT = true
> ```
Hi @wilfred-s, thanks for the review!
Just wanted to double check on the Makefile part: in e2e tests, YuniKorn runs inside the K8s cluster and gets the deadlock env vars via Helm in run-e2e-tests.sh. From what I can see, exporting DEADLOCK_* in make e2e_test would only affect the local Ginkgo runner (specifically the port-forward lock in k8s_utils.go).
Just wanted to see if that's what you had in mind, or if I might have missed something.

chenyulin0719:
>  From what I can see, exporting DEADLOCK_* in make e2e_test would only affect the local Ginkgo runner (specifically the port-forward lock in k8s_utils.go).

Correct. Add DEADLOCK_* to make e2e_test would be meaningless. What we really care about is whether the scheduler itself encounters deadlock. 

With another PR, the equivalent of deadlock detection settings are moved to the default value in helm chart. 
- https://github.com/apache/yunikorn-release/pull/237/changes#diff-9096e97495ff1a9082f744d0a08b7b33dd01025f0746d1b73710e4efbacf4980

Another way would be to add the equivalent flags to `run-e2e-tests.sh`
```
--set deadlockDetection.enabled=true \
--set deadlockDetection.timeoutSeconds=10 \
--set deadlockDetection.exit=true 
```

Both are good to me.

chenyulin0719 (approved):

wilfred-s:
> Another way would be to add the equivalent flags to `run-e2e-tests.sh`
> 
> ```
> --set deadlockDetection.enabled=true \
> --set deadlockDetection.timeoutSeconds=10 \
> --set deadlockDetection.exit=true 
> ```
> 
> Both are good to me.

@PoiBlackTea: I am sorry I did not make myself clear enough. The change in `run-e2e-tests.sh` is the correct location.
I would prefer that we add the settings as @chenyulin0719  shows above to the e2e test. In that way we use the same values for the e2e test as we do for unit tests. That gives us the same time out, shorter than the default, crash on deadlock after enabling it etc.

PoiBlackTea:
> > Another way would be to add the equivalent flags to `run-e2e-tests.sh`
I would prefer that we add the settings as @chenyulin0719 shows above to the e2e test. In that way we use the same values for the e2e test as we do for unit tests. That gives us the same time out, shorter than the default, crash on deadlock after enabling it etc.

NP. Thank you~

chenyulin0719 (approved):
Comment is addressed,  will merge it.

PR #1068, opened 2026-08-16
[YUNIKORN-3374] Revert assumed pods instead of orphaning them on node removal

tigerquoll:
### What is this PR for?

`removeNode` orphans every pod on the removed node's NodeInfo, including pods that were only *assumed* on it. An assumed pod's `Spec.NodeName` was stamped by the shim at assume time -- it is not an assignment the cluster made -- so when the node comes back, the orphan adoption path matches on a node name we invented ourselves: the never-bound pod is assigned to the node again, `Context.addNode` re-registers it, and task recovery reports an existing allocation to the core for a bind that never happened. A node flap (a kubelet re-registration, for example) is enough to create the phantom allocation of [YUNIKORN-3355](https://issues.apache.org/jira/browse/YUNIKORN-3355) without a single failed bind. The shim cache self-heals on the next informer update for the pod, but the core-side allocation persists.

The fix: in `removeNode`, a pod that is still assumed is reverted instead of orphaned -- assumed and assigned state dropped, node name cleared on a `DeepCopy`, pod kept in the cache as pending. It is then simply rescheduled on whatever node is available, which is also the correct outcome, as the node it was assumed on is gone. Pods the cluster actually placed on the node keep the existing orphan/adopt behaviour unchanged: their assignment is real.

This is the same discipline the [YUNIKORN-3355](https://issues.apache.org/jira/browse/YUNIKORN-3355) fix (#1066) established for `forgetPod`. Found by the invariant property fuzzer of [YUNIKORN-3373](https://issues.apache.org/jira/browse/YUNIKORN-3373) (#1067), whose "a pod that never bound is not assigned to a node" invariant fails on master via this path on every seed; with this fix (and #1066) the fuzzer passes.

Note: this PR duplicates two small test helpers (`newTestNode`/`newTestPod`) also added by #1066; whichever merges second will be rebased to drop the duplicate.

- New regression test: assume a pod, remove its node, re-add the node; asserts the pod is unassigned, not assumed, not orphaned, and not adopted. Fails on unmodified master (`assumed pod is returned as an orphan`), passes with the fix.
- Companion test pins the preserved behaviour: a pod the cluster reports as bound to the node IS orphaned on removal and adopted on re-add.
- Full package suites green with `-race` and deadlock detection; `make lint` and `make license-check` clean.
- The [YUNIKORN-3373](https://issues.apache.org/jira/browse/YUNIKORN-3373) fuzzer, run on top of this fix plus #1066, is green over 150+ seeds and 100k-step runs.

### DIFF
--- a/pkg/cache/external/scheduler_cache.go
+++ b/pkg/cache/external/scheduler_cache.go
@@ -207,8 +207,23 @@ func (cache *SchedulerCache) removeNode(node *v1.Node) (*v1.Node, []*v1.Pod) {
 	for _, fwkPod := range nodeInfo.Pods {
 		pod := fwkPod.GetPod()
 		key := string(pod.UID)
+		// a pod that is still assumed was never bound to the node: the node name it carries was set
+		// by the shim when the pod was assumed, it is not an assignment made by the cluster
+		revert := cache.isAssumedPod(key)
 		delete(cache.assignedPods, key)
 		delete(cache.assumedPods, key)
+		if revert {
+			// the assumed node is gone before the bind completed, revert the assignment instead of
+			// orphaning the pod: an orphan is adopted again when the node comes back and the pod
+			// would be recovered as an existing allocation for a bind that never happened. The pod
+			// is just pending again and is scheduled on whatever node is available.
+			// Copy before updating, the pod is shared with the pods map and can be owned by the
+			// informer cache.
+			unassigned := pod.DeepCopy()
+			unassigned.Spec.NodeName = ""
+			cache.podsMap[key] = unassigned
+			continue
+		}
 		cache.orphanedPods[key] = pod
 		orphans = append(orphans, pod)
 	}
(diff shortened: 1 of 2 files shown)


wilfred-s (approved):
LGTM
Not covered in the any of the bind failure scenarios as that does not handle the node removal.

wilfred-s:
This change by itself without proper handling of the bind failures would leave the system in a similar broken state. The cal to `Assume()` triggers an update of the task state which also triggers the async bind of the pod to the node. Without the proper handling of the volume and node binding errors that should happen in this case the task will be left in an allocated state in the k8shim and in a similar state on the core side.
This change cleans up the cache on the k8shim side it does not handle any of the changes on the core side that should be triggered. For that we need YUNIKORN-2884 as that reverts the core side and moves it back to scheduling via the call to `rescheduleOnBindFailure()`

PR #1074, opened 2026-08-22
[YUNIKORN-3367] Fix Kubernetes shim shutdown

andyhuangdev:
### Description

Fix KubernetesShim.Stop() to close its shared stop channel exactly once.

Previously, Stop() sent one value to an unbuffered channel shared by two scheduling loops. Only one loop stopped, while the other could continue scheduling after the shim was stopped. Also, if Run() failed before the scheduling loops started, no receiver was present and Stop() skipped dispatcher and placeholder-manager cleanup.

This change uses sync.Once to make shutdown idempotent, closes the channel to stop both scheduling loops, and always releases dispatcher and placeholder-manager resources during failed startup cleanup.

### Jira issues

- https://issues.apache.org/jira/browse/YUNIKORN-3367
- https://issues.apache.org/jira/browse/YUNIKORN-3368

### Testing

- Added unit tests for closing the scheduler stop channel after a failed Run().
- Added a unit test confirming Stop() closes the channel and is safe to call repeatedly.
- go test ./pkg/shim -run 'TestScheduler(RegistrationFailed|StopClosesStopChannel)$' -count=1
- go test ./pkg/shim -count=1

### DIFF
--- a/pkg/shim/scheduler.go
+++ b/pkg/shim/scheduler.go
@@ -20,6 +20,7 @@ package shim
 
 import (
 	ctx "context"
+	"sync"
 	"time"
 
 	"go.uber.org/zap"
@@ -49,6 +50,7 @@ type KubernetesShim struct {
 	phManager            *cache.PlaceholderManager
 	callback             api.ResourceManagerCallback
 	stopChan             chan struct{}
+	stopOnce             sync.Once
 	lock                 *locking.RWMutex
 	outstandingAppsFound bool
 }
@@ -224,14 +226,19 @@ func (ss *KubernetesShim) Run() error {
 }
 
 func (ss *KubernetesShim) Stop() {
-	log.Log(log.ShimScheduler).Info("stopping scheduler")
-	select {
-	case ss.stopChan <- struct{}{}:
-		// stop the dispatcher
-		dispatcher.Stop()
+	stopped := false
+	ss.stopOnce.Do(func() {
+		stopped = true
+		log.Log(log.ShimScheduler).Info("stopping scheduler")
+		close(ss.stopChan)
+		// stop the client library code that communicates with Kubernetes
+		ss.apiFactory.Stop()
 		// stop the placeholder manager
 		ss.phManager.Stop()
-	default:
+		// stop the dispatcher
+		dispatcher.Stop()
+	})
+	if !stopped {
 		log.Log(log.ShimScheduler).Info("scheduler is already stopped")
 	}
 }
--- a/pkg/shim/scheduler_mock_test.go
+++ b/pkg/shim/scheduler_mock_test.go
@@ -204,7 +204,6 @@ func (fc *MockScheduler) waitAndVerifySchedulerAllocations(
 func (fc *MockScheduler) stop() {
 	close(fc.stopChan)
 	fc.scheduler.Stop()
-	fc.apiProvider.Stop()
 	fc.started.Store(false)
 }
 
--- a/pkg/shim/scheduler_test.go
+++ b/pkg/shim/scheduler_test.go
@@ -175,7 +175,51 @@ func TestSchedulerRegistrationFailed(t *testing.T) {
 	ctx := cache.NewContext(mockedAPIProvider)
 	shim := newShimSchedulerInternal(ctx, mockedAPIProvider, callback)
 	assert.Error(t, shim.Run(), "some error")
+	assertStopChannelClosed(t, shim)
+
+	// Stop must remain safe when Run() already stopped the shim after its failure.
 	shim.Stop()
+}
+
+func TestSchedulerStopClosesStopChannel(t *testing.T) {
+	mockedAPIProvider := client.NewMockedAPIProvider(false)
+	shim := newShimSchedulerInternal(cache.NewContext(mockedAPIProvider), mockedAPIProvider, nil)
+
+	shim.Stop()
+	assertStopChannelClosed(t, shim)
+
+	// Stop must be idempotent because callers can invoke it after a failed Run().
+	shim.Stop()
+}
+
+func TestSchedulerStopStopsAPIFactory(t *testing.T) {
+	mockedAPIProvider := client.NewMockedAPIProvider(false)
+	apiProvider := &trackingAPIProvider{APIProvider: mockedAPIProvider}
+	shim := newShimSchedulerInternal(cache.NewContext(apiProvider), apiProvider, nil)
+
+	shim.Stop()
+	assert.Check(t, apiProvider.stopped.Load(), "API provider should be stopped with the scheduler")
+
+	shim.Stop()
+}
+
+func assertStopChannelClosed(t *testing.T, shim *KubernetesShim) {
+	t.Helper()
+	select {
+	case <-shim.stopChan:
+	default:
+		t.Fatal("scheduler stop channel should be closed")
+	}
+}
+
+type trackingAPIProvider struct {
+	client.APIProvider
+	stopped atomic.Bool
+}
+
+func (p *trackingAPIProvider) Stop() {
+	p.stopped.Store(true)
+	p.APIProvider.Stop()
 }
 
 func TestTaskFailures(t *testing.T) {


pkg/shim/scheduler.go
```diff
@@ -225,13 +227,16 @@ func (ss *KubernetesShim) Run() error {
 
 func (ss *KubernetesShim) Stop() {
 	log.Log(log.ShimScheduler).Info("stopping scheduler")
```
wilfred-s: NIT: move this inside the `stopOnce`. Logging now shows
```
stopping scheduler
scheduler is already stopped
```
Would be nice to only log the stopping line if we really do the work,

wilfred-s (changes requested):

manirajv06 (commented):
Can you also check the `placeholder` & `dispatcher` stop operation and ensure we are doing right job?

Also, Why `apifactory` is not being stopped? I can see `start() `being triggerred from `scheduler` but not stop from nowhere. @tigerquoll Any idea why is this not caught through leaks detector?

wilfred-s:
> Can you also check the `placeholder` & `dispatcher` stop operation and ensure we are doing right job?

The placeholder code has issue, we need a new jira to cleanup. There are tiny race conditions as we do a check and set of the atomic in two separate calls, should use `mgr.running.CompareAndSwap(false, true)` etc. It also does not set running to false on shutdown. The dispatcher has a similar check and set issue and maybe more.
@andyhuangdev could you file follow up jiras: one for the placeholder and one for the dispatcher to clean all that up?

> Also, Why `apifactory` is not being stopped? I can see `start() `being triggerred from `scheduler` but not stop from nowhere. @tigerquoll Any idea why is this not caught through leaks detector?

That is part of the informers and they are known to leak go routines. They are excluded from the checks as we can not fix that.
That said: we should call `ss.apiFactory.Stop()` as soon as part of the `KubernetesShim.Stop()` same way as we do the dispatcher and the phManager to get that done.

@andyhuangdev Please add that one line to this change

andyhuangdev:
@wilfred-s 
Thanks for the review.

I updated `KubernetesShim.Stop()` to:

- move the `stopping scheduler` log inside `stopOnce`, so it is only emitted when shutdown work is actually performed;
- stop the API factory as part of the scheduler shutdown;
- stop components in reverse startup order: API factory, placeholder manager, then dispatcher;
- remove the extra `apiProvider.Stop()` call from `MockScheduler.stop()` to avoid a double close in tests.

I also added a unit test verifying that stopping the scheduler stops its API provider, and ran the shim shutdown tests with the race detector.

I filed the requested follow-up issues for the existing lifecycle races:

- PlaceholderManager Start/Stop races: https://issues.apache.org/jira/browse/YUNIKORN-3404
- Dispatcher Start/Stop races: https://issues.apache.org/jira/browse/YUNIKORN-3405

The API factory informer goroutines are still covered by the existing leak exclusions, but the scheduler now correctly releases the API factory resources it owns during shutdown.

manirajv06 (approved):
looks good

tigerquoll (commented):
Change looks correct to me — closing the channel under a `sync.Once` is the right fix for both 3367 and 3368, and I agree with keeping the `stopped` local so we log exactly one of the two lines per call.

One request inline: a test that exercises the actual two-loop defect, since none of the new ones reach `doScheduling()`. Rest are nits.

While going through the shutdown path I noticed a few things that are pre-existing rather than introduced here, but that would be good to fold into the follow-up JIRAs already filed:

- **YUNIKORN-3404** (placeholder manager): beyond the check-then-set race, `Stop()` is an unbuffered send with no timeout. The only receiver alternates with `cleanOrphanPlaceholders()`, which holds the manager lock and issues deadline-less `Delete` calls, so a SIGTERM during an orphan sweep against a slow apiserver blocks shutdown for the whole sweep. Closing a channel under a `Once` (as this PR does for the shim) would fix it.
- **YUNIKORN-3405** (dispatcher): if `Stop()` hits its 5s timeout it leaves `running=true`, so the next `Start()` no-ops and once the stale loop eventually exits the package-global dispatcher has no event loop for the rest of the process. Only really bites the test binary, but worth covering in the same cleanup.
- `Stop()` still doesn't join the two `wait.Until` goroutines, so an in-flight `schedule()` iteration runs to completion after `Stop()` returns (`wait.Until` only checks the channel between calls). Same as before this PR and harmless in production since `main` exits immediately, but a `WaitGroup` would make the shutdown genuinely complete. Happy to file a separate JIRA for that if you'd rather keep this PR small.

pkg/shim/scheduler_test.go
```diff
@@ -175,9 +175,53 @@ func TestSchedulerRegistrationFailed(t *testing.T) {
 	ctx := cache.NewContext(mockedAPIProvider)
 	shim := newShimSchedulerInternal(ctx, mockedAPIProvider, callback)
 	assert.Error(t, shim.Run(), "some error")
+	assertStopChannelClosed(t, shim)
+
+	// Stop must remain safe when Run() already stopped the shim after its failure.
+	shim.Stop()
+}
+
+func TestSchedulerStopClosesStopChannel(t *testing.T) {
```
tigerquoll, line 184: None of the new tests actually reach the defect this PR fixes. The 3367 bug is "one send stops only one of the two `wait.Until` loops", but:

- `TestSchedulerRegistrationFailed` fails inside `registerShimLayer()` before `doScheduling()` is ever called
- `TestSchedulerStopClosesStopChannel` and `TestSchedulerStopStopsAPIFactory` never call `Run()`

So the loops don't exist in any of them, and `assertStopChannelClosed` is a non-blocking receive that a regression back to a buffered single send would also pass. The `MockScheduler` tests do run `Run()`+`Stop()` but assert nothing about the goroutines exiting.

Could we add one test that runs `Run()` successfully against the mock, calls `Stop()`, and asserts both loop goroutines terminate? `goleak.VerifyNone` scoped to that test, or a small counter/WaitGroup on the shim that the test can wait on, would both work. That's the regression guard for the actual fix.

pkg/shim/scheduler_test.go
```diff
@@ -175,9 +175,53 @@ func TestSchedulerRegistrationFailed(t *testing.T) {
...
+	shim := newShimSchedulerInternal(cache.NewContext(mockedAPIProvider), mockedAPIProvider, nil)
+
+	shim.Stop()
+	assertStopChannelClosed(t, shim)
+
+	// Stop must be idempotent because callers can invoke it after a failed Run().
+	shim.Stop()
+}
+
+func TestSchedulerStopStopsAPIFactory(t *testing.T) {
```
tigerquoll, line 195: Minor: this and `TestSchedulerStopClosesStopChannel` are the same test with one assertion swapped — `Stop()` closes the channel and stops the API factory in the same `Once` body, so a single `TestSchedulerStop` can `assert.Check` both after one `Stop()` call.

Also, `trackingAPIProvider` re-implements what `MockedAPIProvider` already tracks in its `running` field (cleared in `Stop()`). An `IsRunning()` getter on the mock would let this test drop the wrapper — note the test would need to call `RunEventHandler()` first, otherwise `running` is never set and the assertion is vacuous.

pkg/shim/scheduler_mock_test.go
```diff
@@ -204,7 +204,6 @@ func (fc *MockScheduler) waitAndVerifySchedulerAllocations(
 func (fc *MockScheduler) stop() {
 	close(fc.stopChan)
 	fc.scheduler.Stop()
```
tigerquoll, line 206: Removing this call fixes the double close, but it does so by coupling the harness to the shim's internal call order. `MockedAPIProvider.Stop()` already takes the lock and has a `running` flag, so guarding it with `if !m.running { return }` makes it safe for any caller and would let this line stay. Either is fine for this PR; the guard is more robust for future tests that build their own provider.

andyhuangdev:
@tigerquoll 
Thanks for the detailed review.

I updated the follow-up issues with the additional shutdown findings:
- PlaceholderManager lifecycle and shutdown blocking behavior:
https://issues.apache.org/jira/browse/YUNIKORN-3404
- Dispatcher lifecycle, timeout recovery, and running-state behavior:
https://issues.apache.org/jira/browse/YUNIKORN-3405

I also filed a separate issue for the missing regression coverage of the two `doScheduling()` loops:
- Add regression test coverage for KubernetesShim scheduling loop shutdown:
https://issues.apache.org/jira/browse/YUNIKORN-3431

PR #1079, opened 2026-08-25
[YUNIKORN-3408] Adopt the vet-lock static lock analyser (annotations + build wiring)

tigerquoll:
### Description

Machine-checks the lock conventions that are today enforced only by review (which fields a lock guards, which functions require a lock held on entry, which must not be called with a lock held, `String()` taking its own lock). Adds the [vet-lock](https://github.com/tigerquoll/vet-lock) static analyser pinned in an isolated tools module, wires it into `make vetlock` / `test_all` / pre-commit, and annotates the shim. Companion of the core PR apache/yunikorn-core#1137.

Behaviour-neutral apart from one test-support fix (`pkg/common/test/schedulerapi_mock.go`: `RegisterFunction` now takes the mock's lock like its three sibling setters). The annotations are comments; the only other runtime code change is the `pkg/locking` named-field lock + forwarding methods (inlined) that let annotations resolve to the wrapper field.

- **Analysers gated**: `checklocks`, `lockstringer`, `lockblocking` (as in core).
- **Tool pinned**: `github.com/tigerquoll/vet-lock v0.8.0` in `scripts/vetlock`; `-checklocks.inferred=false`.
- **Annotated**: `pkg/locking`, `pkg/cache` (+ `external`), `pkg/admission`, `pkg/client`, `pkg/dispatcher`, `pkg/shim`, `pkg/plugin`, `pkg/common/test`, `pkg/cmd/admissioncontroller`. FSM callback dispatch sites carry line-level ignores with a comment: the lock is held by `Application`/`Task.handle` but the acquisition is invisible across the fsm library boundary.
- **Canary** fixture fails the build if an analyser silently stops covering what it should.
- `+lockclass:<name>` declarations kept for `lockblocking`; lock-ordering relations are a separate PR.

**The suppression list is a burn-down list.** Pre-existing locking issues surfaced while annotating are marked in-code with a one-line `// YUNIKORN-NNNN:` marker plus the ignore directive; each is its own JIRA (linked from YUNIKORN-3408) with the analysis and the fix shape, and each fix PR deletes its marker. None is fixed here — same split as YUNIKORN-3357.

Not in this PR: lock-order analyser / runtime lock-class checker; `lockgap`.

Worth noting: rebasing this branch onto current master made the gate report two new sites in code merged since the annotations were written — YUNIKORN-2884 runs both bind attempts through `retry.OnError` (up to ~121 s of backoff) while holding the task write lock. That is the class of change the gate is meant to stop at review time; it is marked and tracked as YUNIKORN-3426 (its own JIRA), not fixed here.

### Type of change

- [x] Improvement

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3408

### AI Tooling

### How has this been tested?
- [ ] New unit tests were added to cover new or changed code paths. (The canary fixture is the test of the gate.)
- [ ] `make e2e_test` was run, and no failures reported. (No runtime change to exercise end-to-end.)

### Questions:

### Screenshots or other details
`make vetlock` produces no output on this branch. The canary stage of the same target proves each analyser still reports: it runs the fixture with one known violation per check and fails the build if any expected report is missing.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -23,10 +23,20 @@ jobs:
         with:
           go-version-file: .go_version
           check-latest: true
+      - name: Cache and Restore required tools
+        uses: actions/cache@v5
+        with:
+          path: |
+            tools
+          key: ${{ runner.os }}-tools-${{ hashFiles('Makefile', 'scripts/vetlock/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-tools-
       - name: Check license
         run: make license-check
       - name: Go lint
         run: make lint
+      - name: Check lock annotations
+        run: make vetlock
       - name: Run Version Check
         run: make pseudo
       - name: Run ShellCheck
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@
 .PHONY: run build build_web_test_server_prod build_web_test_server_dev
 # test targets
 .PHONY: test_all test bench fsm_graph clean distclean arch
-.PHONY: lint check_scripts license-check go-license-check pseudo
+.PHONY: lint vetlock vetlock-toolchain check_scripts license-check go-license-check pseudo
 # e2e test targets
 .PHONY: print_kubectl_version print_kind_version print_helm_version
 .PHONY: e2e_test kind-e2e start-cluster stop-cluster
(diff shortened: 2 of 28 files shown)


tigerquoll:
Updated (force-push `968e95b5` → `ad84a2f3`), after a full pass over every annotation against the code (writers enumerated per guarded field, callers per precondition, each opt-out checked, each ignore's stated reason re-verified). No guard turned out to be wrong and no marker misdescribes its site; the changes are small:

- `PlaceholderManager.cleanupTime` is always used under the manager lock but carried no guard; now `+checklocks:RWMutex` (verified load-bearing: stripping the lock from its getter makes the analyser report it).
- `SchedulerCache.LockForReads` carried a redundant `+checklocksexclude` next to `+checklocksacquireread`; removed, the acquire annotation already implies it.
- The `pkg/dispatcher` comment claimed the analysis cannot see the lock through `getDispatcher()`; it can (confirmed with a probe), so that sentence is gone. The directives are unchanged.
- Makefile: the canary check now also asserts the report contains exactly as many diagnostics as expected messages, so a new, unexpected report cannot hide behind the presence check (verified: a ninth violation fails the target). The `VETLOCK_PACKAGES` comment now says `pkg/...`; the e2e helpers under `test/` are outside the analysed set.

All gates green locally (`make vetlock` incl. canary, lint, license-check, check_scripts, full `make test`). The pre-commit workflow run for this PR is waiting on approval; the same commits are green on the fork's CI.

PR #1081, opened 2026-08-31
[YUNIKORN-3404] Make PlaceholderManager lifecycle concurrency-safe

andyhuangdev:
### Description

Make the PlaceholderManager lifecycle atomic, idempotent, and shutdown-safe.

- Ensure concurrent `Start()` calls launch only one cleanup goroutine.
- Close the stop channel once and allow concurrent `Stop()` calls to wait for the same completion signal.
- Guarantee the cleanup goroutine has exited when `Stop()` returns.
- Propagate cancellation to Kubernetes pod deletion requests so shutdown does not wait indefinitely on a blocked API call.
- Avoid holding the orphan-pod lock during Kubernetes API calls.
- Add concurrent lifecycle and blocked-cleanup regression tests.

### Type of change

- [x] Bug Fix

### Jira issue

Jira ID: https://issues.apache.org/jira/browse/YUNIKORN-3404

### How has this been tested?

- [x] `make test` was run, and no failures were reported.
- [x] `go test -race ./pkg/cache -run '^TestPlaceholderManager' -count=1` passed.

### DIFF
--- a/pkg/cache/placeholder_manager.go
+++ b/pkg/cache/placeholder_manager.go
@@ -19,7 +19,9 @@
 package cache
 
 import (
+	"context"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -40,7 +42,12 @@ type PlaceholderManager struct {
 	// and keep retrying deleting them in order to avoid wasting resources.
 	orphanPods  map[string]*v1.Pod
 	stopChan    chan struct{}
+	doneChan    chan struct{}
+	stopOnce    sync.Once
+	started     atomic.Bool
 	running     atomic.Bool
+	cancel      context.CancelFunc
+	cleanupCtx  context.Context
 	cleanupTime time.Duration
 	// a simple mutex will do we do not have separate read and write paths
 	locking.RWMutex
@@ -54,10 +61,14 @@ var (
 func NewPlaceholderManager(clients *client.Clients) *PlaceholderManager {
 	mu.Lock()
 	defer mu.Unlock()
+	cleanupCtx, cancel := context.WithCancel(context.Background())
 	placeholderMgr = &PlaceholderManager{
 		clients:     clients,
 		orphanPods:  make(map[string]*v1.Pod),
 		stopChan:    make(chan struct{}),
+		doneChan:    make(chan struct{}),
+		cleanupCtx:  cleanupCtx,
+		cancel:      cancel,
 		cleanupTime: 5 * time.Second,
 	}
 	return placeholderMgr
@@ -103,18 +114,18 @@ func (mgr *PlaceholderManager) createAppPlaceholders(app *Application) error {
 
 // clean up all the placeholders for an application
 func (mgr *PlaceholderManager) cleanUp(app *Application) {
-	mgr.Lock()
-	defer mgr.Unlock()
 	log.Log(log.ShimCachePlaceholder).Info("start to clean up app placeholders",
 		zap.String("appID", app.GetApplicationID()))
 	for _, task := range app.GetPlaceHolderTasks() {
 		// remove pod
-		err := mgr.clients.KubeClient.Delete(task.GetTaskPod())
+		err := mgr.clients.KubeClient.Delete(context.Background(), task.GetTaskPod())
 		if err != nil {
 			log.Log(log.ShimCachePlaceholder).Warn("failed to clean up placeholder pod",
 				zap.Error(err))
 			if !strings.Contains(err.Error(), "not found") {
+				mgr.Lock()
 				mgr.orphanPods[task.GetTaskID()] = task.GetTaskPod()
+				mgr.Unlock()
 			}
 		}
 	}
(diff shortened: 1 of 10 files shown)


pkg/cache/placeholder_manager.go
```diff
@@ -103,70 +114,97 @@ func (mgr *PlaceholderManager) createAppPlaceholders(app *Application) error {
...
+				delete(mgr.orphanPods, orphan.taskID)
+			}
+			mgr.Unlock()
 		}
 	}
 }
 
 func (mgr *PlaceholderManager) Start() {
-	if mgr.isRunning() {
+	if !mgr.started.CompareAndSwap(false, true) {
```
manirajv06, line 199: Do we still need `running` flag?

pkg/cache/placeholder_manager.go
```diff
@@ -40,7 +42,12 @@ type PlaceholderManager struct {
 	// and keep retrying deleting them in order to avoid wasting resources.
 	orphanPods  map[string]*v1.Pod
 	stopChan    chan struct{}
+	doneChan    chan struct{}
```
manirajv06, line 46: Do we really need this?

pkg/cache/placeholder_manager.go
```diff
@@ -103,70 +114,97 @@ func (mgr *PlaceholderManager) createAppPlaceholders(app *Application) error {
...
 // clean up all the placeholders for an application
 func (mgr *PlaceholderManager) cleanUp(app *Application) {
-	mgr.Lock()
-	defer mgr.Unlock()
 	log.Log(log.ShimCachePlaceholder).Info("start to clean up app placeholders",
 		zap.String("appID", app.GetApplicationID()))
 	for _, task := range app.GetPlaceHolderTasks() {
 		// remove pod
-		err := mgr.clients.KubeClient.Delete(task.GetTaskPod())
+		err := mgr.clients.KubeClient.Delete(context.Background(), task.GetTaskPod())
```
manirajv06: Is there any possibility of executing this operation around the same time when placeholder manager is shutting down as this cleanup is being asked to run asynchronously from app?

pkg/cache/placeholder_manager.go
```diff
@@ -103,70 +114,97 @@ func (mgr *PlaceholderManager) createAppPlaceholders(app *Application) error {
...
+				return
+			}
 			log.Log(log.ShimCachePlaceholder).Warn("failed to clean up orphan pod", zap.Error(err))
 		} else {
-			delete(mgr.orphanPods, taskID)
+			mgr.Lock()
+			if mgr.orphanPods[orphan.taskID] == orphan.pod {
+				delete(mgr.orphanPods, orphan.taskID)
+			}
+			mgr.Unlock()
```
manirajv06, line 193: I am not really sure about the trade off we are making with lock changes here. From code maintainability perspective, I am not in favour of locking model being used here unless there is compelling reason to do. I prefer the earlier way of single write lock model itself without the need to build orphan pods again. Only change required here is checking for the context errors to ensure we are correct job on handling shut down process.

manirajv06 (changes requested):

pkg/cache/placeholder_manager.go
andyhuangdev: Makes sense. I'll refactor this part.

pkg/cache/placeholder_manager.go
andyhuangdev: The purpose of `doneChan` is to make `Stop()` synchronous. When `Stop()` returns, we can guarantee that the cleanup goroutine has already exited.

pkg/cache/placeholder_manager.go
andyhuangdev: You're right. We need to preserve the operation-level serialization here. I'll fix this.

pkg/cache/placeholder_manager.go
```diff
@@ -103,80 +120,121 @@ func (mgr *PlaceholderManager) createAppPlaceholders(app *Application) error {
...
-	log.Log(log.ShimCachePlaceholder).Info("stopping the PlaceholderManager")
-	mgr.stopChan <- struct{}{}
-}
-
-func (mgr *PlaceholderManager) isRunning() bool {
-	return mgr.running.Load()
+	mgr.stopOnce.Do(func() {
+		log.Log(log.ShimCachePlaceholder).Info("stopping the PlaceholderManager")
+		mgr.cleanupLock.Lock()
+		mgr.stopping = true
```
manirajv06: Can we make `stopping` atomic bool to avoid locking here? That way, we `cleanupLock` lock itself could be avoided easily as using `cleanupworkers` does not require to operate upon.

pkg/cache/placeholder_manager.go
```diff
@@ -103,80 +120,121 @@ func (mgr *PlaceholderManager) createAppPlaceholders(app *Application) error {
 
 // clean up all the placeholders for an application
 func (mgr *PlaceholderManager) cleanUp(app *Application) {
-	mgr.Lock()
-	defer mgr.Unlock()
+	mgr.operationLock.Lock()
+	defer mgr.operationLock.Unlock()
```
manirajv06, line 138: I believe reason for this lock is to prevent creation ( createAppPlaceholders() )and deletion (here) at the same time. Isn't it? But we don't use this lock in cleanOrphanPlaceholders() at all.

BTW, I am thinking of something like this:

1. Stop signal arrived.
2. Broadcast above signal to all readers. All cleanup methods and create etc are the readers.
3. All readers should consume the signal and act accordingly.

Can we get this into actionable code?

pkg/cache/placeholder_manager.go
manirajv06: Can we take care of this?

manirajv06 (changes requested):

pkg/cache/placeholder_manager.go
andyhuangdev: I don't think replacing `stopping` with an atomic bool is sufficient to safely remove `cleanupLock`.

The lock currently protects the admission check and `cleanupWorkers.Add(1)` as one operation. Without that serialization, the following interleaving is possible:

1. `cleanUpAsync()` reads `stopping == false`.
2. `Stop()` stores `stopping = true`.
3. `Stop()` calls `cleanupWorkers.Wait()` while the counter is still zero, so it returns.
4. `cleanUpAsync()` calls `cleanupWorkers.Add(1)` and starts a new cleanup goroutine.

This would allow cleanup work to start after `Stop()` has already returned. It also violates the `WaitGroup` requirement that a positive `Add()` from a zero counter must happen before `Wait()`.

I agree that the current lifecycle synchronization can be simplified, but we still need an admission barrier around starting operations and stopping the manager. I have outlined a lifecycle-gate approach in the other thread that could remove `cleanupLock`, `stopping`, and `cleanupWorkers` together.

pkg/cache/placeholder_manager.go
andyhuangdev: Yes, I think a common lifecycle signal and gate for all placeholder operations would make the shutdown behavior more consistent.

My understanding of the proposed model is:

1. The manager owns a cancellable lifecycle context.
2. Placeholder creation, application cleanup, and orphan cleanup all enter through the same lifecycle gate as readers.
3. After entering, each operation checks the lifecycle context before doing work and passes the context to Kubernetes API calls where applicable.
4. `Stop()` cancels the context first, broadcasting the shutdown signal to all active operations.
5. `Stop()` then acquires the lifecycle gate as a writer. This prevents new operations from being admitted and waits for all existing readers to exit.
6. `doneChan` remains responsible for confirming that the periodic orphan-cleanup loop has exited.
7. `operationLock` remains separate and is used only to serialize `createAppPlaceholders()` with application `cleanUp()`, preserving the existing create-versus-delete ordering.

With this model, the lifecycle gate would replace `cleanupLock`, `stopping`, and `cleanupWorkers`. New operations would observe the canceled context and return without doing work, while `Stop()` would not return until operations that were already admitted have exited.

For `cleanOrphanPlaceholders()`, I will also restore the simpler single write-lock model while retaining the context checks.

Is this aligned with the reader/writer and shutdown model you had in mind?

manirajv06: All good except 6 & 7. Great.

6. I am not sure about this. We are not really writing anything into `done` channel anywhere. How do we ensure "The purpose of doneChan is to make Stop() synchronous. When Stop() returns, we can guarantee that the cleanup goroutine has already exited." ? 
7. May be once you make changes based on the above discussion, we can take a re look at this.

PR #1090, opened 2026-09-09
[YUNIKORN-3447] Drop managed fields in the informers

wilfred-s:
### Description
The objects loaded by the informers contain a field called ManagedFields This can be a large part of the object to store. The field is not used in the scheduler and removed in the default scheduler also to limit memory overhead.

Applying the same change to the informers in the K8shim.

### Type of change
- [X] Improvement

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3447

### AI Tooling
If an AI tool was used:

### How has this been tested?
- [ ] New unit tests were added to cover new or changed code paths.
- [X] `make e2e_test` was run, and no failures reported.

### Questions:

### DIFF
--- a/pkg/admission/informers.go
+++ b/pkg/admission/informers.go
@@ -22,9 +22,11 @@ import (
 	"time"
 
 	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/client-go/informers"
 	informersv1 "k8s.io/client-go/informers/core/v1"
 	schedulinginformersv1 "k8s.io/client-go/informers/scheduling/v1"
+	"k8s.io/client-go/tools/cache"
 
 	"github.com/apache/yunikorn-k8shim/pkg/client"
 	"github.com/apache/yunikorn-k8shim/pkg/log"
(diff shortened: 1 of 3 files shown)


manirajv06 (commented):
No possibility of re-using stripManagedFields() ?

tigerquoll:
Also what about annotations like last-applied-configuration that can be a bit hefty

tigerquoll:
> No possibility of re-using stripManagedFields() ?

are you referring to https://pkg.go.dev/k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/client stripManagedFields ?
The code is extremely small and self contained, I would argue pulling a copy of the func locally rather then including the entire module is defensible for this type of function.

wilfred-s:
> Also what about annotations like last-applied-configuration that can be a bit hefty

No we cannot do that it could break updates of pods when we start changing those fields. The scheduler does not own those and should not touch them unless they are scheduling related

wilfred-s:
> > No possibility of re-using stripManagedFields() ?
> 
> are you referring to [pkg.go.dev/k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/client](https://pkg.go.dev/k8s.io/autoscaler/vertical-pod-autoscaler/pkg/utils/client) stripManagedFields ? The code is extremely small and self contained, I would argue pulling a copy of the func locally rather then including the entire module is defensible for this type of function.

No this was about reusing the same function declared once in our code in multiple places. I specifically added the different versions in the different factories as we might want to strip further in the admission controller or tests later on if needed.

manirajv06 (approved):
looks good

PR #1091, opened 2026-09-10
[YUNIKORN-3428] Fix fatal scheduler exit on concurrent taskMap access

rjgoyln:
## Summary

`KubernetesShim.schedule` calls `AreAllTasksTerminated` once per tick for every Failed application without holding the application lock, and that method ranges over `taskMap`. A pod arriving for the same application can then trigger a concurrent map iteration and write, causing the runtime to abort. Although the race window is narrow, it can terminate the shim process.

`String` reads `len(app.taskMap)`, and its only caller logs it as a `zap.Stringer` from inside an FSM callback while holding the application write lock. `String` therefore cannot acquire the application lock itself and must not access `taskMap`.

## Change

* `AreAllTasksTerminated` takes the read lock and delegates to an unlocked `areAllTasksTerminated`, mirroring the existing `GetPlaceHolderTasks` / `getPlaceHolderTasks` pattern.
* `flushReleaseableTasks`, which already holds the write lock, uses the unlocked variant.
* `String` drops the task count; its only caller logs the count separately while holding the application lock.
* `getNonTerminatedTaskAlias` is removed; the alias list it built had no reader.

The remaining fields read by `String` are safe: `applicationID`, `queue`, and `partition` are initialized once in `NewApplication`, while the state machine protects its current state with its own lock.

`postAppAccepted` and `skipReservationStage` also read `taskMap` without the application lock on the same scheduling loop; that is tracked separately by YUNIKORN-3423 and is untouched here. #1079 already adds `+checklocksignore` to the two sites fixed here; the later PR to land should remove those annotations.

## Tests

`TestTaskMapReadersAreRaceFree` races both readers against `addTask`; the race detector reproduces both races on `master` and reports no races on this branch.

### Type of change

- [x] Bug Fix

### Jira issue

Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3428

### AI Tooling

### How has this been tested?

### Questions:

### DIFF
--- a/pkg/cache/application.go
+++ b/pkg/cache/application.go
@@ -68,10 +68,12 @@ type Application struct {
 
 const transitionErr = "no transition"
 
+// String is called from paths that already hold the application lock, so it must stay
+// lock-free. It reads only fields fixed at construction plus the state machine, which
+// guards its own current state.
 func (app *Application) String() string {
-	return fmt.Sprintf("applicationID: %s, queue: %s, partition: %s,"+
-		" totalNumOfTasks: %d, currentState: %s",
-		app.applicationID, app.queue, app.partition, len(app.taskMap), app.GetApplicationState())
+	return fmt.Sprintf("applicationID: %s, queue: %s, partition: %s, currentState: %s",
+		app.applicationID, app.queue, app.partition, app.GetApplicationState())
 }
 
 func NewApplication(appID, queueName, user string, groups []string, tags map[string]string, scheduler api.SchedulerAPI) *Application {
@@ -334,6 +336,13 @@ func (app *Application) getNonTerminatedTaskAlias() []string {
 }
 
 func (app *Application) AreAllTasksTerminated() bool {
+	app.lock.RLock()
+	defer app.lock.RUnlock()
+	return app.areAllTasksTerminated()
+}
+
+// areAllTasksTerminated must be called while the application lock is held.
+func (app *Application) areAllTasksTerminated() bool {
 	return len(app.getNonTerminatedTaskAlias()) == 0
 }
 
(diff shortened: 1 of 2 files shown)


HuangTing-Yao (approved):
LGTM. 
The change aligns with the project's [Concurrency and Object Conventions](https://yunikorn.apache.org/docs/next/developer_guide/concurrency/) for String() on locked objects.

rjgoyln:
> LGTM. The change aligns with the project's [Concurrency and Object Conventions](https://yunikorn.apache.org/docs/next/developer_guide/concurrency/) for String() on locked objects.

Thanks for the review!

chenyulin0719 (approved):
+1 LGTM, 

This PR is worth fixing — it prevents a shim crash in the following scenario:
- Informer add a task to application (hold app write lock)
- At the same time, scheduler detected application has failed and calls `app.AreAllTasksTerminated()` to iterator the app.taskMap (trigger a fatal crash)

I did think about the downside of removing `totalNumOfTasks` from `app.String()` (which will be used in `handleSubmitApplicationEvent`). 
I believe it's acceptable. the count was never a reliable signal to begin with. We only know it's at least 1 task, and the exact number depends on how fast the informer processes events, so it wasn't giving us much actionable information for debugging anyway.

manirajv06 (approved):
looks good.

Can we move the logic from `getNonTerminatedTaskAlias` to `AreAllTasksTerminated` itself? so that test code can be cleaned up altogether.

manirajv06 (changes requested):
Clean up

manirajv06 (approved):
looks good, pending test

rjgoyln:
Thank for the review.
I have removed `getNonTerminatedTaskAlias` and inlined the logic into `areAllTasksTerminated`. Cleaned up the related tests as suggested.

PR #1093, opened 2026-09-16
[YUNIKORN-3382] Re-drive failed Pod deletion with UID fencing

hikaru-212:
### Description

This PR fixes YUNIKORN-3382 by retaining responsibility for scheduler-originated
Pod deletion when the initial Kubernetes DELETE fails.

The motivating failure is preemption:

1. scheduler core selects/releases a victim allocation;
2. k8shim receives the release and calls `Task.DeleteTaskPod()`;
3. the Kubernetes Pod DELETE fails;
4. previously, the error was logged and returned, but no component retained
   responsibility for completing that deletion.

The victim Pod could therefore remain alive while core-side release/preemption
state was already established and normal allocation release still depended on
later Pod lifecycle events.

This PR changes that behavior in three parts.

#### 1. Retain Task-local deletion responsibility

`Task.DeleteTaskPod()` now claims a single deletion lifecycle for the Task.

The claim is represented by `podDeleteClaimed`, which prevents duplicate release
events from starting independent deletion lifecycles for the same Task.

For retryable failures, one background retry worker started by
`startTaskPodDeleteRetries()` re-drives deletion of the captured original Pod
identity using capped exponential backoff:

- initial delay: 100 ms
- maximum delay: 30 s
- no retry-count limit

Duplicate release events do not create additional retry workers.

Non-retryable request errors (`Forbidden`, `Invalid`, `BadRequest`,
`MethodNotSupported`) stop automatic retries and release the claim so a later
explicit release event can acquire it again.

A terminal Task stops the background retry loop.

The initial synchronous DELETE still returns its original error even when
background retry responsibility has been retained.

#### 2. Fence destructive deletes to the original Pod identity

`SchedulerKubeClient.Delete(*Pod)` now sends a Kubernetes UID precondition.

Previously, deletion was effectively name-based:

    delete whichever Pod currently occupies namespace/name

After this change, deletion is identity-based:

    delete only the supplied observed Pod UID

This prevents a stale retry for Pod A from deleting a later Pod B that reuses
the same namespace/name.

A Pod without a UID is rejected locally as `BadRequest`; no unfenced Kubernetes
DELETE is sent.

#### 3. Reconcile DELETE conflicts against current identity

A generic Kubernetes `Conflict` is not treated as sufficient evidence that the
original Pod has disappeared or been replaced.

`reconcileTaskPodDeletion()` reads the current Pod at the same namespace/name
when DELETE returns Conflict:

- `NotFound`
  - no further DELETE for the original UID is required;
- current Pod has a different non-empty UID
  - the name now refers to another Pod, so the stale DELETE is stopped;
- current Pod has the original UID
  - the original Pod still exists, so retry responsibility is retained;
- GET error / missing usable identity
  - current identity is uncertain, so responsibility is retained and retry
    continues.

The reconciliation GET is evidence for deciding whether another DELETE is
required.

It does **not** synthesize Task completion or scheduler-core release
confirmation.

Task completion and allocation release confirmation remain on the existing
informer-driven path.

---

#### Scope / behavioral impact

YUNIKORN-3382 was discovered through the preemption path, but
`Task.DeleteTaskPod()` is shared by other scheduler-originated release paths.

The same re-drive semantics therefore also apply to release types such as:

- `PREEMPTED_BY_SCHEDULER`
- `TIMEOUT`
- `PLACEHOLDER_REPLACED`

`STOPPED_BY_RM` remains filtered before this path.

There is no Go method-signature change, but there **is** an intentional semantic
change to `KubeClient.Delete(*Pod)`:

    before: namespace/name identified the delete target

    after:  namespace/name + observed UID identify the delete target

Because UID fencing is implemented in `SchedulerKubeClient.Delete`, it also
applies to PlaceholderManager cleanup.

This prevents stale placeholder cleanup from deleting a same-name replacement
Pod, but it also means callers of `KubeClient.Delete` must now provide a valid
observed Pod UID.

This is therefore broader than a preemption-only retry change and should be
reviewed as a stronger deletion-identity contract for k8shim.

---

#### What this PR does NOT add

This PR intentionally keeps the fix Task-local and does not introduce a general
release/reconciliation subsystem.

It does **not** add:

- a shared retry work queue;
- a bounded global worker pool;
- retry jitter;
- a global retry rate limiter;
- durable retry state across process restart;
- Context-level ownership transfer;
- explicit cancellation/joining of retry workers on shim/API-factory shutdown;
- a redesign of PlaceholderManager orphan reconciliation.

The current retry model allows at most one automatic retry lifecycle per Task.
It is not globally bounded across all Tasks.

During a correlated Kubernetes/API-server failure, many Tasks could therefore
retain independent retry workers. Exponential backoff limits per-Task request
frequency, but this PR does not attempt to solve system-wide retry-storm or
thundering-herd behavior.

That broader retry architecture is intentionally left outside the scope of
YUNIKORN-3382.

PlaceholderManager also retains its existing orphan-cleanup lifecycle. UID
fencing makes stale cleanup safe for replacement Pods, but this PR does not add
identity reconciliation or stale-entry retirement to that loop.

---

### Type of change

- [x] Bug Fix
- [ ] Improvement
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

### Jira issue

Jira ID: https://issues.apache.org/jira/browse/YUNIKORN-3382

YUNIKORN-3382 already existed and this PR implements the fix for that issue.

### AI Tooling

If an AI tool was used:

- [x] The PR includes the phrase "Generated by <tool>", where <tool> is the name of the AI tool used.

Generated by OpenAI Codex.

### How has this been tested?

Local validation completed successfully:

- `make license-check`
- `make lint`
- `make test`
  - includes race-enabled package tests and `go vet`
- focused YUNIKORN-3382 tests repeated 20 times
- focused tests with deadlock detection enabled repeated 20 times
- focused race-detector tests repeated 20 times
- additional Conflict-path ordinary/deadlock/race repetitions passed
- `git diff --check`

The local environment does not provide a Kubernetes cluster/kubeconfig, so the
repository E2E suite was not run locally.

New regression coverage includes:

- failed release-path DELETE is automatically re-driven;
- duplicate release events retain a single deletion owner;
- non-retryable errors release ownership;
- terminal Tasks stop background retries;
- DELETE requests include the original Pod UID precondition;
- missing UID fails closed without sending Kubernetes DELETE;
- same-name replacement Pods are protected from stale deletion;
- Conflict + replacement UID stops stale retries;
- Conflict + NotFound stops stale retries;
- Conflict + original UID retains retry responsibility;
- Conflict + reconciliation GET error retains retry responsibility;
- a Conflict encountered by the background retry worker uses the same
  reconciliation semantics;
- accepted/resolved deletion does not directly confirm release to scheduler
  core.

### Questions:

#### Commit structure

The PR is intentionally split into four reviewable commits:

1. `[YUNIKORN-3382] Re-drive failed Pod deletion with UID fencing`
   - Task-local ownership and retry
   - UID-fenced Kubernetes DELETE
   - primary release-path regression coverage

2. `[YUNIKORN-3382] Reconcile Pod delete conflicts by UID`
   - Conflict -> GET identity reconciliation
   - stale replacement / NotFound handling
   - empty-UID `BadRequest` classification
   - conflict-path regression coverage

3. `[YUNIKORN-3382] Clarify Pod deletion lifecycle naming`
   - readability-only rename/comment cleanup
   - no behavioral change

4. `[YUNIKORN-3382] Clarify Pod deletion retry flow and coverage`
   - make the reconciliation result explicit as `noFurtherDeleteNeeded`
   - extract Conflict-specific reconciliation and retry-delay calculation helpers
   - add regression coverage for worker-side `NotFound`, repeated transient failures,
     resolved-error classification, and capped exponential backoff
   - behavior-preserving readability/testability cleanup; retry ownership and
     completion semantics are unchanged


wilfred-s (changes requested):
Please have a look at using `ExponentialBackoffWithContext` instead of building a backoff

hikaru-212:
@wilfred-s Thanks. I replaced the manual backoff with Kubernetes `wait.Backoff`. I used `DelayFunc().Until` rather than `ExponentialBackoffWithContext` to preserve the existing retry-until-resolved behavior without introducing a finite retry budget. I also added regression coverage for the initial retry delay and continued retries after the backoff reaches its cap.

PR #1097, opened 2026-09-19
[YUNIKORN-3427] Fix unsynchronised context and task access on deferred release

rjgoyln:
## Summary

`flushReleaseableTasks` is the enter-Accepted callback, so it runs under the application write lock.

There are two race conditions:

* The all-terminated path deleted the application from `ctx.applications` without holding the context lock, while the scheduling loop reads the same map under the context read lock on every tick. This results in a concurrent map read and write, which the Go runtime treats as fatal.
* The other path replayed the release without holding the task lock, reading `allocationKey` while the core callback could update it.

## Change

* Context removal is moved out of the transition and into `handle`, where it runs after the application lock has been released.
* Release replay now takes the task lock before reading the task.
* `Context.removeApplication` is folded back into `Context.RemoveApplication`, leaving no unlocked path for mutating the applications map.

The context lock is acquired before the application lock on every path that holds both. This ensures that removal cannot happen from inside the transition.

The task lock is acquired after both of them. The paths that acquire the task lock first, such as `shouldAppRelease` and `rollbackAllocation`, release it before reaching the replay path, so the replay can safely acquire it directly.

`shouldAppRelease` itself is unchanged. Since `force` is `true` on this path, YUNIKORN-3421 remains out of scope.

The two `+checklocksignore` sites added by #1079 to this function are also no longer needed after this change.

## Behavior Change

Removal now happens immediately after the transition rather than inside it.

A scheduler tick whose snapshot already contains the application could still observe it as `Accepted` and then find that it has already been removed. This window already existed before the change, so its behavior is unchanged in kind.

## Tests

Both new tests fail against the current production code under `-race`.

Each test detected the expected access pair in all 10 out of 10 runs, matching the two races described above.

### Type of change

- [X] Bug Fix

### Jira issue

Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3427

### AI Tooling

- [X] The PR includes the phrase "Generated by \<tool>", where \<tool> is the name of the AI tool used.
- [X] My use of AI contributions follows the ASF legal policy.

### How has this been tested?

- [X] New unit tests were added to cover new or changed code paths.

Also run clean: `go test ./pkg/... -race -tags deadlock -count=1` under `DEADLOCK_DETECTION_ENABLED=true DEADLOCK_TIMEOUT_SECONDS=10 DEADLOCK_EXIT=true`.

### Questions:


HuangTing-Yao (approved):
LGTM 
correct fix, solid tests. I ran the tests locally on this branch: both new tests reproduce the two races on `master` under `-race` and pass on this branch (`-race -count=10`), and `./pkg/cache/...` is clean under `-race -tags deadlock`.

Deferring the removal out of the enter-Accepted callback is the right call — it can't take `ctx.lock` while holding `app.lock` (there's a real `ctx → app` path via `UpdatePod → GetTask`). Lock ordering `ctx → app → task` holds.

PR #1099, opened 2026-09-19
[YUNIKORN-3422] Publish the scheduler cache node lists atomically

rjgoyln:
## Summary

The three node-list getters build their list on the first call after an invalidation and assign it to the cache field. They are reached only from the predicate shared lister, under the cache read lock, and the core runs preemption predicate checks on several goroutines at once. Those readers see the field `nil` together and write it at the same time, so the slice header can be published torn.

Invalidation already runs under the write lock; only the population was unsynchronised.

## Change

* The three cached lists become `atomic.Pointer[[]fwk.NodeInfo]`, published with `Store` and cleared with `Store(nil)`.

Building the lists at invalidation instead — the other option the JIRA offers — moves work proportional to the node count under the write lock, where it would run on every node status update and every pod event carrying affinity. Keeping the population lazy leaves that work on the read path, where it already runs in parallel.

Racing readers still each build a list. The write lock keeps invalidation out of the read critical section, so they build from the same `nodesMap` and agree on membership; only the publication had to change.

#1079 annotates the three assignment sites this PR removes with `+checklocksignore`; the later PR to land should drop them.

## Tests

`TestGetNodesInfoConcurrentPopulation` reads the three lists from eight goroutines under the read lock, re-invalidating between rounds. The race detector reports all three getters on `master` and none here. It only has teeth under `-race`, which is how `make test` runs.

### Type of change

- [x] Bug Fix

### Jira issue

Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3422

### AI Tooling

### How has this been tested?

### Questions:


wilfred-s:
Looks like this is ready for review...

rjgoyln:
Sorry, I forgot to mark the PR as ready for review.

manirajv06 (approved):
looks good

